PR #4, opened 2017-06-26
AC-4501: Add v1 API to support RESTful user calls

mc-nathan:
To test:

1) Get the branch.

2) make dev

3) `make dbload GZ_FILE=../accelerate/test_data/initial_schema.sql.gz`

4) Grant v1_clients permission to a user without any existing permissions and answer 'y' when/if it asks.  E.g.,
```
$ make grant-permissions PERMISSION_USER=37891-user@example.com PERMISSION_CLASSES=v1_clients
Create group name v1_clients? y
granting 37891-user@example.com add to group v1_clients
```

5) Login to the local impact-api server as the given user.

6) Create an application for the user.  Provide name, select Client type => Public, Authorization grant type => Resource owner password-based, click 'Save'.

7) In Postman load the [collection](https://www.getpostman.com/collections/4b28d750ca0b7d27dc77):  File => Import => Import from Link

8) Set the Environment variables for the user and the client id (<eye> => Edit)

9) Send 'oauth token' API from Postman

10)   Send 'v1 user list get' (should see a list of user info in the results).

11) You can test the next and previous links by pasting the URL into the url box next to 'GET V'.  This should behave the same way that they do for more generic DRF API calls like 'startup list'.  Running 'v1 user list get' should set the Postman `variable random_user_id` which is used by the other 'v1 user get' and 'v1 user patch'.

**Yotam: As far as I can see the script that sets `random_user_id` is at "simple user list". I had to add it manually to "v1 user list get", and then it worked as described.**
**Nathan: That was true for an earlier version of the collection, but it should be updated now.  You may need to reload the collection.  Let me know if that doesn't work since that would mean there's something I don't understand about Postman.**

12) You should now be able to test 'v1 user get' and 'v1 user patch'.  Note if you check off items in 'v1 user patch' it should succeed and only update the appropriate fields when you then run 'v1 user get'.

13) POST should work as well and again will set the `random_user_id` appropriately.  If you send the same args more than once you should get an error since the email is already in use.  If you are missing requirement arguments it should complain.  `is_active` is the only optional argument.

To locally test that the users created by the API work in the accelerate platform you need to first connect your local impact-api server to the MySQL instance running on your vagrant machine. Unfortunately this is pretty tricky to do at the moment, but the follow instruction should get you there:

1) Switch your local accelerate repo to the 'development' branch and run `resetdb`.

2) Edit your Vagrantfile and add the following line:
```
      mc_18.vm.network :forwarded_port, guest: 3306, host: 8306, auto_correct: true
```

after the line:
```
       mc_18.vm.network :forwarded_port, guest: 8111, host: 8181, auto_correct: true
```

Note that this is the line that starts with `mc_18.` not `mc_18_safe.`.  More precisely:
```
200a201
>       mc_18.vm.network :forwarded_port, guest: 3306, host: 8306, auto_correct: true
```

3) On your host machine run: `vagrant reload`.

4) As root on your Vagrant machine edit /etc/mysql/my.cnf and comment out the line that starts with "skip-external-locking" and the following like that starts with "bind-address".  More precisely:
```
43c43
> # skip-external-locking
---
< skip-external-locking
47c47
> # bind-address		= 127.0.0.1
---
< bind-address		= 127.0.0.1
```

5) Restart mysql on your Vagrant box by running:

`sudo /etc/init.d/mysql restart`

6) Start the webserver on the vagrant box with `runserver`.  You may want to make a change at this point to the database so you can test it later on.  For example, change the name of startup 10000 to "Foo, Inc.".

7) On your host machine, test that you can access the database by running:

`mysql -u masschallenge -psecret --port 8306 -h 0.0.0.0 mc_dev`

You should be able to see the change you made.  E.g.,

`select name from mc_startup where id=10000;`

8) On your host machine figure out your IP address either by following the instructions here:

http://osxdaily.com/2010/11/21/find-ip-address-mac/

or by running the shell command:

`ifconfig | grep netmask | tail -1 | sed 's/.*inet //' | sed 's/ .*//'`

9) In your impact-api repo edit .dev.env and change the DATABASE_URL to point to Vagrant database by going through the host machine's IP and using the masschallenge database account (password is 'secret').  More precisely:
```
5c5
< DATABASE_URL=mysql://root:root@mysql/mc_dev
---
> DATABASE_URL=mysql://masschallenge:secret@W.X.Y.Z:8306/mc_dev
```
where `W.X.Y.Z` is replaced by your host machine's IP address.

10) In your impact-api repo, make sure you are on the AC-4501 branch and run `make clean; make dev`.  This must be rerun to ensure that necessary migrations run on your local accelerate database.

11) In your impact-api repo, run:

`make grant-permissions PERMISSION_USER=37891-user@example.com PERMISSION_CLASSES=v1_clients`

(or the email of whatever user you want to use in Postman) and answer 'y' when it asks you if you want to create the group 'v1_clients'.

12) Login to the impact-api webserver as the user (37891-user@example.com) and create an application client.

13) In Postman, update the client id (and username/password if you haven't already).

14) In Postman, run 'POST oauth token'.  If you made a change, you should be able to see it either through Postman or with Browsable API.  For example, if you changed startup 10000, you should be able to see the change by going to 'startup list' in Postman.  If you do this, you will need to either be working with a superuser account or you will need to grant the permission `view_startup` (replace 'v1_clients' with 'view_startup' in step 11).

15) Run 'v1 user post'.  Assuming the user is successfully created, you will get back the new user's id.

16) Go to your local accelerate webpage and go through the password reset workflow for 'remerson@walden.pond'.  Specifically, goto login page, click "I forgot my password", enter 'remerson@walden.pond', check the webserver output for email with the reset link, go to that link after replacing the host name with your local server, and enter your new password.

17) Click 'LOGIN', enter 'remerson@walden.pond' and the password you just set.  You should be asked to accept our privacy policy and then you'll be taken to a member profile page. Rejoice!

**IMPORTANT**: After this testing you should disconnect the two environments by running the following:

In your accelerate repo:
% git checkout Vagrantfile
% vagrant reload

You may want to uncomment the lines in my.cnf and restart mysql, but it is not critical since it doesn't have a risk of getting checked in, the port is shutdown by the actions above, and it will all get cleaned up when you rebuild the vagrant environment.

In your impact-api repo:
% git checkout .dev.env
% make clean

### DIFF
--- a/web/impact/impact/models/reference.py
+++ b/web/impact/impact/models/reference.py
@@ -30,7 +30,7 @@ class Reference(MCModel):
     question_1_rating = models.IntegerField(null=True)
     question_2_rating = models.IntegerField(null=True)
     comments = models.TextField(blank=True)
-    user = models.ForeignKey(User, null=True)
+    requesting_user = models.ForeignKey(User, null=True)
 
     class Meta(MCModel.Meta):
         db_table = "mc_reference"
--- a/web/impact/impact/permissions.py
+++ b/web/impact/impact/permissions.py
@@ -94,18 +94,14 @@ class DynamicModelPermissions(BasePermission):
         try:
             boolean_value = literal_eval(text_value.title())
         except:  # pragma: no cover
-            # Regarding coverage, it's not clear that we need this code
-            # any more.  See has_object_permissions.
-            # not a bool? unequivocably reject permission
+            # Regarding coverage, see AC-4573
             raise PermissionDenied  # pragma: no cover
         return boolean_value
 
     def has_object_permission(self, request, view, obj):
         model_name = view.kwargs.get('model', "").lower()
         app_label = 'mc'
-        # This needs to be revised but is outside the scope of AC-4501.
-        # Field level permissions are no longer expected to be handled
-        # through the low level API.
+        # This needs to be revised.  See AC-4573
         for permission in self.get_field_level_perms(app_label, model_name):
             action, perm_model, field, boolean_str = (
                 self.decompose_perm(permission.codename))
--- a/web/impact/impact/permissions.py
+++ b/web/impact/impact/permissions.py
@@ -24,6 +24,15 @@ class V0APIPermissions(BasePermission):
             name=settings.V0_API_GROUP).exists()
 
 
+class V1APIPermissions(BasePermission):
+
+    authenticated_users_only = True
+
+    def has_permission(self, request, view):
+        return request.user.groups.filter(
+            name=settings.V1_API_GROUP).exists()
+
+
 class DynamicModelPermissions(BasePermission):
     perms_map = {
         'GET': ['%(app)s.view_%(model_name)s'],
--- a/web/impact/impact/permissions.py
+++ b/web/impact/impact/permissions.py
@@ -47,11 +47,6 @@ class DynamicModelPermissions(BasePermission):
     authenticated_users_only = True
 
     def has_permission(self, request, view):
-        # Workaround to ensure DjangoModelPermissions are not applied
-        # to the root view when using DefaultRouter.
-        if getattr(view, '_ignore_model_permissions', False):
-            return True
-
         model_name = view.kwargs.get('model', "").lower()
         app_label = 'mc'
         kwargs = {'app': app_label, 'model_name': model_name}
--- a/web/impact/impact/permissions.py
+++ b/web/impact/impact/permissions.py
@@ -93,14 +93,19 @@ class DynamicModelPermissions(BasePermission):
     def convert_string_to_bool(self, text_value):
         try:
             boolean_value = literal_eval(text_value.title())
-        except:
+        except:  # pragma: no cover
+            # Regarding coverage, it's not clear that we need this code
+            # any more.  See has_object_permissions.
             # not a bool? unequivocably reject permission
-            raise PermissionDenied
+            raise PermissionDenied  # pragma: no cover
         return boolean_value
 
     def has_object_permission(self, request, view, obj):
         model_name = view.kwargs.get('model', "").lower()
         app_label = 'mc'
+        # This needs to be revised but is outside the scope of AC-4501.
+        # Field level permissions are no longer expected to be handled
+        # through the low level API.
         for permission in self.get_field_level_perms(app_label, model_name):
             action, perm_model, field, boolean_str = (
                 self.decompose_perm(permission.codename))
@@ -111,7 +116,7 @@ class DynamicModelPermissions(BasePermission):
                         '{}.{}'.format(app_label, permission.codename))
                     if getattr(obj, field, None) is boolean_value:
                         if not get_user(request).has_perm(permission_code):
-                            return False
+                            return False  # pragma: no cover
                     # some other value is in place,
                     # so we don't require permissions
         return True
--- a/web/impact/impact/settings.py
+++ b/web/impact/impact/settings.py
@@ -138,8 +138,14 @@ class Base(Configuration):
     V0_SITE_NAME = bytes(os.environ.get(
         'IMPACT_API_V0_SITE_NAME', 'masschallenge.org'), 'utf-8')
 
+    # This seems too complicated.  Why would we want to
+    # override the name of an API group using environment
+    # variables?  Shouldn't this name just be a constant?
     V0_API_GROUP = bytes(os.environ.get(
         'IMPACT_API_V0_API_GROUP', 'v0_clients'), 'utf-8')
+
+    V1_API_GROUP = bytes(os.environ.get(
+        'IMPACT_API_V1_API_GROUP', 'v1_clients'), 'utf-8')
 
     OAUTH2_PROVIDER = {
         # this is the list of available scopes
--- a/web/impact/impact/settings.py
+++ b/web/impact/impact/settings.py
@@ -138,12 +138,10 @@ class Base(Configuration):
     V0_SITE_NAME = bytes(os.environ.get(
         'IMPACT_API_V0_SITE_NAME', 'masschallenge.org'), 'utf-8')
 
-    # This seems too complicated.  Why would we want to
-    # override the name of an API group using environment
-    # variables?  Shouldn't this name just be a constant?
     V0_API_GROUP = bytes(os.environ.get(
         'IMPACT_API_V0_API_GROUP', 'v0_clients'), 'utf-8')
 
+    # This and the above should get generalized.  See AC-4574.
     V1_API_GROUP = bytes(os.environ.get(
         'IMPACT_API_V1_API_GROUP', 'v1_clients'), 'utf-8')
 
--- a/web/impact/impact/tests/test_user_list_view.py
+++ b/web/impact/impact/tests/test_user_list_view.py
@@ -43,6 +43,14 @@ class TestUserListView(APITestCase):
             response = self.client.post(url, {})
             assert response.status_code == 403
 
+    def test_post_bad_key(self):
+        with self.login(username=self.basic_user().username):
+            url = reverse("user")
+            bad_key = "bad key"
+            response = self.client.post(url, {bad_key: True})
+            assert response.status_code == 403
+            assert any([bad_key in error for error in response.data])
+
     def test_post_existing(self):
         user = UserContext().user
         data = EXAMPLE_USER.copy()
--- a/web/impact/impact/tests/test_user_list_view.py
+++ b/web/impact/impact/tests/test_user_list_view.py
@@ -1,0 +1,53 @@
+# MIT License
+# Copyright (c) 2017 MassChallenge, Inc.
+
+from django.urls import reverse
+
+from impact.tests.contexts import UserContext
+from impact.tests.api_test_case import APITestCase
+from django.contrib.auth import get_user_model
+
+
+EXAMPLE_USER = {
+    "first_name": "First",
+    "last_name": "Last",
+    "email": "test@example.com",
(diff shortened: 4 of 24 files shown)


web/impact/impact/settings.py
```diff
@@ -138,9 +138,15 @@ class Base(Configuration):
     V0_SITE_NAME = bytes(os.environ.get(
         'IMPACT_API_V0_SITE_NAME', 'masschallenge.org'), 'utf-8')
 
+    # This seems too complicated.  Why would we want to
+    # override the name of an API group using environment
+    # variables?  Shouldn't this name just be a constant?
```
yotammanor: Likewise, preferably a ticket rather than a comment

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def test_post(self):
+        with self.login(username=self.basic_user().username):
+            url = reverse("user")
+            response = self.client.post(url, EXAMPLE_USER)
+            id = response.data["id"]
+            user = User.objects.get(id=id)
+            assert user.email == EXAMPLE_USER["email"]
+
+    def test_post_required_missing(self):
```
yotammanor: name of test is not clear enough.

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def test_post_bad_key(self):
+        with self.login(username=self.basic_user().username):
+            url = reverse("user")
+            bad_key = "bad key"
+            response = self.client.post(url, {bad_key: True})
+            assert response.status_code == 403
+            assert any([bad_key in error for error in response.data])
+
+    def test_post_existing(self):
```
yotammanor: This is understandable but I would prefer it to be more explicit.

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -0,0 +1,61 @@
...
+        user = UserContext().user
+        data = EXAMPLE_USER.copy()
+        data["email"] = user.email
+        with self.login(username=self.basic_user().username):
+            url = reverse("user")
+            response = self.client.post(url, data)
+            assert response.status_code == 403
```
yotammanor, line 62: Is there a way to assert that the reason of the 403 is the existence of the user?

yotammanor: Or otherwise, I'd rewrite this to assert that in a given login context you can post successfully, and then post again same data and get a 403.

web/impact/impact/utils.py
```diff
@@ -1,15 +1,64 @@
...
+    return None
+
+
+def find_gender(gender):
+    if not hasattr(gender, "lower"):
+        return None
+    gender = GENDER_TRANSLATIONS.get(gender.lower(), gender)
+    if gender in VALID_GENDERS:
+        return gender
+    return None
```
yotammanor, line 64: should this code exist in accelerate as well?

web/impact/simpleuser/models.py
```diff
@@ -30,10 +29,11 @@ def _create_user(self, email, password,
...
             raise ValueError('An email address must be provided.')
         email = self.normalize_email(email)
+        if "is_active" not in extra_fields:
+            extra_fields["is_active"] = True
         user = self.model(email=email,
-                          is_staff=is_staff, is_active=True,
-                          is_superuser=is_superuser, last_login=now,
-                          date_joined=now, **extra_fields)
+                          is_staff=is_staff, is_superuser=is_superuser,
+                          last_login=now, date_joined=now, **extra_fields)
```
yotammanor, line 36: parallel changes in accelerate/django-accelerator?

yotammanor (changes requested):
Functionality as described (note the edit on test instructions). Tests are passing. Style clean. Coverage looks fine but see comments on choice to mark nocover.

I added a bunch of comments, most of them are informative questions or requests for clarifications, but some changes are mixed in and would love your response an all. Thanks :)

web/impact/impact/models/reference.py
```diff
@@ -30,7 +30,7 @@ class Reference(MCModel):
     question_1_rating = models.IntegerField(null=True)
     question_2_rating = models.IntegerField(null=True)
     comments = models.TextField(blank=True)
-    user = models.ForeignKey(User, null=True)
+    requesting_user = models.ForeignKey(User, null=True)
```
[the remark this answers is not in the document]
shanbady: I believe this particular change is one that is already propagated on accelerate

shanbady:
My suggested method of testing this on an existing accelerate database is to setup an accelerae instance and expose mysql and then set the settings.DATABASES setting in impact appropriately. You should then have accelerate and impact connected to the same running mysql instance.

web/impact/impact/models/reference.py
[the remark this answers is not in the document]
mc-nathan: Yep.  This was just keeping up to date with the other repo.  It was triggered due to a test I believe.

web/impact/impact/permissions.py
```diff
@@ -38,11 +47,6 @@ class DynamicModelPermissions(BasePermission):
     authenticated_users_only = True
 
     def has_permission(self, request, view):
-        # Workaround to ensure DjangoModelPermissions are not applied
-        # to the root view when using DefaultRouter.
-        if getattr(view, '_ignore_model_permissions', False):
-            return True
-
```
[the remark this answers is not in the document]
mc-nathan: It was not covered and as far as I can tell is never true.  I believe it is a remnant from the back when the impact-api was first getting developed.

web/impact/impact/permissions.py
```diff
@@ -89,14 +93,19 @@ def action_matches(self, action, request):
     def convert_string_to_bool(self, text_value):
         try:
             boolean_value = literal_eval(text_value.title())
-        except:
+        except:  # pragma: no cover
+            # Regarding coverage, it's not clear that we need this code
+            # any more.  See has_object_permissions.
```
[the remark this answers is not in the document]
mc-nathan: Removing it will take more thought than I wanted to put into this ticket and it is outside the scope.  It's not just a matter of deleting the class since it is still used for somethings according to Shankar.

web/impact/impact/permissions.py
```diff
@@ -89,14 +93,19 @@ def action_matches(self, action, request):
...
-            raise PermissionDenied
+            raise PermissionDenied  # pragma: no cover
         return boolean_value
 
     def has_object_permission(self, request, view, obj):
         model_name = view.kwargs.get('model', "").lower()
         app_label = 'mc'
+        # This needs to be revised but is outside the scope of AC-4501.
+        # Field level permissions are no longer expected to be handled
+        # through the low level API.
```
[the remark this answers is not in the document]
mc-nathan: Added ticket and revised comment.

web/impact/impact/permissions.py
```diff
@@ -107,7 +116,7 @@ def has_object_permission(self, request, view, obj):
                         '{}.{}'.format(app_label, permission.codename))
                     if getattr(obj, field, None) is boolean_value:
                         if not get_user(request).has_perm(permission_code):
-                            return False
+                            return False  # pragma: no cover
```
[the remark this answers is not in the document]
mc-nathan: Revised comment to reference new ticket.

web/impact/impact/settings.py
mc-nathan: Good catch.  I left it there precisely so a new ticket would get created.

web/impact/impact/tests/test_user_list_view.py
mc-nathan: Updated.

web/impact/impact/tests/test_user_list_view.py
mc-nathan: Updated.

web/impact/impact/tests/test_user_list_view.py
mc-nathan: Done.

web/impact/impact/utils.py
mc-nathan: This is essentially handled by the choices on the gender field of CoreProfile which essentially imposes the same constraint on drop down UI elements representing gender.  The challenge with the API of course is we don't have a UI element to do this work for us.

web/impact/simpleuser/models.py
mc-nathan: Make sense.  PR now exists for that repo as well.

mc-nathan:
@yotammanor I believe I have addressed all your review comments.  I have also added details about how to test the user creation against a local accelerate database instance.

yotammanor (approved):
Code changes are good, and questions are answered. Tests and coverage and style are all clean.
Functionally works as described throughout the entire process.

MAW.

shanbady (approved):
This looks good and functions as advertised - one thing to notice is that this doesn't seem to be creating a fully usable user with a profile as the accelerate platform would expect to have but it does create the django user which is probably all we need for. The other comment (or question rather) i have is - do we need a separate version for this user functionality at all since we don't have any active clients? i think it would cause less confusion to clients if we stick this all under v0 rather than breaking just the user call out to v1.

mc-nathan:
Regarding "creating a fully usable user with a profile", I believe it does and is what all that testing is all about.  If you have a clear case where that's not the case, then let me know.
Regarding v1 vs. v0.  I would like to deprecate v0 before v1.  Also they have different clients ("marketing site" vs. Salesforce) so we probably want to control access separately.

PR #7, opened 2017-06-30
AC-4572: merge nginx service in web service

shanbady:
This contains commits from sarah's PR ([AC-4335] Retool the API deploy process to work with a single public repository) since it was necessary to test that this works on staging. Testing this is going to be a bit tricky (atleast on a live server) since we explicitly redirect to https://api.masschallenge.org if the webpage is accessed via http.

To verify, run make nuke, make build, and make dev and verify that:

- locally the api server works as expected
- staticfiles are loading
- when running ```docker-compose ps``` does not show an "nginx" container running


mc-nathan (changes requested):
The testing as described works.  However, neither the ticket nor the PR are clear about exactly what the problem is.  Specifically I don't know how to reproduce it to tell if this actually fixes the problem.  I also don't understand the source of the travis failure.  I would like to understand this better before approving.

mc-nathan (approved):

shanbady:
@mc-nathan  this fixes an issue which has a couple of pieces to it:

- ecs seems to have a limitaition where you cannot mount volumes dynamically on runtime (unlike docker-compose locally) from other container. nginx needs to be able to get at the static-compiled directory on runtime to properly serve all the static files. The other issue is that the static-compiled directory is non-existent when the actual nginx container gets built (since the files are generated after the "web" container gets built.

- Another issue in testing this is that on prod, the ssl cert is pinned before the web server layer (done via elb automatically) and since we have not allocated a separate elb address for the staging server and pinned the cert to it, the only real way to test this particular fix with SSL and all without that piece is either to get it on production, or get it on staging and remove the portion of the nginx file that automatically redirects to https://api.masschallenge.org

knittingarch (approved):

PR #10, opened 2017-07-05
AC-4604: compiled static assets on production api not getting generated correctly

shanbady:
To test this, run make dev locally and open up a developer console and verify that there no 404's for static files. Part of the difficulty here is that this is something that previously worked locally so nothing will visibly change locally but not on production staticfiles will properly load (please note that this was an issue that was manually patched on production to confirm it works so you will not see this on the current prod)

### DIFF
--- a/web/impact/templates/base.html
+++ b/web/impact/templates/base.html
@@ -9,7 +9,7 @@
   {% endblock %}
   <title>{% block title %}{% if name %}{{ name }} – {% endif %}Impact{% endblock %}</title>
   {% block style %}
-    <link rel="stylesheet" type="text/css" href="{% static "css/build/style.min.css" %}"/>
+    <link rel="stylesheet" type="text/css" href="{% static "css/dev/style.css" %}"/>
     {% block bootstrap_theme %}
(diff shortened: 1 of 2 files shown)


mc-nathan:
This looks like a short term solution just to get things to work.  If we are going to rely on dev/style.css rather than build/style.min.css, then we should get rid of style.min.css entirely.  However, I think we probably ultimately want the minimized version and we should adjust our deploy process to get that right.  As short term fix I'm willing to approve this, but it needs a follow on ticket to fix it the right way.

mc-nathan (changes requested):
At minimum create a ticket for either correctly creating and deploy the *.min.css files or for eliminating the generation and use of the *.min.css files.

mc-nathan:
Per the ticket the follow on ticket is AC-4603

mc-nathan (approved):

knittingarch (approved):
Functional Testing: No 404s present (except the favicon one)
Code: Looks good.
Code Quality: Passes.
Tests: N/A
Coverage: N/A

PR #11, opened 2017-07-17
[AC-4646] add docker run flag that removes process when done

yotammanor:
**Test:**

* On development:
  * `make code-check`
  * `docker ps -a`
  * See that a freshly exited container appears in the list.
* On AC-4646
  * `make code-check`
  * `docker ps -a`
  * See that nothing is added.


mc-nathan (approved):
Works as expected.

knittingarch (approved):
Functional Testing: Works as expected.
Code: Clean
Tests: pass.

PR #13, opened 2017-07-18
[AC-4629] - set Proxy API calls to Test-3's Host

yotammanor:
The current settings redirect Proxy API calls to production, instead of test.
This change solves this.

### DIFF
--- a/web/impact/impact/settings.py
+++ b/web/impact/impact/settings.py
@@ -156,7 +156,9 @@ class Base(Configuration):
     CORS_ORIGIN_ALLOW_ALL = True
     # settings.py
     REST_PROXY = {
-        'HOST': bytes(os.environ.get('ACCELERATE_SITE_URL'), 'utf-8'),
+        'HOST': bytes(
+            os.environ.get('ACCELERATE_SITE_URL',
+                           'https://accelerate.masschallenge.org'), 'utf-8'),
         'AUTH': {
             'user': None,
             'password': None,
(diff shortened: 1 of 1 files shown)


web/impact/impact/settings.py
```diff
@@ -157,7 +157,7 @@ class Base(Configuration):
     CORS_ORIGIN_ALLOW_ALL = True
     # settings.py
     REST_PROXY = {
-        'HOST': 'https://accelerate.masschallenge.org',
+        'HOST': 'https://accelerate-test-3-sf-elb-1843213725.us-east-1.elb.amazonaws.com',
```
shanbady: only suggestion here is to have this derived from the environment as a separate ACCELERATE_SITE_URL setting similar to how we configure the database ALLOWED hosts.

shanbady (changes requested):

web/impact/impact/settings.py
yotammanor: Good idea.

yotammanor: Added ACCELERATE_SITE_URL to travis's settings with TEST3's url.

web/impact/impact/settings.py
```diff
@@ -157,7 +156,7 @@ class Base(Configuration):
     CORS_ORIGIN_ALLOW_ALL = True
     # settings.py
     REST_PROXY = {
-        'HOST': 'https://accelerate.masschallenge.org',
+        'HOST': bytes(os.environ.get('ACCELERATE_SITE_URL'), 'utf-8'),
```
shanbady: we should probably default this to the production accelerate site

shanbady (changes requested):

shanbady (approved):

PR #17, opened 2017-08-09
[AC-4708] - Impact coverage report is not diff-aware

yotammanor:
**Changes Introduced in [AC-4708](https://masschallenge.atlassian.net/browse/AC-4708):**

- rewrite coverage targets to print a diff-aware report.
- rewrite coverage targets to be more similar to the equivalent accelerate aliases.
- add target coverage-html-open as a separate command from coverage-html.

**To test:**
- make a
- on development, change a .py file, run `make coverage`. You'll see the entire file tree printed.
- on branch AC-4708, make that same change, run `make coverage`. You'll see just that file being printed. Also, html won't open automatically.
- change a file in tests/django_migrations/non-.py file, and it won't be printed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,10 @@ targets = \
   code-check \
   comp-message \
   coverage \
+  coverage-run \
+  coverage-report \
   coverage-html \
+  coverage-html-open \
   dbdump \
   dbload \
   dbshell \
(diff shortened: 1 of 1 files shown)


web/impact/impact/models/user_role.py
```diff
@@ -30,11 +30,11 @@ class UserRole(MCModel):
     STAFF = "Staff"
     TEAM = "Team"
 
-    OFFICE_HOUR_ROLES = set([AIR,
-                             PARTNER,
-                             PARTNER_ADMIN,
-                             OFFICE_HOUR_HOLDER,
-                             MENTOR])
+    OFFICE_HOUR_ROLES = {AIR,
```
mc-nathan: This change is not in Accelerate and seems irrelevant to this PR.

mc-nathan (changes requested):
Remove change to user_role.py or make it clear why it's being made here and not in accelerate.

yotammanor:
@mc-nathan reverted the utility commit, and updated testing instructions above.

mc-nathan (approved):

knittingarch (changes requested):
Functional Issues
On the branch, the changed file doesn't show up after running `make coverage`.

Code: Looks good.
Code Quality: Clean.
Tests: Pass
Coverage: N/A

knittingarch:
screen shot 2017-08-11 at 10 33 13 am

mc-nathan:
I was only able to reproduce the behavior reported by @knittingarch by making a change to a file that already had full coverage in a way that it still has full coverage (e.g., add a blank line).  This seems reasonable to me.  I learned that the change that @knittingarch made was to change a True to a False in permission.py  which probably didn't change the coverage.  She's going to test a bit further.

knittingarch (approved):

PR #20, opened 2017-08-14
[AC-4677-api] Change field definitions to mirror mc models

jonkiparsky:
ProgramRole.name, NamedGroup.name, UserLabel.label, and StartupLabel.label should all have same length

To review: check model changes, verify that migrations (both here and accelerate side) are correct

Note that as of Aug 14 none of the affected fields has a length > 50.

### DIFF
--- a/web/impact/impact/models/program_role.py
+++ b/web/impact/impact/models/program_role.py
@@ -7,12 +7,15 @@ from impact.models.mc_model import MCModel
 from impact.models.program import Program
 from impact.models.user_role import UserRole
 from impact.models.user_label import UserLabel
-from impact.models.utils import is_managed
+from impact.models.utils import (
+    is_managed,
+    LABEL_LENGTH,
+)
 
 
 class ProgramRole(MCModel):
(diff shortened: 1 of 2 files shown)


web/impact/impact/models/program_role.py
```diff
@@ -12,7 +12,7 @@
 
 class ProgramRole(MCModel):
     program = models.ForeignKey(Program)
-    name = models.CharField(max_length=30, unique=True, db_index=True)
+    name = models.CharField(max_length=255, unique=True, db_index=True)
```
shanbady: shouldnt we be using the constant defined in utils?

shanbady (changes requested):

web/impact/impact/models/program_role.py
jonkiparsky: Good call, fixed

mc-nathan (approved):

mc-nathan:
@shanbady MAW

shanbady (approved):

PR #21, opened 2017-08-14
Ac 4681: Quick Start Guide for impact-api

mc-nathan:
Has been reviewed by Jeff and has his sea lion.

PR #23, opened 2017-08-15
[AC-4670a-api] - Switch Currency Model to django-accelerator - Blue-Green Step 1 

yotammanor:

**Test:**
- on accelerate, checkout AC-4670a, pull, resetdb, dumpdb.
- on impact-api, checkout AC-4670a-api, pull, make build, make dev.
- Note that new_currency field is now returned, with null values.
- Note that http://localhost:8000/api/accelerator/currency is becoming automatically available just by installing the model.
  
**Note:**
- step 2+3 will be skipped for impact-api, so it's worth repeating steps above with a dbdump from AC-4670b + AC-4670c (which should work just the same).


mc-nathan:
Minor functional issue: I noticed that by default the AC-4670a-api branch '/' redirects to /api/accelerator/.  It seems like for now we should stick with redirecting to /api/impact/.  Ideally I would like the "index" page for the site to list the different APIs and provide index pages for each of them. Specifically /api/impact/, /api/accelerator/, /api/v0/ and /api/v1/.  I think we should *not* list /api/simpleuser/ since it's too confusing.  Created AC-4758 for this work.

PR #24, opened 2017-08-16
[AC-4670d-api] - Switch Currency Model to django-accelerator - Blue-Green Step 4 (skipping 2+3)

yotammanor:

**Test:**
- on accelerate, checkout AC-4670d, pull, resetdb, dumpdb.
- on impact-api, checkout AC-4670d-api, pull, make build, make dev.
  - Note that new_currency field is now returned on Startup, with identical values to currency.
  - Note that http://localhost:8000/api/accelerator/ is now available.
  - Note that http://localhost:8000/api/impact/Currency/ is still available, but does not appear as a link in http://localhost:8000/api/impact/


pbx (approved):
Does what it says on the tin.

mc-nathan:
Note I got an error if I switch accelerate to "e" before moving impact-api to "d".  I'm guessing that I should have moved impact-api to "e" before moving accelerate, but that's not clear from the current notes and we'll need to be clear on that in the deployment process.  It would be helpful to have a deployment process description on the AC-4670 ticket.

mc-nathan:
Ran into the same problem going to accelerate "d" and then moving impact-api to "e" and still got an error.  Didn't have time to dig in more deeply.

PR #25, opened 2017-08-16
[AC-4670e-api] - BLOCKED - Switch Currency Model to django-accelerator

yotammanor:

**Test:**
- on accelerate, checkout AC-4670e, pull, resetdb, dumpdb.
- on impact-api, checkout AC-4670e-api, pull, make build, make dev.
- Note that new_currency field is no longer available, only currency.
- Note that http://localhost:8000/api/impact/Currency/ is no longer available.
  

**Note**
- Step 6 will be skipped for impact-api, so it's worth repeating steps above with a dbdump from AC-4670f (which should work just the same).


web/impact/impact/settings.py
```diff
@@ -79,7 +79,9 @@ class Base(Configuration):
         'oauth2_provider',
         'corsheaders',
         'embed_video',
+        'accelerator.apps.AcceleratorConfig',
```
pbx, line 82: What is this about?

pbx (commented):
Works as it should, I believe. 

When I ran `make dev` I did notice a "models have changes that are not yet reflected in a migration" notice fly by (after having `make dbload`ed the `dumpdb` output from accelerate@AC-4670e). Expected?

mc-nathan (approved):
Looks good.

web/impact/impact/settings.py
mc-nathan: It's the name of the django-accelerator app.

PR #26, opened 2017-08-23
AC-4706- Create schema endpoint for impact api

shanbady:
This branch adds the ability to get a schema for a given api request by sending an OPTIONS request to a particular endpoint. It also resolves some issues when the Organization models went in.

To test:

1 - checkout this branch and make sure you have the latest development checkout out for accelerate.

2 - load the development database by running `make dbload GZ_FILE=../accelerate/test_data/initial_schema.sql.gz`

3 - grant demoadmin access to the v1 apis via `make grant-permissions PERMISSION_USER=demoadmin@masschallenge.org PERMISSION_CLASSES=v1_clients`

4 - login as the demoadmin user and browse to a v1 endpoint:
     - /api/v1/organization/
     - /api/v1/user/

5 - the interface should now allow you to make an 'OPTIONS' call to the endpoint which should return a schema with field definitions for each object.

screen shot 2017-08-23 at 11 51 17 am


yotammanor:
@shanbady note that it's `make dbload`, not `make loaddb` (this should probably change, though)

mc-nathan:
I was emulating 'make dbshell' since that is the Django standard name for that.

yotammanor:
@mc-nathan but it's the django has loaddata and dumpdata which are standard, and accelerate has dumpdb and loaddb.

(plus I think grammatically dbshell is meant to be a noun and loaddb is meant to be a verb, but I might get it wrong.)

web/impact/impact/tests/test_organization_list_view.py
```diff
@@ -27,11 +32,45 @@ def test_get_startup(self):
...
+            organizations[0].id,
+            organizations[1].id]).update(
+            updated_at=lastweek)
+        with self.login(username=self.basic_user().username):
+            url = "{base_url}?updated_at__gt={datestr}".format(
+                base_url=reverse("organization"),
+                datestr=lastweek.strftime("%Y-%m-%dT%H:%M:%S.%fZ"))
+            response = self.client.get(url)
+        json_response = json.loads(response.content)
+        self.assertTrue(json_response['count'] > 0)
```
yotammanor: I'd test here that the result you want is actually there. Otherwise it might be vulnerable to cross-test leakage.

web/impact/impact/v1/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+    """
+
+    def determine_metadata(self, request, view):
+        metadata = super().determine_metadata(request, view)
+        metadata['actions'] = ORGANIZATION_ACTIONS
+
+        return metadata
+
+
+class UserMetadata(SimpleMetadata):
```
mc-nathan: This seems really WET.  At minimum there is no need for the repeated comment.  If we are going to declare these as separate classes, then I would prefer if they were separate files.  A better option in my opinion would be to have a loop that dynamically creates appropriate classes, but there are certainly arguments against this approach.  However, what is here doesn't conform to our standards and has a lot of meaningless duplication (like the comments) that I'd like to see fixed one way or another.

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -3,11 +3,18 @@
...
+    MemberProfile
+)
+import pytz
+from impact.tests.factories import StartupTeamMemberFactory
 from impact.tests.contexts import UserContext
 from impact.tests.api_test_case import APITestCase
+import simplejson as json
 from impact.v1.views.user_list_view import EMAIL_EXISTS_ERROR
-
+import datetime
```
yotammanor, line 17: why not a timezone-aware `from django.utils import timezone [as datetime]`

web/impact/impact/v1/views/organization_users_view.py
```diff
@@ -9,12 +8,14 @@
...
     StartupTeamMember,
 )
+from impact.v1.metadata import OrganizationUsersMetadata
 
 
 class OrganizationUsersView(APIView):
     permission_classes = (
         V1APIPermissions,
     )
+    metadata_class = OrganizationUsersMetadata
```
mc-nathan: Another alternative would be for the ACTIONS variable from metadata.py to be declared in this file and then have a utility function that creates the appropriate metadata class.

web/impact/impact/v1/views/user_detail_view.py
```diff
@@ -68,3 +74,13 @@ def patch(self, request, pk):
                 profile.save()
         user.save()
         return Response(status=200)
+
+    def partner_organizations(self, user):
+        team = PartnerTeamMember.objects.filter(
```
mc-nathan: This should use user.partnerteammember_set() instead of starting from the class.

web/impact/impact/v1/views/user_detail_view.py
```diff
@@ -68,3 +74,13 @@ def patch(self, request, pk):
...
+            team_member=user)
+        return team.values_list('partner__organization_id', flat=True)
+
+    def startup_organizations(self, user):
+        team = StartupTeamMember.objects.filter(
```
mc-nathan: This should use user.startupteammember_set() instead of starting from the class.

web/impact/impact/v1/views/user_list_view.py
```diff
@@ -100,6 +106,44 @@ def _invalid_keys(self, user_keys):
...
+        )
+    if updated_at_gt:
+        queryset.filter(
+            Q(expertprofile__updated_at__isnull=False) |
+            Q(entrepreneurprofile__updated_at__isnull=False) |
+            Q(memberprofile__updated_at__isnull=False)
+        ).exclude(
+            Q(expertprofile__updated_at__lte=updated_at_gt) |
+            Q(entrepreneurprofile__updated_at__lte=updated_at_gt) |
+            Q(memberprofile__updated_at__lte=updated_at_gt)
```
yotammanor, line 144: Why exclude opposite instead of include direct?

web/impact/impact/v1/views/user_list_view.py
```diff
@@ -121,7 +160,25 @@ def _serialize_user(user):
...
+        profile_types = {
+            'EXPERT': ExpertProfile,
+            'ENTREPRENEUR': EntrepreneurProfile,
+            'MEMBER': MemberProfile
         }
+        return profile_types.get(base_profile.user_type)
+
+
+def _get_profile_update_timestamp(user):
+    profile_class = _get_profile_class(user)
```
mc-nathan: Seems like this should use `impact.utils.get_profile` rather than essentially reinventing it with _get_profile_class.

mc-nathan (changes requested):
Caught a few things, but none of it should be too hard.

yotammanor (approved):
Lot of stuff going on here.. 
Functionally seems correct. tests passing, code-check clean, fully covered. Left some non-blocking comments. 

I don't like the fact that we are squeezing all of those changes together and I'm worried that we might miss something, but given the circumstances I'm ok with this.

web/impact/impact/v1/views/user_list_view.py
shanbady: The problem is that the 'get_profile' method in accelerate is something we monkey patch in - the concept of a profile essentially goes away in the latest django which is why this is necessary here.

web/impact/impact/v1/views/user_detail_view.py
shanbady: fixed

web/impact/impact/v1/views/user_detail_view.py
shanbady: fixed

mc-nathan:
Functional issue: '/' now goes to http://localhost:8000/api/accelerator/, but it should go to http://localhost:8000/api/impact/.  I think this is due to the way the merge was done with development.  We had fixed this issue previously on development.

mc-nathan:
Functional issues: OPTIONS for /api/v1/organization/16576/ currently lists POST which is not supported.  OPTIONS for /api/v1/user/14071/ does not include PATCH.

mc-nathan:
OPTIONS for http://localhost:8000/api/v1/user/ does not include POST.

mc-nathan:
OPTIONS for /api/v1/organization/ lists POST, but it's not supported yet.

web/impact/impact/v1/views/user_list_view.py
shanbady: I was seeing strange side-effects when inverting this clause - in particular it seemed to include items that were null etc

web/impact/impact/v1/metadata.py
shanbady: I have compressed them into a single class that derives the appropriate metadata from the view name. tidy!

web/impact/impact/v1/views/organization_users_view.py
shanbady: fixed - the metadata class is now a single class used by all high level views

web/impact/impact/tests/test_user_list_view.py
shanbady: switching to this and removing the pytz.utc didnt seem to get rid of the warning messages when running tests.

mc-nathan:
OPTIONS /api/v1/organization/16576/ returns:
```
{
    "name": "Organization Detail",
    "description": "",
    "renders": [
        "application/json",
        "text/html"
    ],
    "parses": [
        "application/json",
        "application/x-www-form-urlencoded",
        "multipart/form-data"
    ],
    "actions": null
}
```

Should have a "GET".

web/impact/impact/v1/views/user_list_view.py
mc-nathan: My point is that there is already a get_profile method in impact-api (see web/impact/impact/utils.py).  It is different from the one in accelerate.  If profiles go away that's fine, but we should not implement it twice in the meantime.

web/impact/impact/v1/metadata.py
```diff
@@ -0,0 +1,173 @@
...
+METADATA_MAP = {
+    'User List': USER_ACTIONS,
+    'Organization List': ORGANIZATION_ACTIONS,
+    'Organization Users': ORGANIZATION_USER_ACTIONS,
+    'User Organizations': USER_ORGANIZATION_ACTIONS,
+    'User Detail': USER_DETAIL_ACTIONS
+}
+
+
+class ImpactMetadata(SimpleMetadata):
```
mc-nathan: Nice!

web/impact/impact/v1/views/user_list_view.py
```diff
@@ -100,6 +106,44 @@ def _invalid_keys(self, user_keys):
         for key in set(user_keys) - set(ALL_USER_RELATED_KEYS):
             self.errors.append(INVALID_KEY_ERROR.format(key))
 
+    def _user_results(self, limit, offset):
```
mc-nathan: Not a blocker, but it would be nice if this method had the same name here and in organization_list_view.py (or at least conformed to the same pattern).  It would be even nicer if the redundant code was abstracted away into a shared method, but that might not be worth the effort.

web/impact/simpleuser/models.py
```diff
@@ -104,10 +104,13 @@ class User(AbstractUser):
     """
     username = models.CharField(_('username'), max_length=150, blank=True)
 
-    class Meta:
-        managed = False
-        db_table = 'auth_user'
+    def get_short_name(self):  # pragma: no cover
```
mc-nathan, line 107: Add comment about why this is here and uncovered.

shanbady: done

web/impact/impact/v1/views/user_list_view.py
shanbady: I've made the names consistent for now at some level we could pull things into a superclass but I'm going to leave that on the table for another time

mc-nathan (approved):

PR #30, opened 2017-08-29
[AC-4795] Failure when loading Jim Miller's history

mc-nathan:
To test:
- Load the clean database and [grant a superuser account v1_clients permission](https://github.com/masschallenge/impact-api/blob/development/README.md#granting-permissions)
- Login as that user and go to http://localhost:8000/api/v1/user/182/history
- On development you'll get 500/stack trace
- On the branch it provides a reasonable response


yotammanor (approved):
Works as described, code-check clean, coverage good, code is clean. MAW.

pbx (approved):
Works. Got user history on branch; got `'>' not supported between instances of 'datetime.datetime' and 'NoneType'` on development. Code looks good, clean, tests pass.

I also confirmed that the test dies with the above error when transplanted into development branch.

PR #32, opened 2017-09-06
Ac 4788: Move django-accelerator to the 2017.09.06a release

mc-nathan:
```
make clean
make build
make dbload GZ_FILE=../accelerate/db_cache/initial_schema.sql.gz
```
will report "Applying accelerator.0004_update_organization_url_slug" on the branch.  On development it will not run that migration.

Also the help text for reloading the database has /db_cache/ rather than /test_data/ on the branch vs. development.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ target_help = \
 
 
 dbload_error_msg = GZ_FILE must be set. \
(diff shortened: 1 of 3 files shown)


yotammanor (changes requested):
I introduced the better change for handling the db in #33.

Otherwise, works as described.

yotammanor (approved):

pbx (approved):

PR #33, opened 2017-09-07
[AC-4477] - Switch to using S3 clean database

yotammanor:
**Changes Introduced in [AC-4477](https://masschallenge.atlassian.net/browse/AC-4477):**
- add target fetchremotedb, that removes existing cache and downloads
  an up to date copy of the DB_FILE_NAME specified.
- update dbload_error_msg to expect a db to exist in ./db_cache/
- update gitignore to exclude ./db_cache/

**Test:**
- run `make load-remote-db`. You'll get a descriptive error message. Also, **./db_cache/** will be created if it doesn't exist yet.
- run `make load-remote-db DB_FILE_NAME=initial_schema.sql.gz`. This will start a download of the given file, and once the file is downloaded, that file will be loaded to the DB.
- if you run the same command again, the existing local copy will be removed first.
-in a different terminal, run `make dev`
- run `make load-db GZ_FILE=./db_cache/initial_schema.sql.gz`, should work as usual (note that it was renamed).

you can also test:
- `make dbload` shows an error message re rename. Also `make dbshell` and `make dbdump`.
- running `make fetch-remote-db` will do a download only, without loading the DB.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ targets = \
   dbshell \
   deploy \
   dev \
+  fetchremotedb \
   grant-permissions \
   help \
   lint \
@@ -46,6 +47,7 @@ target_help = \
   "dbload - Load gzipped database file. GZ_FILE must be defined." \
   "dbshell - Access to running MySQL." \
   "dev - Start all containers needed to run a webserver." \
+  "fetchremotedb - Updates a DB image from remote container." \
   "grant-permissions - Grants PERMISSION_CLASSES to PERMISSION_USER." \
   "help - Prints this help message." \
   "lint - Runs any configured linters (pylint at the moment)." \
@@ -61,7 +63,10 @@ target_help = \
 
 
 dbload_error_msg = GZ_FILE must be set. \
-  E.g. 'make dbload GZ_FILE=../accelerate/test_data/initial_schema.sql.gz'
+  E.g. 'make dbload GZ_FILE=./db_cache/initial_schema.sql.gz'
+
+remotedbload_error_msg = DB_FILE_NAME must be set. \
+  E.g. 'make remotedbload DB_FILE_NAME=initial_schema.sql.gz'
 
 grant_permissions_error_msg = PERMISSION_USER and PERMISSION_CLASSES must be \
   set.  E.g., 'make grant-permissions PERMISSION_USER=test@example.org PERMISSION_CLASSES=v0_clients'
@@ -182,6 +187,17 @@ endif
 	@echo "drop database mc_dev; create database mc_dev;" | docker-compose run --rm web ./manage.py dbshell
 	@gzcat $(GZ_FILE) | docker-compose run --rm web ./manage.py dbshell
 	@docker-compose run --rm web ./manage.py migrate --no-input
+
+fetchremotedb: $(shell if [ ! -d "./db_cache/" ]; then mkdir ./db_cache/; fi;)
+fetchremotedb:
+ifndef DB_FILE_NAME
+	$(error $(remotedbload_error_msg))
+endif
+	@echo cleaning cache for ${DB_FILE_NAME}
+	@$(shell if [ -f ./db_cache/${DB_FILE_NAME} ]; then rm ./db_cache/${DB_FILE_NAME};fi;)
+	@echo downloading db...
+	@wget -P ./db_cache/ https://s3.amazonaws.com/public-clean-saved-db-states/${DB_FILE_NAME}
+	@echo "DB ${DB_FILE_NAME} is now up to date. You can load it by running 'make dbload GZ_FILE=./db_cache/${DB_FILE_NAME}'."
 
 restart:
 	@docker-compose restart web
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -16,6 +16,7 @@ targets = \
   dbshell \
   deploy \
   dev \
+  fetchremotedb \
```
mc-nathan: I'm not loving "fetchremotedb".  I would rather see "fetch-remote-db".  I can see how you got to smashing it all together from "dbshell" and "dbload".  The only reason I used "dbshell" is because that's the Django standard.  I think we should switch using '-' separators everywhere and make them more readable 'db-shell', 'load-db', and 'fetch-remote-db'.

Makefile
```diff
@@ -183,6 +188,17 @@ endif
 	@gzcat $(GZ_FILE) | docker-compose run --rm web ./manage.py dbshell
 	@docker-compose run --rm web ./manage.py migrate --no-input
 
+fetchremotedb: $(shell if [ ! -d "./db_cache/" ]; then mkdir ./db_cache/; fi;)
```
mc-nathan: A more make-ish way to do this is:
```
fetchremotedb: db_cache

db_cache:
	mkdir -p db_cache
```

Makefile
```diff
@@ -183,6 +188,17 @@ endif
...
+fetchremotedb:
+ifndef DB_FILE_NAME
+	$(error $(remotedbload_error_msg))
+endif
+	@echo cleaning cache for ${DB_FILE_NAME}
+	@$(shell if [ -f ./db_cache/${DB_FILE_NAME} ]; then rm ./db_cache/${DB_FILE_NAME};fi;)
```
mc-nathan: Again more make-ish would be:
```
	@rm -f db_cache/$(DB_FILE_NAME)
```

Makefile
mc-nathan: Also given the number times "db_cache" appears, I would assign this to a make variable.  Say,
```
DB_CACHE_DIR = db_cache/
```
The you would do things like `${DB_CACHE_DIR}${DB_FILE_NAME}`

As a side note, I'm pretty sure the `./` isn't buying anything for you, but you can add it to the variable definition if you want.

mc-nathan (changes requested):
Overall works as advertised.  Some naming quibbles and some suggestions to make the target more "make-ish".

Makefile
```diff
@@ -61,7 +63,10 @@ target_help = \
 
 
 dbload_error_msg = GZ_FILE must be set. \
-  E.g. 'make dbload GZ_FILE=../accelerate/test_data/initial_schema.sql.gz'
+  E.g. 'make dbload GZ_FILE=./db_cache/initial_schema.sql.gz'
+
+remotedbload_error_msg = DB_FILE_NAME must be set. \
+  E.g. 'make remotedbload DB_FILE_NAME=initial_schema.sql.gz'
```
mc-nathan: I think you want to change this to "fetchremotedb" (or "fetch-remote-db" if you accept my previous comment).

Makefile
```diff
@@ -183,6 +188,17 @@ endif
...
+fetchremotedb: $(shell if [ ! -d "./db_cache/" ]; then mkdir ./db_cache/; fi;)
+fetchremotedb:
+ifndef DB_FILE_NAME
+	$(error $(remotedbload_error_msg))
+endif
+	@echo cleaning cache for ${DB_FILE_NAME}
+	@$(shell if [ -f ./db_cache/${DB_FILE_NAME} ]; then rm ./db_cache/${DB_FILE_NAME};fi;)
+	@echo downloading db...
+	@wget -P ./db_cache/ https://s3.amazonaws.com/public-clean-saved-db-states/${DB_FILE_NAME}
+	@echo "DB ${DB_FILE_NAME} is now up to date. You can load it by running 'make dbload GZ_FILE=./db_cache/${DB_FILE_NAME}'."
```
mc-nathan: Not a blocker, but this makes it always a two step process.  I'd add another target, say load-remote-db that is something like:
```
load-remote_db: fetch-remote-db load-db
```
and then tweak load-db so it uses ${DB_CACHE_DIR}${DB_FILE_NAME} is ${GZ_FILE} is not set.

Makefile
```diff
@@ -171,18 +180,36 @@ lint:
...
+	$(error $(load_db_error_msg))
 endif
 	@echo "drop database mc_dev; create database mc_dev;" | docker-compose run --rm web ./manage.py dbshell
 	@gzcat $(GZ_FILE) | docker-compose run --rm web ./manage.py dbshell
 	@docker-compose run --rm web ./manage.py migrate --no-input
 
+db_cache:
+	mkdir -p ${DB_CACHE_DIR}
+
+fetch-remote-db: db_cache
```
mc-nathan: db_cache should be ${DB_CACHE_DIR} her and in the rule above.

mc-nathan (approved):

knittingarch (approved):
Code look good and it works as expected.

PR #34, opened 2017-09-07
AC-4774: Add request tracking to impact api

pbx:
## [AC-4774](https://masschallenge.atlassian.net/browse/AC-4774)

**Testing instructions**

* Start local instance: `make build && make dev`
* Make sure you have a user, then log in. (See [quickstart](https://github.com/masschallenge/impact-api/blob/development/QUICK_START.md) steps 6 and 7)
* Pull up the [API request log admin page](http://localhost:8000/admin/rest_framework_tracking/apirequestlog/).
* Using your browser, hit some DRF [low level](http://localhost:8000/api/impact/) calls
* Add [this Postman collection](https://www.getpostman.com/collections/b9dcecd3190fcd10888d) to your local Postman setup (via Collection > Import > Imporrt From Link)
* Using Postman, hit some v0 calls (e.g. "startup list")
* Using Postman, hit some v1 calls (e.g. "user list get")
* Confirm that all of the the above accesses show up in the log and have the expected information.

**NOTE:** The proxy views are not included in this work because of a crashing bug discovered in them. A new ticket(s) has been created for fixing this, under the "if this gets hairy" clause of this ticket's "DONE".


mc-nathan:
Out of curiosity what do you mean by "straight, DRF, v0, and v1"?  I think of there as being the low-level API (DRF?), v0 and v1.  If I'm right about what you mean by "DRF", I'm not sure what you mean by "straight".  Also from a testing standpoint it's work pointing out that reviewers should try going both through the browsable API and the Postman collection.

mc-nathan:
Functional issue: http://localhost:8000/api/v1/user/ is not getting logged.

mc-nathan:
Blocking Functional Problems:
The following calls did not create entries in the log:
GET http://localhost:8000/api/v1/user/
GET http://localhost:8000/api/v1/organization/16576/users/
GET http://localhost:8000/api/v1/user/33/organizations/
POST http://localhost:8000/oauth/token/  - I could see this one being hard, so we can punt on this if necessary, but it would sure be nice data to have.
POST http://localhost:8000/api/v1/user/
PATCH http://localhost:8000/api/v1/user/{{random_user_id}}/ - random_user_id is the id created by the POST
POST http://localhost:8000/api/v0/job_posting_list/ - Interestingly job posting detail is logged

The Postman collection I'm using for the above is:
https://www.getpostman.com/collections/b9dcecd3190fcd10888d

Non-Blocking Functional Issues:
1) Parameters in PATCH and POST calls not shown in admin display.  For example,
from Postman, I ran http://localhost:8000/api/impact/startup/10001/ with a PATCH that successfully changed the short_pitch.  However, the log interface does not show what was patched.  However, looking in the database, the data is there in the "data" field.  It would be nice if the data field was available in the admin ui, but I don't consider this blocking.

2) Reloading the admin display got pretty slow (~10 seconds) which was surprising and worth checking when this goes out into the world

mc-nathan (changes requested):
The code so far is pretty simple, but there are a lot of views that aren't yet using the mechanism (see functional problems) and there may be some cases that are going to be tricky.

pbx:
Thanks for the catches - "straight" was a typo, I did indeed just mean the low-level ("API Root") calls plus v0 and v1.

mc-nathan:
Unlogged URL not in the Postman collection:
http://localhost:8000/api/accelerator/

mc-nathan:
It would be worthwhile having at least one automated test that ensures that a log entry is created.

pbx:
Re `/api/accelerator/` not being logged -- "API root" views (lists of endpoints) don't get logged, but accesses to http://localhost:8000/api/accelerator/Currency/ do get logged.

pbx:
@nathan - I've fixed all the blockers you identified and added a basic test. I'm going to sping new tickets for 1) logging `/oauth/token/` and 2) surfacing parameter info on PATCH and POST calls.

knittingarch:
Testers will also need to load a db before creating a superuser.

mc-nathan:
One missing line of coverage:
impact/views/general_view_set.py      18      1    94%
screen shot 2017-09-11 at 1 12 52 pm

pbx:
Re coverage of `GeneralViewSet.get_related_model_fields` - after finding no references to that method in our code nor in the DRF code, and getting confirmation from Shankar that it's a leftover from the first version of impact-api, I'm solving the coverage problem by removing the method.

knittingarch (changes requested):
Functional Testing: Looks good.
Code: TBD
Code Quality: Clean.
Tests: Pass.
Coverage: 1 missing line (see below)

Coverage: impact/views/general_view_set.py      18      1    94%

mc-nathan:
Everything is now covered.  Approving.

mc-nathan (approved):

knittingarch (approved):

knittingarch:
Everything looks good!

PR #35, opened 2017-09-15
AC-4836 - Expand support for django-accelerator objects in impact-api

shanbady:
to test:
- checkout this branch
- load up the initial schema from the latest accelerate app via ~```make loaddb```~ **Y.M. -  Updated command is now `make load-db` or `make load-remote-db`.**
- visit the django-accelerate app's root view at localhost:8000/api/accelerator/
- verify that everything is functional and that objects can be created, updated, deleted just as with the impact objects. Also verify that OPTIONS calls return a schema for a given object.


web/impact/impact/tests/test_accelerator_routes.py
```diff
@@ -0,0 +1,174 @@
...
+                "user": perm_user.id,
+                "primary_industry": industry.id
+            }, **view_kwargs)
+            self.assertTrue(
+                Startup.objects.filter(organization=org).exists())
+
+    def test_api_object_list(self):
+        StartupFactory(is_visible=1)
+        StartupFactory(is_visible=1)
+        StartupFactory(is_visible=1)
```
yotammanor: create_batch maybe?

web/impact/impact/tests/test_accelerator_routes.py
```diff
@@ -0,0 +1,174 @@
...
+            codename='view_startup',
+        )
+        perm_user = self.make_user(
+            'perm_user2@test.com', perms=["mc.view_startup"])
+        perm = PermissionFactory.create(codename='change_startup')
+        view_perm = PermissionFactory.create(codename='view_startup')
+        perm_user.user_permissions.add(perm)
+        perm_user.user_permissions.add(startup_permission)
+        perm_user.user_permissions.add(view_perm)
+        perm_user.save()
```
yotammanor: seems like there's an entire section here that's repeated in the test above, and can be extracted and reused (and may also be abstracted better, since there're a lot of internal organs flying around at the moment.)

web/impact/impact/tests/test_accelerator_routes.py
```diff
@@ -0,0 +1,174 @@
...
+            self.response_403(self.get(url_name, **view_kwargs))
+
+        perm_user = self.make_user(
+            'perm_user@test.com', perms=["mc.view_startup"])
+        perm = PermissionFactory.create(codename='change_startup')
+        view_perm = PermissionFactory.create(codename='view_startup')
+        perm_user.user_permissions.add(perm)
+        perm_user.user_permissions.add(startup_permission)
+        perm_user.user_permissions.add(view_perm)
+        perm_user.save()
```
yotammanor: likewise, extract and reuse.

web/impact/impact/tests/test_accelerator_routes.py
```diff
@@ -0,0 +1,174 @@
...
+        with self.login(perm_user):
+            response = self.get(url_name, **view_kwargs)
+            self.response_200(response)
+            response_dict = json.loads(response.content)
+            self.assertIn("is_visible", response_dict.keys())
```
yotammanor, line 163: I was expecting a test that involved the `IndustryFactory()` shown in the first act.. Is there a reason it's missing? add an assert or remove the misleading setup.

web/impact/impact/tests/test_accelerator_routes.py
```diff
@@ -0,0 +1,174 @@
...
+            url.name != 'api-root')]
+        url_count = len(urls)
+        accelerate_models = [
+            model for model in apps.get_models('accelerator') if (
+                model._meta.app_label == 'accelerator' and hasattr(
+                    model, "Meta"))]
+        model_count = len(accelerate_models)
+        # we should have two detail and list views per model
+        self.assertEquals(
+            model_count * 4,
```
yotammanor: magic number. Even though explained in a comment above, still worth improving I believe.

web/impact/impact/tests/test_accelerator_routes.py
```diff
@@ -0,0 +1,174 @@
...
+            content_type=ContentType.objects.get(
+                app_label='mc',
+                model='startup'),
+            codename='view_startup',
+        )
+        perm_user.user_permissions.add(perm)
+        perm_user.user_permissions.add(startup_add_permission)
+        perm_user.user_permissions.add(startup_permission)
+        perm_user.user_permissions.add(view_perm)
+        perm_user.save()
```
yotammanor: I find it very weird that Permission defined on `mc.startup` controls the permissions for `accelerator.startup`. This is a major blocker in my view.

yotammanor (changes requested):
Tests, fpdiff, coverage are all good. Endpoints behave as specified, except for one problem (this is actually a django-accelerator bug, exposed):

- When creating a Startup with `organization: null`, JobPosting GET (list or detail) fails to load. 
  - Steps to reproduce:
    - Create an Industry.
    - Create a Startup without an Organization object.
    - go to http://localhost:8000/api/accelerator/JobPosting/ - This fails badly.
  - Likely cause: `accelerator.Startup.organization(null=True)` + `accelerator.Startup.__str__() = self.organization.name`.


Functional issues:


 - There are now many models that are not displayed in the browsable API's root view, which defaults to impact. Once we move all models to accelerator this problem will take the form of whether the root displays low-level API endpoints or not, but until then, the split is misleading.

- The browsable-API's breadcrumbs' texts are incorrect. On http://localhost:8000/api/accelerator/Industry/3/ the breadcrumbs are:
image - but the links are correct.

Also, a few comments below.

pbx (changes requested):
The new models do show up at `/api/accelerator/`. I can create new objects via POST and they do get listd at the object root e.g. `http://localhost:8000/api/accelerator/Currency/`. However, if I try to view the object detail, I get a content-type related exception: 
```
File "/wwwroot/impact/permissions.py" in get_field_level_perms
  62.         ct = ContentType.objects.get(app_label=app_label, model=model_name)
...
Exception Type: DoesNotExist at /api/accelerator/Currency/1/
Exception Value: ContentType matching query does not exist.
```
This may be because the accelerate database dump I loaded was from September 5th.  However, I'd think we'd want to be robust against that, e.g. by using `ContentType.objects.get_for_model()` (which I believe will create the CT if it doesn't exist).

web/impact/impact/tests/test_accelerator_routes.py
```diff
@@ -0,0 +1,174 @@
...
+            app_label='mc',
+            model__in=[
+                'startup',
+                'organization',
+                'programrole',
+                'recommendationtag']).delete()
+
+    def test_user_with_permissions_can_create_objects(self):
+        url_name = "object-list"
+        StartupStatusFactory(id=1)
```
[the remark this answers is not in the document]
shanbady: good call. removed

web/impact/impact/tests/test_accelerator_routes.py
```diff
@@ -0,0 +1,174 @@
...
+        self.response_401(self.get(url_name, **view_kwargs))
+        startup_add_permission, _ = Permission.objects.get_or_create(
+            content_type=ContentType.objects.get(
+                app_label='mc',
+                model='startup'),
+            codename='add_startup',
+        )
+        perm_user = self.make_user(
+            'perm_user2@test.com', perms=["mc.add_startup"])
+        perm = PermissionFactory.create(codename='change_startup')
```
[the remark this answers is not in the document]
shanbady: no the codename is the component after the '.' in the permission. 'mc.change_startup' - it notes the model for which the permission is for.

web/impact/impact/tests/test_accelerator_routes.py
```diff
@@ -0,0 +1,174 @@
...
+        }
+        self.response_401(self.get(url_name, **view_kwargs))
+        startup_add_permission, _ = Permission.objects.get_or_create(
+            content_type=ContentType.objects.get(
+                app_label='mc',
+                model='startup'),
+            codename='add_startup',
+        )
+        perm_user = self.make_user(
+            'perm_user2@test.com', perms=["mc.add_startup"])
```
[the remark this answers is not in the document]
shanbady: didnt appear to be necessary - removed

web/impact/impact/tests/test_accelerator_routes.py
shanbady: fixed

web/impact/impact/tests/test_accelerator_routes.py
shanbady: done

web/impact/impact/tests/test_accelerator_routes.py
shanbady: refactored into its own utility method and got rid of repetition

web/impact/impact/tests/test_accelerator_routes.py
shanbady: added an assert to lock this down - the industryfactory was needed because primary industry for startup is a required field that needs a value.

web/impact/impact/tests/test_accelerator_routes.py
shanbady: this will unfortunately need to be the case for now due to how internally we reference the models and the fact that they are unmanaged. the 'mc.startup' is in reference to the table name from which the models are derived from. I'm not sure how to get around this limitation and be able to use unmanaged models atop existing data.

shanbady:
@pbx when i first switched to using get_for_model i got this error: `AttributeError: 'str' object has no attribute '_meta'` which made me realize i needed to use django.apps.apps.get_model to get a model instance to feed to this method. 

Doing so started triggering this error `LookupError: No installed app with label 'mc'.` since the contentypes we need are for the 'mc' app (but we have not 'mc' app in impact  - I then special-cased it so that anytime an app name of 'mc' was specified, it would do the 'get_model' using the 'impact' app. This seemingly worked for all models except for the User model `LookupError: App 'impact' doesn't have a 'user' model.` I could also put in an if conditional for the User model but this seems wrong and would likely break when connected to an actual accelerate database. It seems like there is a tradeoff to consider here one way or the other so i'm not sure which way will cause us less issues down the road.

web/impact/impact/tests/factories/accelerator/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .startup_factory import StartupFactory
+from .currency_factory import CurrencyFactory
+from .industry_factory import IndustryFactory
+from .organization_factory import OrganizationFactory
+from .recommendationtag_factory import RecommendationTagFactory
+from .jobposting_factory import JobPostingFactory
```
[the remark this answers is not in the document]
shanbady: thats a good question - my view was that the accelerate system considers them to be totally separate objects (and on initialization of the actual models I'm sure there are some minute internal differences like references to app_name etc) and so should have their own factories. here. open to potentially reusing the impact factories for the accelerator app as well.

web/impact/impact/tests/test_accelerator_routes.py
```diff
@@ -0,0 +1,174 @@
...
+from impact.tests.factories import PermissionFactory
+from accelerator.models import Startup
+
+
+class TestAcceleratorRoutes(TestCase):
+    client_class = APIClient
+    user_factory = UserFactory
+
+    @classmethod
+    def setUpClass(cls):
```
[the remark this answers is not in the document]
shanbady: good call. fixed

web/impact/impact/tests/test_accelerator_routes.py
yotammanor: I understand the problem. Then what I'd do is name somewhere `settings.py`-y `ACC_APP_NAME = mc` and use that wherever. Otherwise it's just too confusing.

web/impact/impact/tests/test_accelerator_routes.py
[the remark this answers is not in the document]
yotammanor: From: `perm = PermissionFactory.create(codename='change_startup')`
Change to:`change_perm = PermissionFactory.create(codename='change_startup')`

:)

web/impact/impact/tests/test_accelerator_routes.py
yotammanor: A Context would've also worked here well, but this is a reasonable alternative for now.

web/impact/impact/tests/factories/accelerator/__init__.py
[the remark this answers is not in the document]
yotammanor: I don't entirely understand you answer here. But generally speaking, if there's a dilemma on an approach to take, then feel free to raise it up on slack or after a standup. If you're asking my opinion, then I don't currently see a compelling reason to not put our factories in  django-accelerator, since I don't see how cases where we'd _want_ them to be diverge across code-bases. Any discrepancy that we have now I'm seeing as a temporary step, which we should not extend. Obviously, if you see such reasons to keep them apart, I'd love to hear them.

web/impact/impact/tests/test_accelerator_routes.py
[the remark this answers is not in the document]
shanbady: i see - in any case this has been refactored out.

web/impact/impact/tests/test_accelerator_routes.py
shanbady: moved into a constant for now

shanbady:
Decided to spin out a bug ticket (AC-4886) to address the JobPosting api issue yotam pointed out.

yotammanor:
@pbx  @shanbady so just to recap on the functional issue you're discussing - If I'll load an up to date version if `initial_schema`, this problem goes away?

yotammanor (changes requested):
Just so we're synced. I believe all of my requests have been either answered or tickets with follow on were created. style and coverage are good.

The only blocking functional issue I see left is the error that you get when going to http://localhost:8000/api/impact/Currency/ , as @pbx  found. Once this is fixed, we can merge this.

yotammanor (approved):
Re-reviewing the current state, I see that the accelerator/currency is now working as expected, and impact/currency is not available as expected of a removed model. We might want to add a more graceful response for a non existing model - such as impact/Accelerator or impact/foo should probably raise a 404 instead of a server error. But this can definitely be another ticket.

MAW.

pbx:
Unfortunately I'm still seeing this problem -- trying to GET any Currency (e.g. [USD](http://localhost:8000/api/accelerator/Currency/1/)) throws a `ContentType matching query does not exist` error. 

I'm on AC-4836 at b67ea6d, and did  fresh `load-remote-db`.

This problem does _not_ happen with other CTs I checked, e.g. Industry or Startup. I was able to create and view those through the API.

Going to try a fresh `make build`.

pbx (approved):
Fresh build solved my problem. Good to go.

PR #36, opened 2017-09-20
AC-4835: API improvements for BlackIron v1

mc-nathan:
[Ticket AC-4835](https://masschallenge.atlassian.net/browse/AC-4835)

Work done:
- Expanded /v1/users
- Expanded /v1/organizations
- Added program and cycle information to user history and organization history calls
- Added admin and primary contact information to user organizations and organization users calls
- Added /v1/industry
- Added industry data to /v1/users and /v1/organizations

Note that there is an explicit focus on just the GET and OPTIONS versions of the calls.  POST and PATCH have been deferred to [AC-4875](https://masschallenge.atlassian.net/browse/AC-4875).

To test:
- Review the output of the /v1/users, /v1/organizations, /v1/user/<id>/history, /v1/organization/<id>/history, /v1/user/<id>/organizations, /v1/organization/<id>/users, /v1/industry GET calls either through the browsable API or through Postman.
- Check the OPTIONS for these GET calls and confirm that they pretty much make sense.  Note that the OPTIONS for the list calls return a description of members of the list not a description of the list.  This will be addressed in [AC-4855](https://masschallenge.atlassian.net/browse/AC-4855).


pbx (changes requested):
This all looks good. I interactively checked out the new and updated calls. OPTIONS looked good. I had a few small detail comments.  Very close to a shipit.

However in evaluating DONE I'm seeing data missing from User detail, e.g.   "Home Program Family" and "Program Families".  Has this been deferred?  Does "Leave the dictionary blank for the items that don't exist" mean omit the key (though we include keys for other empty values)? Hopefully this is just me missing something obvious.

Also a small thing: though it's not code you changed this time, I notice the "history" urlpatterns are the only ones without trailing slashes. I think we should add those for consistency.

web/impact/impact/v1/helpers/organization_helper.py
```diff
@@ -0,0 +1,67 @@
...
+        "name",
+        "url_slug",
+        ]
+    OPTIONAL_KEYS = [
+        "additional_industry_ids",
+        "date_founded",
+        "facebook_url",
+        "full_elevator_pitch",
+        "primary_industry_id",
+        "public_inquiry_email",
```
shanbady, line 18: why are these keys not derived from the "required=true" attribute on the field itself? seems like we'd have to maintain this in 2 or 3 places now (impact->django-accelerator->accelerate and here)

web/impact/impact/v1/helpers/organization_helper.py
```diff
@@ -0,0 +1,67 @@
...
+        "short_pitch",
+        "twitter_handle",
+        "video_elevator_pitch_url",
+        "website_url",
+        ]
+    INPUT_KEYS = REQUIRED_KEYS + OPTIONAL_KEYS
+    READ_ONLY_KEYS = [
+        "id",
+        "is_partner",
+        "is_startup",
```
shanbady, line 33: again - we should probably be inferencing as many readonly fields as we can directly from the model rather than hardcoding lists of keys

web/impact/impact/v1/helpers/organization_helper.py
```diff
@@ -0,0 +1,67 @@
...
+    def is_startup(self):
+        return self.startup is not None
+
+    @property
+    def is_partner(self):
+        return self.partner is not None
+
+    @property
+    def additional_industry_ids(self):
+        if self.startup:
```
shanbady, line 53: should we consider returning an empy list as a default?

web/impact/impact/v1/helpers/organization_helper.py
```diff
@@ -0,0 +1,67 @@
...
+            if categories.exists():
+                return list(categories.values_list("id", flat=True))
+
+    def field_value(self, field):
+        result = super(OrganizationHelper, self).field_value(field)
+        if result is not None:
+            return result
+        if self.startup:
+            result = getattr(self.startup, field, None)
+        if result is not None:
```
shanbady: it seems like you can get rid of the redundant checks for ```result is not None``` by using elifs:

```
if result:
    return result
elif self.startup:
    result = getattr(self.startup, field, None)
elif self.partner:
    result = getattr(self.partner, field, None)
return result
```

web/impact/impact/v1/helpers/profile_helper.py
```diff
@@ -0,0 +1,82 @@
+from impact.v1.helpers.model_helper import ModelHelper
+
+VALID_GENDERS = ["f", "m", "o", "p"]
```
shanbady: would be less redundant if ```VALID_GENDERS = GENDER_TRANSLATIONS.values()```

web/impact/impact/v1/helpers/profile_helper.py
```diff
@@ -0,0 +1,82 @@
...
+        "phone",
+        "referred_by",
+        "speaker_topics",
+        "title",
+        "twitter_handle",
+        ]
+    OPTIONAL_BOOLEAN_KEYS = [
+        "judge_interest",
+        "mentor_interest",
+        "office_hours_interest",
```
shanbady, line 37: could we gather these by inspecting the fields themselves?

web/impact/impact/v1/helpers/profile_helper.py
```diff
@@ -0,0 +1,82 @@
...
+    @property
+    def mentoring_specialties(self):
+        if hasattr(self.subject, "mentoring_specialties"):
+            specialties = self.subject.mentoring_specialties
+            if specialties:
+                return [specialty.name for specialty in specialties.all()]
+
+
+def find_gender(gender):
+    if not hasattr(gender, "lower"):
```
shanbady: shouldnt we be checking that its a string instead?
```if type(gender) == str```

File: web/impact/impact/v1/helpers/profile_helper.py [hunk 8]
```diff
@@ -0,0 +1,82 @@
...
+        return None
+    gender = GENDER_TRANSLATIONS.get(gender.lower(), gender)
+    if gender in VALID_GENDERS:
```
shanbady: are we missing 'unknown' from VALID_GENDERS?

web/impact/impact/v1/helpers/profile_helper.py
```diff
@@ -0,0 +1,82 @@
...
+        return gender
+    return None
```
shanbady, line 82: should we consider returning 'unkown' here since we have that as an option on accelerate.

web/impact/impact/v1/views/utils.py
```diff
@@ -0,0 +1,15 @@
+def merge_data_by_id(data):
+    result = {}
+    for datum in data:
+        id = datum["id"]
```
shanbady: using builtin keyword "id" - rename to _id

shanbady (changes requested):
overall functions as described. I'm a little worried about maintaining the schema and fields since it seems that these are being less and less autogenerated from the model definitions themselves

mc-nathan:
@pbx Regarding you over arching comment:

'However in evaluating DONE I'm seeing data missing from User detail, e.g. "Home Program Family" and "Program Families". Has this been deferred?'

**This is a time-boxed ticket and the ProgramFamily information has been explicitly deferred to ticket AC-4875 as stated in the comments for AC-4835.**

'Does "Leave the dictionary blank for the items that don't exist" mean omit the key (though we include keys for other empty values)? Hopefully this is just me missing something obvious.'

**I believe we only include keys for values that are not None.  False and "" are not None so they are included.  The alternatives are to always have every field and have a None value or to add a mechanism to identify optional fields.  The history events are the clear example of why this is important.  It seems non-sensical to have program or cycle fields for events that have no associated program or cycle (e.g., user creation) especially since some event, such as "became entrant" potentially have multiple programs and hence have a separate field.  In my view it is better to be simple and lean and just not mention fields that have no value.**

"Also a small thing: though it's not code you changed this time, I notice the "history" urlpatterns are the only ones without trailing slashes. I think we should add those for consistency."

**Good catch.  Fixed.**

web/impact/impact/tests/contexts/user_context.py
```diff
@@ -5,21 +5,30 @@
...
             self.profile = EntrepreneurProfileFactory(user=user)
             user.entrepreneurprofile = self.profile
         elif user_type == "EXPERT":
-            self.profile = ExpertProfileFactory(user=self.user)
+            self.primary_industry = primary_industry or IndustryFactory()
+            self.additional_industries = additional_industries or []
+            self.profile = ExpertProfileFactory(
+                user=self.user,
+                primary_industry_expertise=self.primary_industry,
+                additional_industry_expertise=additional_industries)
```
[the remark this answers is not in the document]
mc-nathan: Good catch. Done.

web/impact/impact/tests/test_organization_detail_view.py
```diff
@@ -26,6 +27,20 @@ def test_get_startup(self):
             assert response.data["is_startup"] is True
             assert response.data["is_partner"] is False
 
+    def test_get_startup_with_industries(self):
+        primary_industry = IndustryFactory()
+        additional_industries = IndustryFactory.create_batch(2)
+        startup = StartupFactory(
+            primary_industry=primary_industry,
+            additional_industry_categories=additional_industries)
```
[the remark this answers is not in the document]
mc-nathan: Outside the scope of this ticket, but there's already a ticket for this, [AC-4880](https://masschallenge.atlassian.net/browse/AC-4880).

web/impact/impact/v1/helpers/organization_helper.py
mc-nathan: See comment at the end.

web/impact/impact/v1/helpers/organization_helper.py
mc-nathan: See comment at the end.

web/impact/impact/v1/helpers/organization_helper.py
mc-nathan: I could see this for Organizations that have Startups, but not for ones that only have Partners (since there really is no additional industry list.  However, given that it can be missing, I don't see the benefit of returning an empty list some of the time.  Won't fix unless I hear a clear argument why.

web/impact/impact/v1/helpers/organization_helper.py
mc-nathan: I wasn't thrilled by this code either.  I noodled it a bit more and came up with something I like a bit more.

Your proposed code has two problems.  First, it is important that we check against None versus falsey since some fields can have a valid value of False or "".  Second, your proposed elifs don't work in the case where an Organization has both a startup and a partner (currently 303 cases).

web/impact/impact/v1/helpers/profile_helper.py
mc-nathan: Fair.  Done.

web/impact/impact/v1/helpers/profile_helper.py
mc-nathan: See comment at the end.

web/impact/impact/v1/helpers/profile_helper.py
mc-nathan: This seems like a bit of philosophical question.  My check is explicitly checking that the next line will succeed while accepting as rich a range of potential input as possible.  Your approach restricts the type of the argument to strings which currently are the only values we pass in, so practically there is no different. I would argue that my approach is more future proof.  In any case if we're going your way then really we should be doing isinstance(gender, str).  Obviously I have a slight preference for my approach, but I'm willing to make the change to isinstance if it's important to you.

web/impact/impact/v1/helpers/profile_helper.py
mc-nathan: 'unknown' is in a bit of a gray area.  It is the value that we used to backfill data that we didn't have, but in accelerate we prevent users from choosing that value (they can use "Prefer Not to State" instead).  Since find_gender is only expected to be used in PATCH and POST calls the current code is consistent with accelerate.

web/impact/impact/v1/helpers/profile_helper.py
mc-nathan: Not in my opinion.  None is a better signal of an invalid value.  Some gender information is considered required in the accelerate platform and I don't see that we want to change that.

web/impact/impact/v1/views/utils.py
mc-nathan: Good point.  Ended up renaming the function and generalizing `id` to an argument called `merge_field` which defaults to "id".

mc-nathan:
@shanbady I'm going to push back for now on three of your comments that have to do with inferring fields for the API based on the underlying model classes.  The follow on ticket, [AC-4875](https://masschallenge.atlassian.net/browse/AC-4875), has a goal of getting rid of metadata.py and moving that information into the helper classes.  That seems like the right time to improve the handling of these values.  I think we agree that we want to reduce the redundancy and where possible have information represented in only one place.

There is a good conversation that we should have about the inference question.  I believe the root question is how do we want to handle exceptions.  There are definitely fields that we want to add, e.g. `is_startup` and `is_partner` to Organization, and there are fields we don't want to go out through the API without a higher level of permission, e.g., the `internal_notes` field.  This could be handled in the current framework with methods on the helper class that just return None assuming that we continue to simply exclude fields that have a value of None.

With this is in mind, the question essentially comes down to whether we use a white list (which fields do we include) or a black list (which fields do we exclude).  The challenge with a black list is that it has implications for potential API versioning.  If we add a new field to the user profile should that just get added to the API output with no code and no notification or should we plan to change the version of the API?  What about more interesting cases like changing the representation of some of the data say from a single value to a more detailed concept (like we do in this PR with the user/<id>/organizations and organization/<id>/users calls).  If we had active clients relying on the current data representation then no matter what we would have to have a mechanism for delivering the old data. If we are using a black list method we have to remember to black list any new fields that are used in the new representation in the old versions or we risk creating weird in between states where the old information is presented inter-mixed with some of the new information.

For now I'm going to stick with the white list approach, but I'd be happy to talk about it.

web/impact/impact/v1/helpers/profile_helper.py
shanbady: I'm going to be a stickler on the string check because I think the context for the next set of instructions is reliant on it being a string and not just having a ```lower``` method defined - Also curious as to why this was added; are there places we expect this to be called with non-string inputs?

mc-nathan: It's defensive coding habit because we have no control over what we might get from the user.

mc-nathan: Done.

shanbady (approved):

pbx (approved):

PR #39, opened 2017-09-25
[AC-4897] "count" field in list views should reflect size of full result set

mc-nathan:
To test:
- resetremotedb (`make load-remote-db DB_FILE_NAME=initial_schema.sql.gz`)
- Go to http://localhost:8000/api/v1/organization/?limit=5
- On development the count field is 5, but on the branch it's 20247
- Try the same thing with http://localhost:8000/api/v1/user/?limit=5
- On development the count field is 5, but on the branch it's 46186

### DIFF
--- a/web/impact/impact/tests/test_user_list_view.py
+++ b/web/impact/impact/tests/test_user_list_view.py
@@ -32,10 +32,25 @@ class TestUserListView(APITestCase):
         user2 = UserContext().user
         with self.login(username=self.basic_user().username):
             url = reverse("user")
+            user_count = User.objects.count()
             response = self.client.get(url)
+            results = response.data["results"]
+            assert user_count == min(len(results), 10)
             emails = [result["email"] for result in response.data["results"]]
             assert user1.email in emails
             assert user2.email in emails
+            assert user_count == response.data["count"]
+
+    def test_get_with_limit(self):
+        UserContext().user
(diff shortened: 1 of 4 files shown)


shanbady (approved):
looks good. MAW

shanbady (approved):

pbx (approved):

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -32,10 +32,25 @@ def test_get(self):
...
+            results = response.data["results"]
+            assert user_count == min(len(results), 10)
             emails = [result["email"] for result in response.data["results"]]
             assert user1.email in emails
             assert user2.email in emails
+            assert user_count == response.data["count"]
+
+    def test_get_with_limit(self):
+        UserContext().user
+        UserContext().user
```
pbx, line 46: Are these two lines doing something that one `UserContext()` wouldn't do alone?

shanbady: presumably this is so that we have 2 users total but want to limit the api call to 1 for comparison

pbx: Ah, ok, thanks.  The `.user` seems superfluous to me, and we could use `create_batch` instead, but it works.

PR #40, opened 2017-09-25
[AC-4899] Fix date filters for list views

mc-nathan:
Work done:
- Switched from update_at__gt and update_at__lt to update_at.after and update_at.before.
- Ensure that the expected results are right (per ticket).

To test:
On the branch with URLs like http://localhost:8000/api/v1/user/?update_at.after=<date>, http://localhost:8000/api/v1/organization/?update_at.before=<date> until you are satisfied that they are return reasonable sets.  In most cases the updated_at.after and update_at.before sets for a given resource should be distinct unless there happens to be an element right on the date given.  Resources that for historical reasons have updated_at set to None should end up in the .before set.

Note: The industry resource does not work since it does not yet have an updated_at date.

### DIFF
--- a/web/impact/impact/tests/test_user_list_view.py
+++ b/web/impact/impact/tests/test_user_list_view.py
@@ -1,20 +1,20 @@
 # MIT License
 # Copyright (c) 2017 MassChallenge, Inc.
 
+import pytz
+import datetime
+
 from django.urls import reverse
 from django.contrib.auth import get_user_model
-from impact.models import (
-    EntrepreneurProfile,
-    ExpertProfile,
-    MemberProfile
-)
-import pytz
+
 from impact.tests.factories import StartupTeamMemberFactory
 from impact.tests.contexts import UserContext
 from impact.tests.api_test_case import APITestCase
-import simplejson as json
+from impact.utils import (
+    get_profile,
+    override_updated_at,
+)
 from impact.v1.views.user_list_view import EMAIL_EXISTS_ERROR
-import datetime
 
 EXAMPLE_USER = {
     "first_name": "First",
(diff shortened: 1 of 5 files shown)


web/impact/impact/utils.py
```diff
@@ -12,6 +13,15 @@
         "2010-01-01T00:00:00Z",
         "%Y-%m-%dT%H:%M:%SZ"))  # format based on browsable API output
 
+UPDATE_STATEMENT = "UPDATE {table} SET updated_at = %s WHERE id = %s"
+
+
+def override_updated_at(model, value):
```
shanbady: interesting - as long as this is here only for tests. I think we should leave the actual field as readonly

web/impact/impact/utils.py
```diff
@@ -12,6 +13,15 @@
...
+    # Yes, this big of stick is needed to change the update_at field
+    with connection.cursor() as cursor:
+        sql = UPDATE_STATEMENT.format(table=model._meta.db_table)
```
shanbady, line 22: interesting - as long as this is here only for tests. I think we should leave the actual field as readonly

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -84,51 +84,50 @@ def test_post_with_existing_email(self):
...
-        user3 = UserContext().user
-        StartupTeamMemberFactory(user=user)
-        StartupTeamMemberFactory(user=user2)
-        StartupTeamMemberFactory(user=user3)
-        response = ""
-        lastweek = datetime.datetime.now(pytz.utc) - datetime.timedelta(days=7)
-        EntrepreneurProfile.objects.filter(user__id=user.id).update(
-            updated_at=lastweek)
+    def test_updated_at_before_datetime_filter(self):
+        updated_none = _user_for_date(None)
```
shanbady, line 88: good to lock down tests for the "none" case

web/impact/impact/v1/views/base_list_view.py
```diff
@@ -42,31 +42,27 @@ def get(self, request):
...
-        queryset.filter(
-            Q(updated_at__isnull=False)
-        ).exclude(
-            Q(updated_at__lte=updated_at_gt)
-        )
+def _filter_by_date(queryset, updated_at_after, updated_at_before):
+    updated_at_after = parse_date(updated_at_after)
+    updated_at_before = parse_date(updated_at_before)
+    if updated_at_after:
+        queryset = queryset.filter(Q(updated_at__gte=updated_at_after))
```
shanbady, line 63: interesting that we were able to drop the need for including elements with isnull=False - The reason I initially included them was due to this advisory note in the django docs about how different databases may return different results if you don't filter out null values for that field( https://docs.djangoproject.com/en/1.11/ref/models/querysets/ if you search for "earliest() and latest() may return instances with null dates.")

shanbady (changes requested):
nothing too much of a blocker other than the concern i express with removing the null filters

web/impact/impact/utils.py
mc-nathan: It pretty much is already.  They are marked as allowing None/NULL which they have to be due to the old data, but it's a pain in  the butt to get them to actually be any value other than what Django wants them to be so they are effectively read-only.

web/impact/impact/v1/views/base_list_view.py
mc-nathan: That seems to only be about ordering which is not relevant here.  We are already assuming in the query that the value not existing is causing it to get filtered out in the .after case due to the different types of user profiles so I think things would fail pretty hard in the tests or live if this didn't work and I don't see any evidence of either problem currently.  If we were to switch databases then we might have to address this then, but I would guess not (again since the warning is about ordering not filter/exclude).  Also we specifically don't want to filter out the null values from the .before case.  Finally, the old code wasn't do any of this right anyway. :-)

shanbady (approved):

knittingarch (approved):
Looks good functionally and code wise.

PR #42, opened 2017-09-26
Ac 4875: GET API Calls for Black Iron

mc-nathan:
Work done:
- Improves handling of OPTIONS calls by localizing the data in the helper objects.  More redundancy could be removed here, but it would be better to wait until we have decided on a standard per AC-4855.
- Provides read-only Program Family Resource (/api/v1/program_family)
- Provides read-only Confidential User Call (/api/v1/user/<id>/confidential) which is managed by a separate permission.
- Added Judging Round id and name information to User History call
- Added Newsletter events to User History call

To test:
- Program Family Resource:
-- Go to http://localhost:8000/api/v1/program_family/, check the OPTIONS, pick an "id" and go to http://localhost:8000/api/v1/program_family/<id>

- Confidential User Information:
-- Visit http://localhost:8000/api/v1/user/182/confidential/ and see that you get a permission denied warning.
-- Give your user the appropriate Django permission:
```
HQ-Dev-8:impact-api nathan$ make shell
Starting impactapi_mysql_1 ... done
Starting impactapi_redis_1 ... done
Python 3.6.2 (default, Jul 24 2017, 19:47:39) 
Type "copyright", "credits" or "license" for more information.

IPython 4.0.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import impact.models as im

In [2]: user = im.User.objects.get(id=13514)

In [3]: from django.contrib.auth.models import Group

In [4]: from django.conf import settings

In [5]: group = Group.objects.create(name=settings.V1_CONFIDENTIAL_API_GROUP)

In [6]: user.groups.add(group)

In [7]: user.save()
```
-- See this http://localhost:8000/api/v1/user/182/confidential/ succeed
-- Note that there is currently only information related to experts in this call.

- Judging Round information in User History call
-- Find a user who has been a confirmed judge, say the user with id=182
-- Make the appropriate User History call, e.g., http://localhost:8000/api/v1/user/182/history/
-- Search for the string "judging_round_name"

- Newsletter Events in User History call
-- On the above page search for "newsletter_name"

- Make sure the OPTIONS calls are reasonable for the above GET calls.

### DIFF
--- a/web/impact/impact/permissions.py
+++ b/web/impact/impact/permissions.py
@@ -33,6 +33,15 @@ class V1APIPermissions(BasePermission):
             name=settings.V1_API_GROUP).exists()
 
 
+class V1ConfidentialAPIPermissions(BasePermission):
+
+    authenticated_users_only = True
+
+    def has_permission(self, request, view):
+        return request.user.groups.filter(
+            name=settings.V1_CONFIDENTIAL_API_GROUP).exists()
+
+
 class DynamicModelPermissions(BasePermission):
     perms_map = {
         'GET': ['%(app)s.view_%(model_name)s'],
--- a/web/impact/impact/v1/helpers/industry_helper.py
+++ b/web/impact/impact/v1/helpers/industry_helper.py
@@ -1,10 +1,21 @@
 from impact.models import Industry
 from impact.v1.helpers.model_helper import ModelHelper
+from impact.v1.metadata import (
+    OPTIONAL_ID_TYPE,
+    OPTIONAL_STRING_TYPE,
+    PK_TYPE,
+)
 
 
 class IndustryHelper(ModelHelper):
     MODEL = Industry
 
+    DETAIL_METADATA = {
+        "id": PK_TYPE,
+        "name": OPTIONAL_STRING_TYPE,
+        "full_name": OPTIONAL_STRING_TYPE,
+        "parent_id": OPTIONAL_ID_TYPE,
+        }
     REQUIRED_KEYS = ["name"]
     OPTIONAL_KEYS = ["parent_id"]
     INPUT_KEYS = REQUIRED_KEYS + OPTIONAL_KEYS
--- a/web/impact/impact/v1/helpers/program_family_helper.py
+++ b/web/impact/impact/v1/helpers/program_family_helper.py
@@ -1,7 +1,6 @@
 from impact.models import ProgramFamily
 from impact.v1.helpers.model_helper import ModelHelper
 from impact.v1.metadata import (
-    OPTIONAL_ID_TYPE,
     OPTIONAL_STRING_TYPE,
     PK_TYPE,
 )
--- a/web/impact/impact/v1/helpers/program_family_helper.py
+++ b/web/impact/impact/v1/helpers/program_family_helper.py
@@ -1,0 +1,33 @@
+from impact.models import ProgramFamily
+from impact.v1.helpers.model_helper import ModelHelper
+from impact.v1.metadata import (
+    OPTIONAL_ID_TYPE,
+    OPTIONAL_STRING_TYPE,
+    PK_TYPE,
+)
+
+
+class ProgramFamilyHelper(ModelHelper):
+    MODEL = ProgramFamily
+
+    DETAIL_METADATA = {
+        "id": PK_TYPE,
+        "name": OPTIONAL_STRING_TYPE,
+        "email_domain": OPTIONAL_STRING_TYPE,
+        "phone_number": OPTIONAL_STRING_TYPE,
+        "short_description": OPTIONAL_STRING_TYPE,
+        "url_slug": OPTIONAL_STRING_TYPE,
+        }
+
+    REQUIRED_KEYS = ["name"]
+    OPTIONAL_KEYS = ["email_domain",
+                     "phone_number",
+                     "short_description",
+                     "url_slug"]
+    INPUT_KEYS = REQUIRED_KEYS + OPTIONAL_KEYS
+    READ_ONLY_KEYS = ["id"]
+    OUTPUT_KEYS = READ_ONLY_KEYS + INPUT_KEYS
+
+    @property
+    def full_name(self):
+        return str(self.subject)
--- a/web/impact/impact/v1/helpers/program_family_helper.py
+++ b/web/impact/impact/v1/helpers/program_family_helper.py
@@ -26,7 +26,3 @@ class ProgramFamilyHelper(ModelHelper):
     INPUT_KEYS = REQUIRED_KEYS + OPTIONAL_KEYS
     READ_ONLY_KEYS = ["id"]
     OUTPUT_KEYS = READ_ONLY_KEYS + INPUT_KEYS
-
-    @property
-    def full_name(self):
-        return str(self.subject)
--- a/web/impact/impact/v1/helpers/user_helper.py
+++ b/web/impact/impact/v1/helpers/user_helper.py
@@ -9,6 +9,8 @@ from impact.v1.metadata import (
     OPTIONAL_DATE_TYPE,
     OPTIONAL_LIST_TYPE,
     OPTIONAL_ID_TYPE,
+    PK_TYPE,
+    READ_ONLY_STRING_TYPE,
 )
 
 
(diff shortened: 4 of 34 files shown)


web/impact/impact/v1/helpers/program_family_helper.py
```diff
@@ -0,0 +1,28 @@
...
+    PK_TYPE,
+)
+
+
+class ProgramFamilyHelper(ModelHelper):
+    MODEL = ProgramFamily
+
+    DETAIL_METADATA = {
+        "id": PK_TYPE,
+        "name": OPTIONAL_STRING_TYPE,
```
shanbady: same as before - the name appears as OPTIONAL_STRING_TYPE but is in the required keys dictionary

web/impact/impact/permissions.py
```diff
@@ -33,6 +33,15 @@ def has_permission(self, request, view):
             name=settings.V1_API_GROUP).exists()
 
 
+class V1ConfidentialAPIPermissions(BasePermission):
+
```
shanbady, line 37: what is the purpose of this permission? I'm not seeing what necessitates a new permission here. If our v1 permission class is not robust enough to support read-only type access (I think we should invest in adding that capability instead of splitting out a new group+permission especially considering blackiron is the only one with access to this v1 api). thoughts?

web/impact/impact/v1/views/user_history_view.py
```diff
@@ -22,4 +23,5 @@ class UserHistoryView(BaseHistoryView):
                      UserBecameConfirmedMentorEvent,
                      UserBecameFinalistEvent,
                      UserCreatedEvent,
-                     UserJoinedStartupEvent]
```
shanbady, line 25: I was seeing this odd result when looking at the history api call (note the programs dont line up)
screen shot 2017-09-26 at 1 30 30 pm
not sure if this is a consequence of the data cleaning procedure or not

web/impact/impact/v1/views/user_confidential_view.py
```diff
@@ -0,0 +1,36 @@
...
+
+class UserConfidentialView(APIView):
+    metadata_class = ImpactMetadata
+
+    permission_classes = (
+        V1ConfidentialAPIPermissions,
+    )
+
+    METADATA_ACTIONS = {"GET": {
+            "expert_group": OPTIONAL_STRING_TYPE,
```
shanbady: pep8

web/impact/impact/v1/views/user_detail_view.py
```diff
@@ -29,6 +29,11 @@ class UserDetailView(LoggingMixin, APIView):
     )
     metadata_class = ImpactMetadata
 
+    METADATA_ACTIONS = {
+        "GET": UserHelper.DETAIL_GET_METADATA,
+        "PATCH": UserHelper.DETAIL_PATCH_METADATA,
+        }
```
shanbady: pep8

web/impact/impact/v1/views/user_list_view.py
```diff
@@ -24,6 +24,11 @@
 
 
 class UserListView(BaseListView):
+    METADATA_ACTIONS = {
+        "GET": UserHelper.DETAIL_GET_METADATA,
+        "POST": UserHelper.LIST_POST_METADATA,
+        }
```
shanbady: pep8

web/impact/impact/v1/views/user_organizations_view.py
```diff
@@ -25,6 +28,10 @@ class UserOrganizationsView(LoggingMixin, APIView):
     metadata_class = ImpactMetadata
     model = get_user_model()
 
+    METADATA_ACTIONS = {
+        "GET": {"organizations": READ_ONLY_LIST_TYPE},
```
shanbady: pep8

web/impact/impact/v1/views/organization_detail_view.py
```diff
@@ -20,6 +20,9 @@ class OrganizationDetailView(LoggingMixin, APIView):
     permission_classes = (
         V1APIPermissions,
     )
+    METADATA_ACTIONS = {
+        "GET": OrganizationHelper.DETAIL_METADATA
```
shanbady: pep8

web/impact/impact/v1/helpers/user_helper.py
```diff
@@ -3,11 +3,54 @@
...
+    "gender": OPTIONAL_STRING_TYPE,
+    "phone": OPTIONAL_STRING_TYPE,
+    "additional_industry_ids": OPTIONAL_LIST_TYPE,
+    "expert_category": OPTIONAL_STRING_TYPE,
+    "mentoring_specialties": OPTIONAL_LIST_TYPE,
+    "primary_industry_id": OPTIONAL_ID_TYPE,
+    "updated_at": OPTIONAL_DATE_TYPE,
+}
+USER_POST_OPTIONS.update(dict([
+            (key, OPTIONAL_BOOLEAN_TYPE)
```
shanbady: pep8

web/impact/impact/v1/helpers/industry_helper.py
```diff
@@ -1,10 +1,21 @@
...
+)
 
 
 class IndustryHelper(ModelHelper):
     MODEL = Industry
 
+    DETAIL_METADATA = {
+        "id": PK_TYPE,
+        "name": OPTIONAL_STRING_TYPE,
+        "full_name": OPTIONAL_STRING_TYPE,
```
shanbady: pep8

shanbady (changes requested):

web/impact/impact/v1/views/program_family_detail_view.py
```diff
@@ -0,0 +1,33 @@
...
+class ProgramFamilyDetailView(LoggingMixin, APIView):
+    model = ProgramFamily
+    metadata_class = ImpactMetadata
+
+    permission_classes = (
+        V1APIPermissions,
+    )
+
+    METADATA_ACTIONS = {
+        "GET": ProgramFamilyHelper.DETAIL_METADATA
```
shanbady, line 25: I fee like these metadata constants are a bit all over the place - wont make this a blocker but im seeing that some of the metadatas are coming off helper classes, some only have a DETAIL_METADATA that is used for all views (post/get/list) - it might be worth wrangling this in for consistency.

shanbady (changes requested):

web/impact/impact/permissions.py
mc-nathan: John Harthorne wants the confidential information to be separately controlled from the other calls for future use of the other calls to folks who are not concerned about that data.  This was a late breaking requirement.

web/impact/impact/v1/views/user_history_view.py
mc-nathan: Good catch!  It was using str() rather than .short_name().  Fixed.

web/impact/impact/v1/views/program_family_detail_view.py
mc-nathan: The METADATA_ACTIONS information needs to be findable from the view (see v1.metadata.ImpactMetadata) and it varies based on the view (UserListView has different actions than UserHistoryView).  However, there is information that is shared between views (ProgramFamilyHelper.DETAIL_METADATA is used by both UserListView and UserDetailView).  I don't want to import this data from one view to another, so putting the object detail descriptions in the helper class seems logical to me.  Ideally more of this data would be derived from the object, but that gets into the black list/white list issue and the metadata needed to really implement OPTIONS correctly.  I don't see a way to more cleanly handle this that doesn't involve significant development work that would be done better once we have a standard.

web/impact/impact/permissions.py
shanbady: ok i see - is there any accelerate-side permission that already exists that we could latch onto? If not i suppose creating its own group is a reasonable way of segmenting this. - i dont want this point to block getting this out since it seems to work as-is

mc-nathan: Don't think there's an equivalent accelerate-side permission.

shanbady (approved):

knittingarch (approved):
Looks good to me! Code looks good.  Functional testing works as expected.  Code quality and tests are up to par!

PR #44, opened 2017-09-28
[AC-4642b] - Add Overview Start and End Times to the Program model

yotammanor:
**Changes Introduced [AC-4642b](https://masschallenge.atlassian.net/browse/AC-4642):**

- add new fields to Program.
- remove deprecated field from Program.

**Test:**
- in impact-api, `make dev`
- in accelerate, checkout AC-4642a.
- run `resetdb`
- then dump the db: `mysqldump -u root -proot mc_dev | gzip > /vagrant/db_cache/initial_schema_AC_4642a.sql.gz`
- in impact-api, in a new terminal, run `make load-db GZ_FILE=../accelerate/db_cache/initial_schema_AC_4642a.sql.gz`
- login to browsable API as demoadmin, go to http://localhost:8000/api/impact/Program/
- see that the new fields are added, and that the old field is removed.

**Note:**
- AC-4642b should be deployed in impact-api only after AC-4642a is deployed in accelerate.
- AC-4642c in accelerate should be deployed only after AC-4642b.

### DIFF
--- a/web/impact/impact/models/program.py
+++ b/web/impact/impact/models/program.py
@@ -110,8 +110,6 @@ class Program(MCModel):
         max_length=30,
         default="",
     )
-    # accepting_company_overviews =
-    #  models.BooleanField(default=False)  # todo: remove
     mentor_program_group = models.ForeignKey(
         NamedGroup,
         blank=True,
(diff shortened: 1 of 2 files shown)


jonkiparsky (changes requested):
This change does not agree with the change on the AC-4642a PR. (accepting_company_overviews field is not deleted there)

yotammanor:
This is on purpose. Impact api needs to change first, before the migration
is applied (which is done in AC-4642c)

pbx (approved):
Works for me. I do think we need a more explicit deployment outline to follow when we're reviewing multi-branch, multi-repo stories like this.

jonkiparsky (approved):

PR #46, opened 2017-10-02
AC-4880 Unify the way we refer to industries in user and organization related objects

pbx:
## [AC-4880](https://masschallenge.atlassian.net/browse/AC-4880)

This ticket changes three repos: accelerate, impact-api, and django-accelerator. 
* [accelerate PR](https://github.com/masschallenge/accelerate/pull/1567)
* [django-accelerator PR](https://github.com/masschallenge/django-accelerator/pull/25)

Testing: 
* **Database setup** - Branch AC-4880 in the accelerate project changes the database schema; impact-api needs that updated database to run this code. In accelerate, check out the AC-4880 branch, run`resetdb` or `resetremotedb`, then run `dumpdb`. Then, in impact-api, run `make load-db GZ_FILE=../accelerate/db_cache/initial_schema.sql.gz`
* Permissions setup - run `make grant-permissions PERMISSION_USER=demoadmin@masschallenge.org PERMISSION_CLASSES=v1_clients` -- then you can log in as demoadmin@ to test.
* Confirm that everything using the affected fields (now known as `primary_industry` and `additional_industries` functions as before.


mc-nathan:
Not a blocker, but I noticed that sql/initial.sql still refers to `primary_industry_expertise_id`.  To fix this I believe would require either smashing or at least squashing the migrations.  Probably worth generating a ticket about that, but not worrying about it for this PR.

mc-nathan (approved):
One minor non-blocker and a suggested follow on ticket.

mc-nathan:
Reviewing this code base for "expertise", I only found two cases and they are both user facing strings that we may not what to correct.  They are in:
```
api.md
web/impact/impact/models/scenario_preference.py
```

knittingarch (changes requested):
After running `load-remote-db`, I get the following error when going to http://127.0.0.1:8000/api/impact/ExpertProfile/: 

OperationalError at /api/impact/ExpertProfile/
(1054, "Unknown column 'mc_expertprofile.primary_industry_id' in 'field list'")

pbx:
@knittingarch - I cant reproduce that failure. Is it persistent for you?

On the branch in impact-api, I ran: `make load-remote-db` and then  `make dev`, logged in with a superuser, then loaded [ExpertProfile](http://127.0.0.1:8000/api/impact/ExpertProfile/), no errors.

(Postscript: problem was that my testing instructions hadn't mentioned updating the database; now corrected, sorry about that.)

knittingarch (approved):

PR #49, opened 2017-10-06
AC-4947: Fix is_authenticated error reported by Travis

mc-nathan:
To test:
- `make clean; make build; make test` works on branch, but not on development


yotammanor (changes requested):
Functionally as described.
Blocking because this branch is on top of AC-4900 and should not be merged before it.

Otherwise, MAW.

PR #50, opened 2017-10-09
[AC-4947] Switch to using user.is_authenticated for DRF 3.7.0

mc-nathan:
To test:
`make clean; make build; make test` works on branch, but not on development

This is a new PR for this ticket that only has the one relevant commit.


yotammanor (approved):
MAW

PR #52, opened 2017-10-13
[AC-4959] Add support for accepting the X-HTTP-Method-Override Header

knittingarch:
**To Test**
Browser Window
- Login to the API (http://localhost:8000/api/impact/) and create a new Client ID.
  - Select Client Type "Public".
  - Select Authorization grant type "Resource owner password-based"
  - Copy the Client ID. 
- Go to http://localhost:8000/api/v1/user/45679/.  Observe the name and email fields.

Postman
- Open up Postman
- Go to Import > Import From Link
  - Use https://www.getpostman.com/collections/e9040bab8bb0b9691a83
- Under localhost > v1, you should see a call called "v1 user patch via post".
- In the body, make changes to the existing fields.
- Click send.
  - You should see a response of 204.

Browser Window 
- Refresh the page.
  - You should see your changes in the fields you altered.


mc-nathan (approved):
Functionality, coverage, tests, and code are all good.  Approved!

web/impact/impact/middleware/method_override_middleware.py
```diff
@@ -0,0 +1,13 @@
...
+METHOD_OVERRIDE_HEADER = 'HTTP_X_HTTP_METHOD_OVERRIDE'
+
+
+class MethodOverrideMiddleware(object):
+    def process_request(self, request):
+        if request.method != 'POST':
+            return
+        if METHOD_OVERRIDE_HEADER not in request.META:
+            return
+        request.method = request.META[METHOD_OVERRIDE_HEADER]
```
shanbady, line 13: I think changewise this all looks good - I think it might be adding a mention of this functionality in the README or quickstart document about how this works.

web/impact/impact/tests/test_method_override_middleware.py
```diff
@@ -0,0 +1,24 @@
...
+from impact.tests.api_test_case import APITestCase
+from impact.tests.contexts import UserContext
+
+
+class TestMethodOverrideMiddleware(APITestCase):
+
+    def test_patch(self):
+        context = UserContext()
+        user = context.user
+        with self.login(username=self.basic_user().username):
```
shanbady: I would rename this test to 'test_patch_as_post' and would also add another test that tests setting HTTP_X_HTTP_METHOD_OVERRIDE to GET

shanbady (changes requested):

web/impact/impact/tests/test_method_override_middleware.py
knittingarch: @shanbady You mean a test that shows it failing if a GET is passed in?

web/impact/impact/tests/test_method_override_middleware.py
```diff
@@ -0,0 +1,37 @@
...
+            url = reverse("user_detail", args=[user.id])
+            new_first_name = "David"
+            self.client.post(
+                url,
+                **{METHOD_OVERRIDE_HEADER: "PATCH"},
+                data={"first_name": new_first_name})
+            user.refresh_from_db()
+            assert user.full_name == new_first_name
+
+    def test_patch_via_get_makes_no_change(self):
```
shanbady, line 26: I would make this test that the "patch via override header" gives the same response content as a standard GET request. not a blocker.

QUICK_START.md
```diff
@@ -148,6 +148,8 @@ returned by the "startup" call.  You can click on the value of
 "short_pitch" in the body of the "startup patch" call and then click
 change to change it to any other value.
 
+It should be noted that due to client, browser, or firewall restrictions, it may be necessary to send PATCH calls as POSTS and override them.  The middleware which overrides these types of POSTs is included in the code base.
```
shanbady: the readme should make mention of how to do the override (providing the name of the header to pass and what the value should be)

shanbady (changes requested):
ready to approve once the readme note is fixed

QUICK_START.md
knittingarch: Ah, ok.

web/impact/impact/tests/test_method_override_middleware.py
knittingarch: @shanbady Hmm, I believe this test is testing that there is no change to the data after attempting to PATCH via GET.  So the initial name is unchanged.  Does that work for you?

shanbady (approved):

PR #56, opened 2017-10-23
[AC-4857] - OPTIONS from descriptions

mc-nathan:
Done:
- Revised OPTIONS call for /v1/user, /v1/user/<id>, /v1/organization, /v1/ogranization/<id>, /v1/industry, /v1/industry/<id>, /v1/program_family, /v1/program_family/<id>, /v1/user/<id>/history, /v1/organization/<id>/history, /v1/user/<id>/organizations, /v1/organization/<id>/users, /v1/user/<id>/confidential.
- All these calls are now based on a description of the call that is then used to derive the OPTIONS response.  This description is expected to be leveraged for other calls (GET, POST, PATCH), but that has been deferred to a follow on ticket.
- This has been made reasonable abstract and should be pretty straight forward to support for future.  A good example is the UserConfidentialView where the needed code is captured in USER_CONFIDENTIAL_FIELDS and the metadata method.  Ideally, the CONFIDENTIAL_KEYS would be derived from USER_CONFIDENTIAL_FIELDS, but that's the next level that's been moved to a follow on ticket.  In this case it is simple, but it gets hairy for stuff like /v1/user as well as POST and PATCH.

To test:
- Load up the code with a fresh clean db and compare and contrast the results/arguments to the calls with what OPTIONS claims.  In Postman (collection is https://www.getpostman.com/collections/8df2bd6e977c11a3b776) you can get the OPTIONS for each call by temporarily changing the dropdown next to the URL to 'OPTIONS'.  We should probably add these calls to the Postman collection, but didn't have the time to do that for everything. **Y.M: An updated Postman collection with OPTIONS [here](https://www.getpostman.com/collections/0a95773f88df764b15ee)**


yotammanor:
**Functional issues:**

- This is the OPTIONS response on User List:
```
{
    "name": "User List",
    "description": "",
    "renders": [
        "application/json",
        "text/html"
    ],
    "parses": [
        "application/json",
        "application/x-www-form-urlencoded",
        "multipart/form-data"
    ],
    "actions": {
        "GET": {
            "type": "array",
            "item": {...}
        },
       "POST": {...}
    }
}
```
This response says to me that if I'll send a GET on `/v1/user/`, the response will be an "array". But, the GET response is an object:
```
{
    "count": 47027,
    "next": "http://localhost:8000/api/v1/user/?limit=10&offset=10",
    "previous": null,
    "results": [
    # The array is here
    ]
}
```
**Fixed**

- On User List/Detail:GET , "primary_industry_id" has to be an existing Industry ID, but this requirement is not stated in the description.
**Resolved. After some discussion we came to a consensus approach for this.  The issue comes down to where it makes sense to use words like 'allow' and 'require' for GET OPTIONS.  We ended up agreeing on "Will always be an existing Program Family ID." for GET and using 'allowed' and 'required' for PATCH and POST**

- On Organization List/Detail:GET, some fields have the readOnly attribute and/or a description that refers to the POST/PATCH restrictions, but POST/PATCH are not allowed for these views, which is a bit confusing.
**Resolved. This is partially addressed by the above comment and some other text changes.**

- Re Program Family - I was not aware this existed in v1. I'm noting that: 1. It also appears in the low level API. 2. low level's style is `/impact/ProgramFamily/` and v1 is now `/v1/program_family/`, and I don't like very much the discrepancy.
**We both prefer snake_case over CamelCase for URLs so this is really an issue for the low level API and should be addressed in a separate ticket.  See AC-5010**

- REGEX patterns come up slightly unescaped, not sure if it's an actual issue or not. Example can be found on Program Family List:GET:url_slug = "^[\\w-]+$", note the extra backslash (that escapes the one after that?)
**Not sure where the extra \ is coming from.  Worked around it by switching to "^[a-zA-Z0-9_-]+$"**

- Re User Events List, I don't like the fact that there's no ID to an event and that there's no Detail view on a particular event. But that's for a further ticket, obviously.
**I don't feel strongly enough about this one way or the other to really care.  @yotammanor If you care about this, please create a ticket.  To me this implies that all events are handled in the database as objects with their own PK and I'm not sure what the real win is for representing this effectively duplicate data in this way.  There is some appeal to having all of this explicitly logged since it will be more accurate (currently deletes leave no trace).  However, the cost is fairly high.**

- Re Organization History:GET:datetime (and other datetime fields as well) - we might want to consider adding a `format:date-time` attribute to the field, if it is indeed in jsonschema's expected format (might not be).
**I'll squeeze this in if I find time or I'll make another ticket.  I note that format:date-time and format:uri also exist which could be applied in appropriate places.  I am not considering this a blocker.**

- User Organizations:GET - missing field descriptions - startup_administrator, partner_administrator, primary_contact. (Also true for Organization Users).
**Seems out of scope to me.  We need clearer standards on when field descriptors can be left off and exactly what information they contain.  These are booleans that have pretty obvious meanings to me so I'm not sure why they need descriptors.  @yotammanor If this is important to you, please create a ticket about creating a standard around descriptions strings that can then be applied consistently to the API.**

- This is probably a follow on ticket, but I think it's a good practice for a given Node to describe it's available edges (e.g. User Detail describes available - Organizations, History and confidential (if has permissions).
**Sounds like a follow on ticket and I'm not sure how to write it.  @yotammanor feel free to create a ticket for this as well.**

web/impact/impact/v1/helpers/profile_helper.py
```diff
@@ -23,10 +27,45 @@
...
-)
+
+
+EXPERT_BOOLEAN_FIELD = {
+    "json-schema": {
+        "type": "boolean",
+    },
+    "GET": {
+        "included": "could_be_expert",
+        "description": "This field exists only when user_type is 'expert'",
```
yotammanor: The description value should be extracted to a CONSTANT and assigned to the places where it makes sense, so it's more DRY.

web/impact/impact/v1/events/base_history_event.py
```diff
@@ -0,0 +1,63 @@
...
+        if not self.earliest and hasattr(self, "calc_datetimes"):
+            self.calc_datetimes()
+
+    def description(self):
+        return None  # pragma: no cover
+
+    def serialize(self):
+        result = {}
+        for key in self.all_fields().keys():
+            value = getattr(self, key)()
```
yotammanor: I'd consider writing it as `getattr(self, key).__call__()` just so that the parenthesis won't get lost in skimming.

web/impact/impact/v1/views/industry_list_view.py
```diff
@@ -2,10 +2,14 @@
...
+from impact.v1.helpers import (
+    INDUSTRY_FIELDS,
+    IndustryHelper,
+)
 
 
 class IndustryListView(ListView):
     helper_class = IndustryHelper
-    METADATA_ACTIONS = IndustryDetailView.METADATA_ACTIONS
+
```
yotammanor, line 13: Shouldn't IndustryListView inherit from ImpactView as well?

yotammanor (changes requested):
A few functional issues above. a few code issues below. Mostly good.

web/impact/impact/v1/views/industry_list_view.py
mc-nathan: It does.  impact.v1.views.list_view.ListView inherits from BaseListView & AutoMetadataMixin and BaseListView inherits from ImpactView.

I suspect you are expecting ListView to refer to `django.views.generic.list.ListView`.  That's fair, but should be handled in a different ticket.

web/impact/impact/v1/events/organization_became_finalist_event.py
```diff
@@ -21,23 +35,23 @@ def events(cls, organization):
...
+            self.earliest = self._program.start_date
+
+    def description(self):
+        return self.DESCRIPTION_FORMAT.format(
+            self.startup_status.program_startup_status.program.name)
+
+    def cycle(self):
+        return self._program.cycle.name
+
+    def cycle_id(self):
```
shanbady, line 50: it seems like a lot of these do the exact same thing (i see ```def cycle_id``` duplicated in several places. Can we pull these into a base class?

web/impact/impact/v1/helpers/model_helper.py
```diff
@@ -131,3 +243,31 @@ def validate_url(helper, field, value):
...
+
+
+def json_list_wrapper(item):
+    return json_object(
+        {
+            "count": INTEGER_FIELD,
+            "next": URL_FIELD,
+            "previous": URL_FIELD,
+            "results": json_array(item),
+            })
```
shanbady: indent mis-match

web/impact/impact/tests/test_organization_detail_view.py
```diff
@@ -80,3 +110,21 @@ def test_organization_has_no_startup_nor_partner(self):
...
+            assert_fields(STARTUP_GET_FIELDS, get_options)
+
+    def test_partner_options(self):
+        organization = PartnerFactory().organization
+        with self.login(username=self.basic_user().username):
+            url = reverse("organization_detail", args=[organization.id])
+            response = self.client.options(url)
+            assert response.status_code == 200
+            get_options = response.data["actions"]["GET"]["properties"]
+            assert_fields(PARTNER_GET_FIELDS, get_options)
```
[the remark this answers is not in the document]
mc-nathan: Good idea.  Added and it caught a problem that I've now fixed.

web/impact/impact/v1/views/user_detail_view.py
```diff
@@ -1,46 +1,63 @@
...
+
 
 INVALID_KEYS_ERROR = "Recevied invalid key(s): {invalid_keys}."
 NO_USER_ERROR = "Unable to find user with an id of {}"
 
 User = get_user_model()
 
 
-class UserDetailView(LoggingMixin, APIView):
+class UserDetailView(ImpactView):
```
shanbady, line 22: why were all the LoggingMixins removed? this is required for rest-framework-tracking (logging requests to django admin)

shanbady (changes requested):

web/impact/impact/tests/test_organization_users_view.py
```diff
@@ -33,3 +36,12 @@ def test_get(self):
...
             self.assertFalse(partner_user["partner_administrator"])
+
+    def test_options(self):
+        startup = StartupTeamMemberFactory(startup_administrator=True).startup
+        with self.login(username=self.basic_user().username):
+            url = reverse("organization_users", args=[startup.id])
+            response = self.client.options(url)
+            assert response.status_code == 200
+            get_options = response.data["actions"]["GET"]["properties"]
+            assert_fields(ORGANIZATION_USERS_GET_FIELDS, get_options)
```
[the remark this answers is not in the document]
mc-nathan: Added.  Parallel change for user_organizations_view.

web/impact/impact/tests/utils.py
```diff
@@ -30,3 +30,23 @@ def match_errors(data, errors):
...
+
+
+def assert_fields_missing(fields, data):
+    for field in fields:
+        assert field not in data, "%s present" % field
+
+
+def assert_fields_required(fields, data):
+    for field in fields:
+        assert "required" in data[field], "%s not required" % field
```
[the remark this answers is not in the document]
mc-nathan: Not sure I understand what isn't clear.  In cases like this a suggestion would be helpful.  I went with: "Expected %s to be required".

web/impact/impact/tests/utils.py
```diff
@@ -30,3 +30,23 @@ def match_errors(data, errors):
...
+
+
+def assert_fields_not_required(fields, data):
+    for field in fields:
+        assert "required" not in data[field], "%s required" % field
```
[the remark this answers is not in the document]
mc-nathan: Went with: "Expected %s to not be required".

web/impact/impact/v1/events/base_history_event.py
mc-nathan: Directly calling `__` calls feels weird to me, but I don't care enough to really argue about it.  Changed.

web/impact/impact/v1/events/organization_became_entrant_event.py
```diff
@@ -39,68 +65,59 @@ def events(cls, organization):
...
+
+    def cycle_id(self):
+        return self._cycle.id
 
-    def _datetimes(self):
+    def programs(self):
+        raw_data = self.startup.startupprograminterest_set.filter(
+            applying=True,
+            program__cycle=self._cycle
+            ).order_by("order").values_list("program_id",
```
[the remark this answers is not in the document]
mc-nathan: It is needed (at least for now).  See the definition of StartupProgramInterest and note that:
```
In [1]: from ordered_model.models import OrderedModel
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-028439014c40> in <module>()
----> 1 from ordered_model.models import OrderedModel

ModuleNotFoundError: No module named 'ordered_model'
```

web/impact/impact/v1/events/organization_became_finalist_event.py
```diff
@@ -1,13 +1,27 @@
...
 
+    CLASS_FIELDS = {
+        "cycle": STRING_FIELD,
+        "cycle_id": INTEGER_FIELD,
+        "program": STRING_FIELD,
+        "program_id": INTEGER_FIELD,
+    }
+
     def __init__(self, startup_status):
+        super(OrganizationBecameFinalistEvent, self).__init__()
```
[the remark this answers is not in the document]
mc-nathan: Several cases of this had crept in.  Removed all of them except the ones in models.

web/impact/impact/v1/helpers/profile_helper.py
mc-nathan: Done.

web/impact/impact/v1/views/user_detail_view.py
mc-nathan: ImpactView inherits from LoggingMixin and APIView.

web/impact/impact/v1/helpers/model_helper.py
mc-nathan: Fixed.

shanbady (approved):

web/impact/impact/v1/helpers/model_helper.py
```diff
@@ -1,8 +1,120 @@
...
+
+def merge_fields(field, extension):
+    if not isinstance(field, dict) or not isinstance(extension, dict):
+        return field
+    result = {}
+    for key, value in field.items():
+        result[key] = merge_fields(value, extension.get(key))
+    for key in set(extension.keys()) - set(field.keys()):
+        result[key] = extension[key]
+    return result
```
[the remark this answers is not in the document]
mc-nathan: I'm going to push back on this.  It is fully covered.  It is used in several places and is doing what it is expected to do.  It isn't clear to me what tests you would want that would address your concern.

web/impact/impact/v1/events/organization_became_finalist_event.py
mc-nathan: Pushing back on this.  The implementation of cycle and cycle_id differ in the two organization classes and that's intentional (entrants are for cycles and finalists are associated with programs).  There is very little additional duplication.

mc-nathan:
@yotammanor It's all you.  Bring it home!

web/impact/impact/tests/utils.py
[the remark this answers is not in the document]
yotammanor: I couldn't think of a clear suggestion. But I think your choice is a good one.

web/impact/impact/v1/events/organization_became_entrant_event.py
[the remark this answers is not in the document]
yotammanor: Got it.

web/impact/impact/v1/views/industry_list_view.py
yotammanor: Did not trace down beyond BaseListView. Fair enough.

yotammanor:
Open functional issue:

User Organizations OPTIONS/GET mismatch: 

_OPTIONS:_
```
{
    "name": "User Organizations",
    "description": "",
    "renders": [
        "application/json",
        "text/html"
    ],
    "parses": [
        "application/json",
        "application/x-www-form-urlencoded",
        "multipart/form-data"
    ],
    "actions": {
        "GET": {
            "type": "object",
            "properties": {
                "organizations": {
                    "type": "array",
                    "item": {
                        "type": "integer"
                    }
                }
            }
        }
    }
}
```
_GET_:
```
{
    "organizations": [
        {
            "id": 11887,
            "startup_administrator": true,
            "primary_contact": true
        },
        {
            "id": 19165,
            "partner_administrator": true
        },
        {
            "id": 19166,
            "partner_administrator": true
        },
        {
            "id": 18515,
            "partner_administrator": false
        }
    ]
}
```

web/impact/impact/tests/test_accelerator_routes.py
```diff
@@ -41,13 +41,13 @@ class TestAcceleratorRoutes(TestCase):
...
     def setUpClass(cls):
-        super(TestAcceleratorRoutes, cls).setUpClass()
+        super().setUpClass()
         for model in MC_CONTENTTYPES:
             ContentTypeFactory(app_label='mc', model=model)
 
     @classmethod
     def tearDownClass(cls):
-        super(TestAcceleratorRoutes, cls).tearDownClass()
+        super().tearDownClass()
```
yotammanor, line 50: 👍

web/impact/impact/v1/helpers/profile_helper.py
```diff
@@ -141,15 +142,16 @@
...
     },
 }
 
 PRIMARY_INDUSTRY_ID_FIELD = {
     "json-schema": {"type": "integer"},
     "GET": {
         "included": "could_be_expert",
-        "description": "This field exists only when user_type is 'expert'",
+        "description": ("This field exists only when user_type is 'expert'. "
+                        "Will always be an existing Industry ID."),
```
yotammanor: this pains me, but not a blocker

yotammanor (changes requested):
Code looks good, most of the issues were addressed here or in a follow on ticket.

Travis build is currently failing, trying to restart another time, but until it succeeds it's an issue. Plus there's one outstanding functional issue which I consider blocking:

- User Organizations OPTIONS/GET mismatch: 

_OPTIONS:_
```
{
    "name": "User Organizations",
    "description": "",
    "renders": [
        "application/json",
        "text/html"
    ],
    "parses": [
        "application/json",
        "application/x-www-form-urlencoded",
        "multipart/form-data"
    ],
    "actions": {
        "GET": {
            "type": "object",
            "properties": {
                "organizations": {
                    "type": "array",
                    "item": {
                        "type": "integer"
                    }
                }
            }
        }
    }
}
```
_GET_:
```
{
    "organizations": [
        {
            "id": 11887,
            "startup_administrator": true,
            "primary_contact": true
        },
        {
            "id": 19165,
            "partner_administrator": true
        },
        {
            "id": 19166,
            "partner_administrator": true
        },
        {
            "id": 18515,
            "partner_administrator": false
        }
    ]
}
```

yotammanor (approved):
DONE!

PR #59, opened 2017-10-27
[AC-5019] Code cleanup of api_data.py

mc-nathan:
Resolves duplication and code complexity issues reported by CodeClimate.

### DIFF
--- a/web/impact/impact/api_data/api_data.py
+++ b/web/impact/impact/api_data/api_data.py
@@ -15,6 +15,10 @@ class APIData(object):
     YES = "Y"
     NO = "N"
     YES_NO_VALUES = [YES, NO]
+    KEY_TO_CLASS_LOOKUP = {
+        "ProgramKey": (Program, "name"),
+        "StartupKey": (Startup, "organization__url_slug"),
+    }
 
     def __init__(self, data, *args, **kwargs):
         self.errors = []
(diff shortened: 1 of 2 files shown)


pbx (approved):
Looks good. 

It does make me wish this module was tested directly, or at least showed up in the coverage report.

web/impact/impact/api_data/api_data.py
```diff
@@ -34,34 +38,25 @@ def validate_field(self, field, valid_values):
...
             return None
+        return self._lookup_object(key, obj_key)
+
+    def _lookup_object(self, key, obj_key):
+        obj_class, lookup = self.KEY_TO_CLASS_LOOKUP[obj_key]
+        if isinstance(key, int) or key.isdigit():
+            result = obj_class.objects.filter(id=key).first()
+        elif isinstance(key, str):
+            result = obj_class.objects.filter(**{lookup: key}).first()
+        if not result:
```
yotammanor: Since this is a cleanup PR I'm going to nitpick - I think that line 60-61 should be done in  `_validate_object` (or in a separate function - `_lookup` object should only be about looking up, not about validating that thing that was looked up.

Also, `result` might be undefined if `key` is neither a string nor an int.

yotammanor (changes requested):

mc-nathan:
Good catch on the refactor.  All concerns addressed.  Also caught some missing coverage that was covered on AC-4995 so moved the test to this branch.

yotammanor (approved):

PR #60, opened 2017-10-27
[AC-4960] fixing pagination protocol via setting appropriate protocol headers

shanbady:
The fix for this turned out to be far simpler - the reason for it was that nginx was using the "$scheme" variable which reports incorrectly when sitting under an ELB which in turn affected what djkango was seeing for protocol. Django latches on the X-Forwarded-Proto to determine http vs https.

Unfortunately testing this is a bit pauinful and involves setting up a sidecar letsencrypt image or a self-signed cert in nginx and fiddling with the settings.
However I have gone through this pain and tested and verified that this works.

EDIT - We have setup an elb for staging and I've pushed this branch out there for testing. 

to test:
- Login using your superuser masschallenge account here: https://staging.api.masschallenge.org
- visit a view that paginates (https://staging.api.masschallenge.org/api/impact/Industry/) and note that the "next" field shows https for a protocol instead of http (as you would see in a local environment)

### DIFF
--- a/web/impact/impact/settings.py
+++ b/web/impact/impact/settings.py
@@ -28,6 +28,8 @@ class Base(Configuration):
 
     USE_TZ = True
 
+    USE_X_FORWARDED_HOST = True
+
     ALLOWED_HOSTS = [
         '*'
     ]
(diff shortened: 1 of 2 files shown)


mc-nathan:
Can this get deployed to a staging system so we can functionally test it?  The description for how to test it is incomplete and doesn't seem worth the time to unpack.

mc-nathan (approved):
Looks reasonable to me and the staging systems seems to be doing the right thing.

pbx (approved):
Thanks for the fix.

PR #61, opened 2017-10-30
[AC-5023] Unpinned oauthlib due to fix in open source project

mc-nathan:
Thread from oauthlib open source project:
https://github.com/idan/oauthlib/issues/490#issuecomment-338269111

To test:
- Verify that tests pass on Travis compared to https://travis-ci.org/masschallenge/impact-api/builds/289533334

### DIFF
--- a/web/impact/requirements/base.txt
+++ b/web/impact/requirements/base.txt
@@ -7,7 +7,7 @@ coreapi
 hiredis==0.2.0
 markdown # Markdown support for the browsable API.
 mysqlclient
-oauthlib==2.0.4
+oauthlib
 
 # Misc
 gunicorn


web/impact/requirements/base.txt
```diff
@@ -7,7 +7,7 @@ coreapi
 hiredis==0.2.0
 markdown # Markdown support for the browsable API.
 mysqlclient
-oauthlib==2.0.4
```
pbx, line 10: I'd really like to see us get into the practice of pinning dependencies to explicit version numbers. Many times we've had to scramble to deal with problems when an automatic upgrade of a dependency has broken our application. 

I'm planning to make  a ticket on that subject, so this doesn't have to be a blocker for me on this PR, but I wanted to raise it -- especially given that an unpinned dependency was the cause of the original bug here ([AC-4989](https://masschallenge.atlassian.net/browse/AC-4989)).

yotammanor (approved):
travis works, also builds locally with no problem.

pbx (approved):
Looks good, glad the fix came so quickly.

Nathan and I had a good exchange about pinning version dependencies. As a result I will write an even better ticket about how we should pin our dependencies.

PR #66, opened 2017-11-07
[AC-5045] fix link to bootstrap's CDN.

yotammanor:
**Test:**

- on development, login to browsable api, see that the css is not nice.
- on branch, very nice.


mc-nathan (approved):
Works. Change is simple.  Would be interesting to know why this change and if there's anything we could reasonably have done to be aware of this ahead of time.

pbx (approved):

PR #67, opened 2017-11-14
AC-5067: Fix 500 for GET of /v1/user/35908 (or any user without a profile)

mc-nathan:
Work done:
- Code was fixed to work with users that have no profile objects.

To test:
- `make load-remote-db DB_FILE_NAME=initial_schema.sql.gz`
- Grant the appropriate permissions to a super user account.  E.g., `make grant-permissions PERMISSION_USER=13514-user@example.com PERMISSION_CLASSES=v1_clients`
- Login as that user.  E.g., 13514-user@example.com
- Visit http://localhost:8000/api/v1/user/35908/
- On development it should generate a 500 and on the branch it should be fine.
- Try other calls with this user or with any of the other "broken" users (39658, 38978, 41341, and 44399)
- Note: Some PATCH calls will now return 403s with appropriate error messages if you try to PATCH a field that doesn't exist.  You should not be able to generate any 500s.


web/impact/impact/tests/test_user_detail_view.py
```diff
@@ -525,6 +542,33 @@ def test_patch_invalid_primary_industry_id(self):
...
+        user.save()
+        with self.login(email=self.basic_user().email):
+            url = reverse(UserDetailView.view_name, args=[user.id])
+            new_last_name = user.last_name + ", Jr."
+            response = self.client.patch(url, {"last_name": new_last_name})
+            assert response.status_code == 204
+            user.refresh_from_db()
+            assert user.last_name == new_last_name
+
+    def test_patch_profile_file_with_no_profile(self):
```
yotammanor: name seems to be erroneous. `profile_file` bit.

yotammanor (changes requested):
one bad name test, otherwise all is good and as advertised.

pbx (approved):
Confirmed error on development and absence thereof on the branch.

yotammanor (approved):

PR #68, opened 2017-11-15
[AC-5010] Switch low-level API to use snake_case rather than CamelCase

knittingarch:
**To Test**

**Set Up The Environment**
- Start the api server: `make clean; make build; make dev`
- Load the DB: `make load-remote-db DB_FILE_NAME=initial_schema.sql.gz`
- Grant permissions: `make grant-permissions PERMISSION_USER=demoadmin@masschallenge.org PERMISSION_CLASSES=v0_clients,v1_clients`
- In a separate terminal window, run `make runserver`. This seems to help with load speed.  A follow-on ticket will be created to address the load slowness.

**Check Things Still Work**
- Open up a browser window and navigate to the browsable API.
-  You should see the model names at the end of the URLs in snake_case.
- Click on (at least) the following URLs:
  - [http://127.0.0.1:8000/api/impact/startup_additional_industries/](http://127.0.0.1:8000/api/impact/startup_additional_industries/)
  - [http://127.0.0.1:8000/api/impact/user_label/](http://127.0.0.1:8000/api/impact/user_label/)
  - [http://127.0.0.1:8000/api/impact/expert_profile_mentoring_specialties/](http://127.0.0.1:8000/api/impact/expert_profile_mentoring_specialties/)
  - [http://127.0.0.1:8000/api/impact/scenario/](http://127.0.0.1:8000/api/impact/scenario/)
  - [http://127.0.0.1:8000/api/impact/pay_pal_refund/](http://127.0.0.1:8000/api/impact/pay_pal_refund/)
- All calls should work!


web/impact/impact/models/utils.py
```diff
@@ -2,8 +2,28 @@
 # Copyright (c) 2017 MassChallenge, Inc.
 
 
+import re
+
```
yotammanor: some redundant spaces around here.

web/impact/impact/tests/test_api_routes.py
```diff
@@ -111,6 +111,40 @@ def test_api_object_get(self):
...
+            codename='view_startup_additional_industries'
+        )
+        url_name = "object-list"
+        view_kwargs = {'app': 'impact',
+                       "model": "Startup_additional_industries"}
+        self.response_401(self.get(url_name, **view_kwargs))
+
+        basic_user = self.make_user('basic_user@test.com')
+        with self.login(basic_user):
+            self.response_403(self.get(url_name, **view_kwargs))
```
yotammanor, line 132: The 401/403 asserts seem unrelated to the test you're attempting.

web/impact/impact/tests/test_api_routes.py
```diff
@@ -111,6 +111,40 @@ def test_api_object_get(self):
...
+
+        perm_user = self.make_user(
+            'perm_user@test.com',
+            perms=["mc.view_startup_additional_industries"])
+        perm_user.user_permissions.add(view_perm)
+        perm_user.save()
+        with self.login(perm_user):
+            response = self.get(url_name, **view_kwargs)
+            self.response_200(response)
```
yotammanor: the name of the test suggests there's something that should be displayed, but I see no statement that checks that. Therefore name of the test should change or an assert should be added/replaced.

web/impact/impact/tests/test_snake_and_camel_casing.py
```diff
@@ -0,0 +1,16 @@
...
+from impact.models.utils import (
+    camel_case_to_snake,
+    snake_to_camel_case,
+)
+
+camel_case = 'ProgramStartupStatus'
+snake_case = 'program_startup_status'
+
+
+class TestSnakeAndCamelCasing(APITestCase):
```
yotammanor: Our standards require that the name of this class (and module) should be `TestUtils` and `test_utils.py`, respectively (or the original functions should be moved).

web/impact/impact/tests/test_snake_and_camel_casing.py
```diff
@@ -0,0 +1,16 @@
...
+    def test_camel_case_to_snake_case(self):
+        assert camel_case_to_snake(camel_case) == snake_case
+
+    def test_snake_case_to_camel_case(self):
+        assert snake_to_camel_case(snake_case) == camel_case
```
yotammanor: Was very glad to see those unit tests here!
I'd love to see also a test on an empty string, and lower-case word, and see that they're handled as expected.

web/impact/impact/urls.py
```diff
@@ -1,11 +1,16 @@
 # MIT License
 # Copyright (c) 2017 MassChallenge, Inc.
 
+
```
yotammanor: redundant linebreak.

web/impact/impact/urls.py
```diff
@@ -34,9 +39,6 @@
     UserListView,
     UserOrganizationsView,
 )
-from rest_framework import routers
-from drf_auto_endpoint.router import router as schema_router
-from django.apps import apps
```
yotammanor, line 43: thanks for sorting!

web/impact/impact/views/general_view_set.py
```diff
@@ -18,9 +34,14 @@ class GeneralViewSet(LoggingMixin, viewsets.ModelViewSet):
 
     @property
     def model(self):
-        return apps.get_model(
-            app_label=self.kwargs['app'],
-            model_name=str(self.kwargs['model']))
+        if self.kwargs['model'] in model_attribute_calls:
+            return apps.get_model(
+                app_label=self.kwargs['app'],
+                model_name=self.kwargs['model'])
```
yotammanor: This smells fishy to me, and might be related to functional issues.

web/impact/impact/views/general_view_set.py
```diff
@@ -6,10 +6,26 @@
...
+    'StartupLabel_startups',
+    'RefundCode_programs',
+    'UserLabel_users',
+    'Observer_newsletter_cc_roles',
+    'ExpertProfile_functional_expertise',
+    'ExpertProfile_additional_industries',
+    'ExpertProfile_mentoring_specialties',
+    'Newsletter_recipient_roles',
+    'Section_interest_categories',
+]
```
yotammanor: I don't like the fact that this is a manually generated list. There are ways to automate this, but better yet - those attribute calls should be parsed differently, as stated above.

web/impact/impact/models/utils.py
```diff
@@ -2,8 +2,28 @@
...
 def is_managed(db_table):
     return False
+
+
+def camel_case_to_snake(value):
+    original_model_string = re.findall('[A-Z][^A-Z]*', value)
+    holder = ""
+    for word in original_model_string:
+        holder += word.lower() + "_"
+        new_model_string = holder[:-1]
```
yotammanor: This line should probably be outdented.

web/impact/impact/models/utils.py
```diff
@@ -2,8 +2,28 @@
...
+
 LABEL_LENGTH = 255
 
 
 def is_managed(db_table):
     return False
+
+
+def camel_case_to_snake(value):
+    original_model_string = re.findall('[A-Z][^A-Z]*', value)
```
yotammanor: This regex fails to catch a string that starts with lowercase characters. Not sure if we care, but it's worth noting that official camelCase is considered to start with lower-case characters for the first words (https://en.wikipedia.org/wiki/Camel_case).

yotammanor: And you got to love the irony in that wikipedia entry.

yotammanor (changes requested):
Functional issues:

- When I try to go to /startup_additional_industries/, I get the a 500 error `App 'impact' doesn't have a 'startupadditionalindustries' model.`
- trying to get into the old version,
http://localhost:8000/api/impact/Startup_additional_industries/ - this actually worked (after a few what I think were timeout errors)
- As I've mentioned before the attribute endpoint in the edge of the model is actually a parsing bug. The correct form should be `/startup/additional_industries/`. This is how well-behaved APIs look like, the related objects are referred to separately. I'd very much like this to be addressed in this ticket, and I still think it would make the changes here cleaner and easier than trying to account for this list of edge cases.

Also, a few code related comments below.

web/impact/impact/tests/test_snake_and_camel_casing.py
knittingarch: I don't understand this point.  Can you expound, please?

knittingarch: Done!

web/impact/impact/models/utils.py
knittingarch: Removed.

web/impact/impact/models/utils.py
knittingarch: Name changed.

web/impact/impact/models/utils.py
knittingarch: Done.

web/impact/impact/urls.py
knittingarch: Removed.

web/impact/impact/views/general_view_set.py
knittingarch: Skipping per earlier conversation. Follow-on ticket to be written by @yotammanor.

web/impact/impact/tests/test_api_routes.py
knittingarch: Name changed.

web/impact/impact/tests/test_api_routes.py
knittingarch: This was an adaptation of another test and was pretty complex to figure out.  I'm going to leave this as is unless you feel very strongly about it.

web/impact/impact/tests/test_snake_and_camel_casing.py
knittingarch: Can you clarify what you are looking for with "lower-case word"?  The snake case already checks lower case...

web/impact/impact/urls.py
```diff
@@ -38,23 +42,20 @@
...
-        schema_router.register(model, url=model.__name__)
-
+    if model._meta.app_label == 'impact' and not model._meta.auto_created:
+        schema_router.register(model, url=model_name_to_snake(model.__name__))
 
 for model in apps.get_models('accelerator'):
-    if model._meta.app_label == 'accelerator' and hasattr(model, "Meta"):
+    if (model._meta.app_label == 'accelerator' and
+            hasattr(model, "Meta") and
+            not model._meta.auto_created):
```
yotammanor: this piece of code is a bit nasty, and should probably be refactored.

web/impact/impact/urls.py
```diff
@@ -146,10 +147,25 @@
...
+        r'(?P<related_model>[a-z_]+)/'
+        r'(?P<pk>[0-9]+)/$',
+        GeneralViewSet.as_view({
+            'get': 'retrieve',
+            'put': 'update',
+            'patch': 'partial_update',
+            'delete': 'destroy'
+        }),
+        name='related-object-detail'),
+    url(r'^api/(?P<app>\w+)/(?P<model>[a-z_]+)/$',
```
yotammanor, line 167: maybe the `(?P<app>\w+)` part should also be lower-case only? not sure.

web/impact/impact/utils.py
```diff
@@ -45,6 +45,22 @@ def get_profile(user):
...
+        return model + '_' + related_model
+    return model
+
+
+def model_case_model_name(view):
+    model = snake_to_model_name(view.kwargs.get('model', ''))
+    related_model = view.kwargs.get('related_model')
+    if related_model:
+        return model + '_' + related_model
+    return model
```
yotammanor: Those two functions are very similar and can probably be merged (note the only difference in `model=...`)

yotammanor (changes requested):
A little bit of cleanup, but mostly looks good.
We should also add a test that uses `related-detail-view` to make sure that it works (seems like it currently isn't :) )

web/impact/impact/utils.py
knittingarch: Refactored!

web/impact/impact/urls.py
knittingarch: Chose to follow this style guide: https://www.naftaliharris.com/blog/python-multiline-if-statements/

web/impact/impact/urls.py
knittingarch: This doesn't seem to matter.  Leaving as is.

yotammanor (approved):
MAW

web/impact/impact/models/utils.py
```diff
@@ -1,9 +1,27 @@
...
 
 LABEL_LENGTH = 255
 
 
 def is_managed(db_table):
     return False
+
+
+def model_name_to_snake(value):
+    original_model_string = re.findall('[A-Z][^A-Z]*', value)
```
shanbady: instead of using your own pattern you can use a built-in django util:
```
from django.utils.text import camel_case_to_spaces
camel_case_to_spaces('ThisisATest') # outputs u'thisis a test'

```
it automatically lower-cases and splits by camelcasing

web/impact/impact/utils.py
```diff
@@ -45,6 +47,13 @@ def get_profile(user):
         return None
 
 
+def model_name_case(view, model):
+    related_model = view.kwargs.get('related_model')
```
shanbady: it seems that you are only using the 'view' argument to get the 'related model' keyword argument. having the 'view' as an argument forces it to do some odd invocations like passing in 'self' for viewsets that use it. Instead make this argument the 'related_model' and pass that in directly from wherever it gets used.

shanbady (dismissed):

web/impact/impact/models/utils.py
knittingarch: Oooh!  Didn't realize this existed!

web/impact/impact/utils.py
knittingarch: @yotammanor @jonkiparsky @pbx I am unable to pass in related_model with the current configuration.  It's taking the first parameter put in as the instance of GeneralViewSet.  So, I'm not sure what Shanks had in mind here to remove the view instance from this helper method...  Any ideas?

jonkiparsky: @knittingarch it looks like @shanbady is suggesting something like the following:

```
def model_name_case(related_model, model):
    if related_model:
        return model + '_' + related_model
    return model
```

And then extract the 'related_model' field at the place where you call it. Does this not work?

ie
```
related_model = view.kwargs.get('related_model')
foo = model_name_case(related_model, model)
```

knittingarch: @jonkiparsky Yes, that is also how I interpreted his comment.  Unfortunately, this doesn't work as "related_model" ends up being essentially a renamed "view" parameter.  I don't see how to get around that weirdness.

web/impact/impact/models/utils.py
```diff
@@ -16,3 +18,17 @@ def model_has_field(model, field_name):
...
+    original_model_string = camel_case_to_spaces(value)
+    new_model_string = original_model_string.replace(" ", "_")
+    return new_model_string
+
+
+def snake_to_model_name(value):
+    old_value = value.split('_')
+    new_value = ""
+    for word in old_value:
+        new_value += (word[0].upper() + word[1:])
```
jonkiparsky: In more idiomatic python, this would be:
```
def snake_to_model_name(value):
    pieces = value.split("_")
    return "".join(map(str.title, pieces))
```
or just 
```
def snake_to_model_name(value):
    return "".join(map(str.title, value.split("_")))
```

jonkiparsky: Do we want to worry about the possibility of spaces existing in `value`? 
To me, that would be an invalid input and should throw an exception, but that might be over-thinking it.

web/impact/impact/tests/test_api_routes.py
```diff
@@ -111,6 +111,41 @@ def test_api_object_get(self):
...
+        StartupFactory(is_visible=1, organization__url_slug="test1")
+        StartupFactory(is_visible=1, organization__url_slug="test2")
+        StartupFactory(is_visible=1, organization__url_slug="test3")
+        Permission.objects.get_or_create(
+            content_type=ContentType.objects.get(
+                app_label='mc',
+                model='startup'),
+            codename='view_startup_additional_industries'
+        )
+        view_perm, _ = Permission.objects.get_or_create(
```
jonkiparsky: this looks like a duplication of the previous statement, delete

web/impact/impact/tests/test_api_routes.py
```diff
@@ -111,6 +111,41 @@ def test_api_object_get(self):
             response_dict = json.loads(response.content)
             self.assertIn("is_visible", response_dict.keys())
 
+    def test_api_related_object_list_get(self):
```
jonkiparsky, line 114: Not a blocker, but this name is not informative. I don't know what this test is actually intended to test. 
Mentioning it because we do a lot of this and it'd be nice to get out of this habit.

web/impact/impact/tests/test_api_routes.py
```diff
@@ -111,6 +111,41 @@ def test_api_object_get(self):
...
+
+        basic_user = self.make_user('basic_user@test.com')
+        with self.login(basic_user):
+            self.response_403(self.get(url_name, **view_kwargs))
+
+        perm_user = self.make_user(
+            'perm_user@test.com',
+            perms=["mc.view_startup_additional_industries"])
+        perm_user.user_permissions.add(view_perm)
+        perm_user.save()
```
jonkiparsky: .save not needed

web/impact/impact/tests/test_utils.py
```diff
@@ -1,22 +1,28 @@
 # MIT License
```
jonkiparsky, line 1: It looks like the permission mode on this file was changed. Is this intentional?

web/impact/impact/tests/test_utils.py
```diff
@@ -1,22 +1,28 @@
...
+    model_name_to_snake,
+    snake_to_model_name,
 )
+from impact.tests.api_test_case import APITestCase
+
+
+empty_string = ''
+model_name = 'ProgramStartupStatus'
+lowercase_model_name = 'programstartupstatus'
+snake_case = 'program_startup_status'
```
jonkiparsky: this should be `SNAKE_CASE_MODEL_NAME`

web/impact/impact/tests/test_utils.py
```diff
@@ -1,22 +1,28 @@
...
-    MemberProfile,
+from impact.models.utils import (
+    model_name_to_snake,
+    snake_to_model_name,
 )
+from impact.tests.api_test_case import APITestCase
+
+
+empty_string = ''
+model_name = 'ProgramStartupStatus'
```
jonkiparsky: I would prefer if this were named `CAMEL_CASE_MODEL_NAME` (since it's a constant, and we're differentiating between different cases)

web/impact/impact/tests/test_utils.py
```diff
@@ -1,22 +1,28 @@
...
+lowercase_model_name = 'programstartupstatus'
```
jonkiparsky: Upcase the name, since it's a constant

web/impact/impact/tests/test_utils.py
```diff
@@ -1,22 +1,28 @@
...
 
+    def test_model_name_to_snake_case(self):
+        assert model_name_to_snake(model_name) == snake_case
 
-class TestUtils(TestCase):
-    def test_get_profile_for_expert(self):
-        user = UserContext("EXPERT").user
-        assert isinstance(get_profile(user), ExpertProfile)
+    def test_lowercase_model_name_not_snaked(self):
+        assert model_name_to_snake(lowercase_model_name) != snake_case
```
jonkiparsky: I would prefer `assert model_name_to_snake(lowercase_model_name) == lowercase_model_name`

web/impact/impact/tests/test_utils.py
```diff
@@ -1,22 +1,28 @@
...
+
+class TestUtils(APITestCase):
+    def test_empty_string_snake_case_returns_nothing(self):
+        assert model_name_to_snake(empty_string) == ''
 
+    def test_model_name_to_snake_case(self):
+        assert model_name_to_snake(model_name) == snake_case
 
-class TestUtils(TestCase):
-    def test_get_profile_for_expert(self):
```
jonkiparsky: Why is this test deleted?

web/impact/impact/tests/test_utils.py
```diff
@@ -1,22 +1,28 @@
...
-        user = UserContext("EXPERT").user
-        assert isinstance(get_profile(user), ExpertProfile)
+    def test_lowercase_model_name_not_snaked(self):
+        assert model_name_to_snake(lowercase_model_name) != snake_case
 
-    def test_get_profile_for_member(self):
```
jonkiparsky: And this one?

web/impact/impact/urls.py
```diff
@@ -38,23 +42,22 @@
...
+    if model._meta.app_label == 'impact' and not model._meta.auto_created:
+        schema_router.register(model, url=model_name_to_snake(model.__name__))
 
 for model in apps.get_models('accelerator'):
-    if model._meta.app_label == 'accelerator' and hasattr(model, "Meta"):
+    if (
+        model._meta.app_label == 'accelerator'
+        and hasattr(model, "Meta")
+        and not model._meta.auto_created
+    ):
```
jonkiparsky, line 60: I don't love the hanging `):`
Can we remove the line breaks at 56 and 59?

jonkiparsky (changes requested):
Some style suggestions.

web/impact/impact/models/utils.py
knittingarch: Updated.  Value will never have spaces as that isn't how the models are formatted.  They're always pseudo camel case or underscores.

web/impact/impact/tests/test_api_routes.py
knittingarch: So I thought this was actually quite informative.  It gets the list of related objects as opposed to the tests above it which are just for lists of objects...

web/impact/impact/tests/test_api_routes.py
knittingarch: Deleted.

web/impact/impact/tests/test_api_routes.py
knittingarch: Removed.

web/impact/impact/tests/test_utils.py
knittingarch: This file was created from scratch, so nothing was changed.

web/impact/impact/tests/test_utils.py
knittingarch: Changed.

web/impact/impact/tests/test_utils.py
knittingarch: Changed.

web/impact/impact/tests/test_utils.py
knittingarch: Changed.

web/impact/impact/tests/test_utils.py
knittingarch: @yotammanor Did you make these changes?  I know you merged things yesterday, and I created this file from scratch.  I have no idea what this code is.

web/impact/impact/tests/test_utils.py
knittingarch: @yotammanor And this one as well?

web/impact/impact/urls.py
knittingarch: See the previous discussion.  I will let you and @yotammanor debate this.

web/impact/impact/tests/test_utils.py
knittingarch: Changed.

web/impact/impact/tests/test_utils.py
jonkiparsky: I think I see what happened. Not a big deal, I was just curious.

web/impact/impact/urls.py
jonkiparsky: Ideally, complex conditionals would be named for the condition we actually care about and turned into functions. Part of the problem for me in this code review is I have no idea what the purpose of this conditional is - what does it mean when a model has a `.meta.app_label` of `'accelerator'` and has an attribute called `Meta` and meta says it's auto-created? 

Since we just want to get this merged, I'm not going to push this point, but our house style (many examples) for a complex conditional looks like this:

```
    if (model._meta.app_label == 'accelerator' and
        hasattr(model, "Meta") and
       not model._meta.auto_created):
```

jonkiparsky: Now that I look at that list of conditions, though - does `not hasattr(model, 'Meta')` imply that the lookup `model._meta` will fail? I don't know off hand, but it seems safer to do the hasattr check first.

jonkiparsky (approved):

PR #69, opened 2017-11-15
AC-5058 "a": Add ProgramCycle and Program resources

mc-nathan:
Also tweaked ProgramFamily as requested in the ticket.

To test:
- http://localhost:8000/api/v1/program_cycle and http://localhost:8000/api/v1/program should work as should the detail views for a given id
- http://localhost:8000/api/v1/program_family should include an "is_open" field for each ProgramFamily.


yotammanor (changes requested):
Functionally looks good (did not create a postman collection, browsable-api based tests were done).

Code looks good, one small thing to be fixed.

yotammanor (approved):
just added it.

MAW.

pbx (approved):
Looks good

PR #70, opened 2017-11-16
[AC-5059] - Build a public directory that searches Organization

yotammanor:
**Changes Introduced in [AC-5059](https://masschallenge.atlassian.net/browse/AC-5059)**

- Add an ability to filter Organization List View by the name field (contains insensitive).
- Add some of the needed apparatus to make filtering a generic capability of List Views.
- Test this.

**Test:** 
- Start Impact-api
- On OrganizationListView, add query parameter `?name=100` (or some other string, with test data all names are `r"Organization \d+"`), and send request. See the results change with the parameter.
- Value of "Count" should change as well.
- Value of "Next" should still contain the filtered parameter.

**Note:**
- This feature is tested implicitly by https://github.com/masschallenge/accelerator-directory/pull/1


web/impact/impact/v1/views/base_list_view.py
```diff
@@ -63,6 +74,41 @@ def _filter_by_date(self, qs, after, before):
...
+def _update_query_param(url, query_param_key, query_param_value=None):
+    parsed = urlparse(url)
+    qs = parsed.query
+    parsed_qs = parse_qs(qs) if qs else {}
+    parsed_qs.update({query_param_key: query_param_value})
+    return _build_url_with_query(parsed, parsed_qs)
+
+
+def _build_url_with_query(parsed_url, parsed_qs):
+    return urlunparse((parsed_url[0],
```
mc-nathan, line 122: Not a blocker, but I'd rather see these handled using the ParseResult object that actually comes back from urlparse.  It should allow you to just do:
```
parsed_url.query = urlencode(parsed_qs, doseq=True)
urlunparse(parsed_url)
```
rather than using numeric indexes.

web/impact/impact/v1/views/organization_list_view.py
```diff
@@ -13,6 +13,13 @@ class OrganizationListView(BaseListView):
     view_name = "organization"
     helper_class = OrganizationHelper
 
+    def filter(self, queryset):
```
mc-nathan: Not a blocker, but it would be nice to push this down into BaseListView so any object that has a .name automatically gets the ability to search by name substring.

mc-nathan (approved):
Code looks good (a couple non-blocking suggestions).  Functionality is as expected.

web/impact/impact/v1/views/base_list_view.py
yotammanor: I tried that and it didn't allow me to do `set_attribute` on it, which was indeed a sad moment.

web/impact/impact/v1/views/organization_list_view.py
yotammanor: @mc-nathan done.

web/impact/impact/v1/views/base_list_view.py
mc-nathan: Interesting.  I was almost certain that I got it to work for me in a python shell, but now I can't replicate it.

pbx (approved):
Looks good, works as described.

PR #73, opened 2017-11-17
AC-5058b: Separate "became entrant" events and deeper industry info in GET

mc-nathan:
Work done:
- "became entrant" events no longer list the set of Programs that a Startup is interested in, but instead there is a separate event for each Program in a Cycle that the Startup is interested in.
- GET calls for user list, user detail, organization list and organization detail provide full Industry details for the "primary_industry" and "additional_industries".  Note: This makes POST and PATCH a bit asymmetrical since they still only take the "primary_industry_id".

To test:
- `make load-remote-db DB_FILE_NAME=initial_schema.sql.gz`
- `make grant-permissions PERMISSION_USER=demoadmin@masschallenge.org PERMISSION_CLASSES=v1_clients`
- Create client application in the new system.
- Best to use Postman.  Remember to update the client_id with the new one from the above.
- Look at the output of the GET options for OrganizationHistory (e.g., http://localhost:8000/api/v1/organization/10542/history/) and note that on the branch there are multiple events for the appropriate cycles, but on development these events are combined through the "programs" key.  In the example, look for "cycle_id": 6.  This startup has two StartupProgramInterest objects for the same Program which is a data problem not an issue addressed by this ticket.
- Test GET, POST, PATCH as appropriate for /v1/user, /v1/user/<id>, /v1/organization, and /v1/organization/<id> with special attention to the primary industry and additional industries related fields.
- Look at the OPTIONS for all the above.  They should accurate reflect the changes, values returned, and the values expected for these calls.


shanbady:
@mc-nathan I'm not seeing ```make reset-remote-db``` as a makefile command

web/impact/impact/v1/helpers/model_helper.py
```diff
@@ -24,6 +24,10 @@ def merge_fields(field, extension):
     return result
 
 
+def json_schema(type):
+    return {"json-schema": {"type": type}}
```
shanbady: rename 'type' so the name doesnt clash with a built-in python method

web/impact/impact/v1/helpers/profile_helper.py
```diff
@@ -386,10 +395,20 @@ def find_user_type(self):
             return self.subject.user_type
 
     @property
-    def additional_industry_ids(self):
+    def additional_industries(self):
         if hasattr(self.subject, "additional_industries"):
-            return self.subject.additional_industries.values_list(
-                "id", flat=True)
+            industries = self.subject.additional_industries
+            result = []
```
shanbady: seems like a lot of this is duplicated from https://github.com/masschallenge/impact-api/pull/73/files#diff-69930414b4eeea3bab51315196d2df4eR103

it could be consolidated.

web/impact/impact/v1/helpers/user_helper.py
```diff
@@ -61,9 +62,10 @@
     "company": EXPERT_STRING_FIELD,
     "title": EXPERT_STRING_FIELD,
     "bio": NON_MEMBER_STRING_FIELD,
+    "primary_industry": EXPERT_INDUSTRY_FIELD,
     "primary_industry_id": PRIMARY_INDUSTRY_ID_FIELD,
```
shanbady, line 66: does this field (primary_industry_id) need to exist anymore?

shanbady (dismissed):

web/impact/impact/v1/helpers/profile_helper.py
mc-nathan: Yeah, CodeClimate caught that as well.  Duplication removed.

web/impact/impact/v1/helpers/user_helper.py
mc-nathan: Yes, for PATCH and POST.

web/impact/impact/v1/helpers/profile_helper.py
mc-nathan: Also fixed the Code Complexity issue reported by CodeClimate.

web/impact/impact/v1/helpers/model_helper.py
mc-nathan: Fixed.

web/impact/impact/v1/helpers/profile_helper.py
```diff
@@ -182,10 +187,10 @@ def post_description(classname):
                      RESOURCE_MUST_EXIST.format(classname=classname)])
 
 
-def object_id_field(klass):
+def object_id_field(klass, type="integer"):
```
yotammanor: `type` is a builtin python function, but either way I don't understand why this parameter exist, it is not used at all. If it meed exist for future work, the name of the kwarg should change slightly.

yotammanor (changes requested):
Functional issues:

- In V1 User List View GET, results have a "primary_industry_id", but it does not appear in V1 User List View OPTIONS:GET.
- Likewise in V1 User Detail View GET.
- V1 User Detail PATCH/POST - trying to use an invalid primary_industry_id (a string, or non existing id) does not raise an indicative enough error. Probably non blocker. 

Otherwise too small comments.

web/impact/impact/tests/test_organization_history_view.py
```diff
@@ -176,20 +176,44 @@ def test_startup_became_entrant(self):
...
         cycle = ProgramCycleFactory()
+        application = ApplicationFactory(
+            application_status=SUBMITTED_APP_STATUS,
+            application_type=cycle.default_application_type,
+            cycle=cycle)
+        startup = application.startup
+        StartupProgramInterestFactory(startup=startup,
+                                      program__cycle=cycle,
+                                      applying=True)
+        StartupProgramInterestFactory(startup=startup,
```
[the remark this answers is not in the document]
mc-nathan: Done.  Just so you know I don't love create_batch.  It's bitten me in weird ways before, but it seems to work in this case.

web/impact/impact/v1/helpers/profile_helper.py
mc-nathan: It was a left over from an approach I abandoned.  Removed.

mc-nathan:
@yotammanor @shanbady  I have addressed all review comments.

yotammanor (approved):
MAW

PR #74, opened 2017-11-27
AC-5123: Make cycle_id and program_id just integers for "became finalist"

mc-nathan:
To test:
/api/v1/organization/66/history/ should have an integer for "cycle_id" and "program_id" in the "became finalist" event rather than a single element list.
On the clean database you should be able to go their as 13514-user@example.com.


shanbady (approved):
looks good.

yotammanor (approved):

PR #81, opened 2017-12-19
[AC-5087] - Improve previous-next mechanism in BaseListView in impact-api

yotammanor:
**Changes Introduced in [AC-5087](https://masschallenge.atlassian.net/browse/AC-5087):**

- validate offset and limit values to be correct (offset is a non-negative integer, limit is a positive integer smaller than MAX_LIMIT=200).
- fix generation of previous/next urls around edge cases (close to offset==0 or offset==count)

**Test:**

- glossary:
  - limit - the number of results returned in the response ("results per page").
  - offset - the index of the element that begins the response window ([offset:offset+limit])
  - count - the total number of results in a request (NOT a parameter)

- Setup:
  - `make dev; 
make load-remote-db DB_FILE_NAME=initial_schema.sql.gz; 
make grant-permissions PERMISSION_USER=test@example.org PERMISSION_CLASSES=v1_clients`
  - [Create an application](http://localhost:8000/oauth/applications/register/) for the user with the permissions.
  - Load [Postman collection](https://www.getpostman.com/collections/0a95773f88df764b15ee), initialize the token
  - in V1 folder, open v1 user list GET
  - make the default call, see that it works, and is the same as development
- Test:
  - set the query parameters ?limit=20 . Same as development.
  - set the query parameters ?limit=10&offset=5 . On development, "previous" would have a negative offset. On branch it does not (will be have an implicit offset=0).
  - set the query parameters ?limit=10&offset=15. Should be the same as development.
  - set the query parameters ?limit=10&offset=47915 (count=47920). "next" will be null on branch.
  - set the query parameters ?limit=10&offset=50000 (count=47920). "previous" will be 47910(= count-limit) on branch.
  - Go to "v1 program family list GET". set the query parameters ?limit=20 (count=11). "next" and "previous" should both be null on branch.
  - set the query parameters ?limit=300. See error on branch.
  - set the query parameters ?limit=5.5. See error on branch.
  - set the query parameters ?limit=0. See error on branch.
  - set the query parameters ?limit=-5. See error on branch.
  - set the query parameters ?limit=foo. See error on branch.
  - set the query parameters ?offset=5.5. See error on branch.
  - set the query parameters ?offset=-5. See error on branch.
  - set the query parameters ?offset=foo. See error on branch.
  - set the query parameters ?offset=0. The same as not adding the parameter, successful.

### DIFF
--- a/web/impact/impact/models/utils.py
+++ b/web/impact/impact/models/utils.py
@@ -31,7 +31,7 @@ def snake_to_model_name(value):
 
 def is_int(s):
     try:
-        int(s)
+        int(str(s))
     except ValueError:
         return False
-    return True
+    return
--- a/web/impact/impact/models/utils.py
+++ b/web/impact/impact/models/utils.py
@@ -34,4 +34,4 @@ def is_int(s):
         int(str(s))
     except ValueError:
         return False
-    return
+    return True
--- a/web/impact/impact/models/utils.py
+++ b/web/impact/impact/models/utils.py
@@ -4,7 +4,6 @@
 from django.core.exceptions import FieldDoesNotExist
 from django.utils.text import camel_case_to_spaces
 
-
 LABEL_LENGTH = 255
 
 
@@ -28,3 +27,11 @@ def model_name_to_snake(value):
 
 def snake_to_model_name(value):
     return "".join(map(str.title, value.split("_")))
+
+
+def is_int(s):
+    try:
+        int(s)
+    except ValueError:
+        return False
+    return True
--- a/web/impact/impact/tests/test_user_list_view.py
+++ b/web/impact/impact/tests/test_user_list_view.py
@@ -202,6 +202,44 @@ class TestUserListView(APITestCase):
 
             assert response.data["next"] is None
 
+    def test_get_pagination_attrs_for_offset_equals_count(self):
+        limit = 4
+        for _ in range(limit * 5):
+            UserContext()
+        with self.login(email=self.basic_user().email):
+            current_offset = User.objects.count()
+            limit_arg = "limit={}".format(limit)
+            offset_arg = "offset={}".format(current_offset)
+            url = self.url + "?" + limit_arg + "&" + offset_arg
+
+            response = self.client.get(url)
+            results = response.data["results"]
+            assert len(results) == 0
+
+            prev_offset_arg = "offset={}".format(current_offset - limit)
+            assert prev_offset_arg in response.data["previous"]
+
+            assert response.data["next"] is None
+
+    def test_get_pagination_attrs_for_offset_greater_than_count(self):
+        limit = 4
+        for _ in range(limit * 5):
+            UserContext()
+        with self.login(email=self.basic_user().email):
+            count = User.objects.count()
+            current_offset = count + 1
+            limit_arg = "limit={}".format(limit)
+            offset_arg = "offset={}".format(current_offset)
+            url = self.url + "?" + limit_arg + "&" + offset_arg
+            response = self.client.get(url)
+            results = response.data["results"]
+            assert len(results) == 0
+
+            prev_offset_arg = "offset={}".format(count - limit)
+            assert prev_offset_arg in response.data["previous"]
+
+            assert response.data["next"] is None
+
     def test_get_adjacent_offsets_has_unique_users(self):
         limit = 3
         for _ in range(limit * 3):
--- a/web/impact/impact/tests/test_user_list_view.py
+++ b/web/impact/impact/tests/test_user_list_view.py
@@ -108,16 +108,96 @@ class TestUserListView(APITestCase):
             assert user2.email in emails
             assert user_count == response.data["count"]
 
-    def test_get_with_limit(self):
-        UserContext().user
-        UserContext().user
+    def test_get_returns_correct_count_attribute(self):
+        for _ in range(10):
+            UserContext()
         with self.login(email=self.basic_user().email):
-            url = self.url + "?limit=1"
             user_count = User.objects.count()
+            response = self.client.get(self.url)
+            assert user_count == response.data["count"]
+
+    def test_get_with_limit_returns_correct_number_of_results(self):
+        limit = 1
+        for _ in range(limit * 3):
+            UserContext()
+        with self.login(email=self.basic_user().email):
+            url = self.url + "?limit={}".format(limit)
             response = self.client.get(url)
             results = response.data["results"]
-            assert 1 == len(results)
-            assert user_count == response.data["count"]
+            assert limit == len(results)
+
+    def test_get_correct_pagination_attributes_for_offset_zero(self):
+        limit = 3
+        current_implicit_offset = 0
+        for _ in range(limit * 3):
+            UserContext()
+        with self.login(email=self.basic_user().email):
+            limit_arg = "limit={}".format(limit)
+            url = self.url + "?" + limit_arg
+            response = self.client.get(url)
+            results = response.data["results"]
+            assert limit == len(results)
+            assert response.data["previous"] is None
+            assert limit_arg in response.data["next"]
+            next_offset_arg = "offset={}".format(
+                current_implicit_offset + limit)
+            assert next_offset_arg in response.data["next"]
+
+    def test_get_pagination_attrs_for_offset_between_zero_and_limit(self):
+        limit = 4
+        current_offset = limit - 2
+        for _ in range(limit * 3):
+            UserContext()
+        with self.login(email=self.basic_user().email):
+            limit_arg = "limit={}".format(limit)
+            offset_arg = "offset={}".format(current_offset)
+            url = self.url + "?" + limit_arg + "&" + offset_arg
+            response = self.client.get(url)
+            results = response.data["results"]
+            assert limit == len(results)
+
+            assert response.data["previous"] is not None
+            assert "offset" not in response.data["previous"]
+
+            assert limit_arg in response.data["next"]
+            next_offset_arg = "offset={}".format(current_offset + limit)
+            assert next_offset_arg in response.data["next"]
+
+    def test_get_pagination_attrs_for_offset_in_the_middle(self):
+        limit = 4
+        current_offset = limit + 2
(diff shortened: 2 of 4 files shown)


web/impact/impact/models/utils.py
```diff
@@ -28,3 +27,11 @@ def model_name_to_snake(value):
 
 def snake_to_model_name(value):
     return "".join(map(str.title, value.split("_")))
+
+
+def is_int(s):
```
pbx, line 32: A little bit of bikeshedding - I like this function but the name is potentially misleading. 

    >>> is_int(3.14159)
    True

Something like `can_cast_to_int` would be more precise. Not a blocker but it would make me happy.

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -108,16 +114,257 @@ def test_get(self):
             assert user2.email in emails
             assert user_count == response.data["count"]
 
-    def test_get_with_limit(self):
-        UserContext().user
-        UserContext().user
+    def test_get_returns_correct_count_attribute(self):
+        for _ in range(10):
+            UserContext()
```
pbx, line 120: Why not `UserContext.create_batch(10)`?

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -108,16 +114,257 @@ def test_get(self):
...
         with self.login(email=self.basic_user().email):
-            url = self.url + "?limit=1"
             user_count = User.objects.count()
+            response = self.client.get(self.url)
+            assert user_count == response.data["count"]
+
+    def test_get_with_limit_returns_correct_number_of_results(self):
+        limit = 1
+        for _ in range(limit * 3):
+            UserContext()
```
pbx, line 129: As above. Can we use `.create_batch` here?

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -108,16 +114,257 @@ def test_get(self):
...
-        UserContext().user
-        UserContext().user
+    def test_get_returns_correct_count_attribute(self):
+        for _ in range(10):
+            UserContext()
         with self.login(email=self.basic_user().email):
-            url = self.url + "?limit=1"
             user_count = User.objects.count()
+            response = self.client.get(self.url)
+            assert user_count == response.data["count"]
```
pbx, line 124: Are you going with `assert` instead of `self.assertEqual` and the like because the file already uses that style exclusively? I prefer the latter style because the failure messages are more informative. But I think I've seen you post similar review comments in the past so perhaps there's a good reason! Not a blocker.

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -108,16 +114,257 @@ def test_get(self):
...
             results = response.data["results"]
-            assert 1 == len(results)
-            assert user_count == response.data["count"]
+            assert limit == len(results)
+
+    def test_get_correct_pagination_attributes_for_offset_zero(self):
+        limit = 3
+        current_implicit_offset = 0
+        for _ in range(limit * 3):
+            UserContext()
```
pbx, line 140: `.create_batch`?

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -108,16 +114,257 @@ def test_get(self):
...
+            assert limit_arg in response.data["next"]
+            next_offset_arg = "offset={}".format(
+                current_implicit_offset + limit)
+            assert next_offset_arg in response.data["next"]
+
+    def test_get_pagination_attrs_for_offset_between_zero_and_limit(self):
+        limit = 4
+        current_offset = limit - 2
+        for _ in range(limit * 3):
+            UserContext()
```
pbx, line 157: `.create_batch`?

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -108,16 +114,257 @@ def test_get(self):
...
+
+            assert limit_arg in response.data["next"]
+            next_offset_arg = "offset={}".format(current_offset + limit)
+            assert next_offset_arg in response.data["next"]
+
+    def test_get_pagination_attrs_for_offset_in_the_middle(self):
+        limit = 4
+        current_offset = limit + 2
+        for _ in range(current_offset + limit + 2):
+            UserContext()
```
pbx, line 177: `.create_batch`?

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -108,16 +114,257 @@ def test_get(self):
...
+            assert prev_offset_arg in response.data["previous"]
+
+            assert limit_arg in response.data["next"]
+            next_offset_arg = "offset={}".format(current_offset + limit)
+            assert next_offset_arg in response.data["next"]
+
+    def test_get_pagination_attrs_for_offset_between_count_and_limit(self):
+        limit = 4
+        for _ in range(limit * 5):
+            UserContext()
```
pbx, line 196: `.create_batch`?

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -108,16 +114,257 @@ def test_get(self):
...
+
+            prev_offset_arg = "offset={}".format(current_offset - limit)
+            assert prev_offset_arg in response.data["previous"]
+
+            assert response.data["next"] is None
+
+    def test_get_pagination_attrs_for_offset_equals_number_of_results(self):
+        limit = 4
+        for _ in range(limit * 5):
+            UserContext()
```
pbx, line 215: `.create_batch`?

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -108,16 +114,257 @@ def test_get(self):
...
+
+            prev_offset_arg = "offset={}".format(current_offset - limit)
+            assert prev_offset_arg in response.data["previous"]
+
+            assert response.data["next"] is None
+
+    def test_get_pagination_attrs_for_offset_greater_than_num_of_results(self):
+        limit = 4
+        for _ in range(limit * 5):
+            UserContext()
```
pbx, line 234: `.create_batch`?

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -108,16 +114,257 @@ def test_get(self):
...
+            assert len(results) == 0
+
+            prev_offset_arg = "offset={}".format(count - limit)
+            assert prev_offset_arg in response.data["previous"]
+
+            assert response.data["next"] is None
+
+    def test_get_pagination_attrs_for_limit_greater_than_num_of_results(self):
+        for _ in range(5):
+            UserContext()
```
pbx, line 252: `.create_batch`?

web/impact/impact/tests/test_user_list_view.py
```diff
@@ -108,16 +114,257 @@ def test_get(self):
...
+            url = self.url + "?" + offset_arg
+            response = self.client.get(url)
+            assert response.status_code == 200
+            implicit_zero_response = self.client.get(self.url)
+            assert response.data == implicit_zero_response.data
+
+    def test_get_adjacent_offsets_has_unique_users(self):
+        limit = 3
+        for _ in range(limit * 3):
+            UserContext()
```
pbx, line 357: `.create_batch`?

pbx (changes requested):
I like this. Manual testing against development confirmed all the fixes. Tests are nice and thorough. Only a couple concerns (one of them repeated many times for completeness' sake).

web/impact/impact/v1/views/base_list_view.py
```diff
@@ -32,18 +39,45 @@ def metadata(self):
...
             "count": count,
-            "next": _url(base_url, limit, offset + limit),
-            "previous": _url(base_url, limit, offset - limit),
+            "next": _next_url(base_url, limit, offset, count),
+            "previous": _previous_url(base_url, limit, offset, count),
             "results": results,
         }
         return Response(result)
 
+    def _validate_kwarg(self, val, key):
```
mc-nathan: This would be simpler as two functions, say `_get_limit(request)` and `_get_offset(request)`

web/impact/impact/models/utils.py
yotammanor: Actually I'd expect `is_int(3.14159)` to be `False`. fixing.

web/impact/impact/v1/views/base_list_view.py
```diff
@@ -32,18 +39,45 @@ def metadata(self):
...
+        if key == "limit":
+            self._validate_limit(key, val)
+        if key == "offset":
+            self._validate_offset(key, val)
+        return int(val)
+
+    def _validate_offset(self, key, val):
+        if int(val) < 0:
+            self.errors.append(
+                KWARG_VALUE_IS_NON_POSITIVE_ERROR.format(key))
```
mc-nathan: Technically 0 is not a positive integer and it seems like it would be better to have two error messages since the two places this is used are different.  I would say something that doesn't hinge on the user understanding exactly what you mean by "positive integer" like "value of '{}' should be greater than zero" or "value of '{}" should be greater than or equal to zero".

web/impact/impact/tests/test_user_list_view.py
yotammanor: `UserContext` doesn't have a `create_batch` since it is not a subclass of `Factory`. and likewise.

web/impact/impact/tests/test_user_list_view.py
yotammanor: for consistency's sake. plus my hope now is that we'll change to pytest, which produces informative errors with regular assert statements.

mc-nathan (changes requested):
Overall looks good.  In addition to Paul's comments I spotted a couple more relatively minor things, but I think they are worth cleaning up.

web/impact/impact/tests/test_user_list_view.py
pbx: Ah. I was fooled by the fact that `JudgeFeedbackContext` has a `create_batch`; turns out it's manually defined on the class.

pbx (approved):

mc-nathan (approved):

PR #86, opened 2017-12-20
[AC-5209] Refactor of problematic condition and method_to_perm

mc-nathan:
Work done:
- Minor refactor to address "C" rated violation for CodeClimate.

No functionality should have changed.

### DIFF
--- a/web/impact/impact/permissions.py
+++ b/web/impact/impact/permissions.py
@@ -14,6 +14,16 @@ from impact.utils import model_name_case
 
 User = get_user_model()
 
+METHOD_TO_ACTION = {
+    "POST": "add",
+    "GET": "view",
+    "HEAD": "view",
+    "OPTIONS": "view",
+    "PUT": "change",
+    "PATCH": "change",
+    "DELETE": "delete",
+}
+
 
 class V0APIPermissions(BasePermission):
     authenticated_users_only = True
@@ -39,17 +49,11 @@ class V1ConfidentialAPIPermissions(BasePermission):
             name=settings.V1_CONFIDENTIAL_API_GROUP).exists()
 
 
-class DynamicModelPermissions(BasePermission):
-    perms_map = {
-        'GET': ['%(app)s.view_%(model_name)s'],
-        'OPTIONS': ['%(app)s.view_%(model_name)s'],
-        'HEAD': ['%(app)s.view_%(model_name)s'],
-        'POST': ['%(app)s.add_%(model_name)s'],
-        'PUT': ['%(app)s.change_%(model_name)s'],
-        'PATCH': ['%(app)s.change_%(model_name)s'],
-        'DELETE': ['%(app)s.delete_%(model_name)s'],
-    }
+def method_to_perm(method):
+    return ["%(app)s.{}_%(model_name)s".format(METHOD_TO_ACTION[method])]
 
+
+class DynamicModelPermissions(BasePermission):
     authenticated_users_only = True
 
     def has_permission(self, request, view):
@@ -58,7 +62,7 @@ class DynamicModelPermissions(BasePermission):
         model_name = model_name_case(model, related_model)
         app_label = 'mc'
         kwargs = {'app': app_label, 'model_name': model_name}
-        perms = [perm % kwargs for perm in self.perms_map[request.method]]
+        perms = [perm % kwargs for perm in method_to_perm(request.method)]
         return (
             request.user and (
                 request.user.is_authenticated or (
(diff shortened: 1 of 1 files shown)


web/impact/impact/permissions.py
```diff
@@ -39,17 +49,11 @@ def has_permission(self, request, view):
...
-        'GET': ['%(app)s.view_%(model_name)s'],
-        'OPTIONS': ['%(app)s.view_%(model_name)s'],
-        'HEAD': ['%(app)s.view_%(model_name)s'],
-        'POST': ['%(app)s.add_%(model_name)s'],
-        'PUT': ['%(app)s.change_%(model_name)s'],
-        'PATCH': ['%(app)s.change_%(model_name)s'],
-        'DELETE': ['%(app)s.delete_%(model_name)s'],
-    }
+def method_to_perm(method):
+    return ["%(app)s.{}_%(model_name)s".format(METHOD_TO_ACTION[method])]
```
shanbady: maybe we should do METHOD_TO_ACTION.get(method, 'get') and default to 'get'

shanbady (changes requested):

web/impact/impact/permissions.py
```diff
@@ -87,14 +91,7 @@ def decompose_perm(self, perm_codename):
...
     def action_matches(self, action, request):
-        if ((request.method in ('POST') and action == 'add') or (
-            request.method in ('GET', 'HEAD', 'OPTIONS') and (
-                action == 'view')) or (
-            request.method in (
-                'PUT', 'PATCH') and (
-                action == 'change')) or (
-                request.method in ('DELETE') and (action == 'delete'))):
-            # the permission applies to our case
+        if action == METHOD_TO_ACTION.get(request.method):
```
pbx, line 94: Wonderful to see that hairy old conditional obsoleted!

web/impact/impact/permissions.py
pbx: It's a little mind-bending to have both old- and new-style string interpolation used in a single line. I'd prefer it to be all new-style.

web/impact/impact/permissions.py
```diff
@@ -58,7 +62,7 @@ def has_permission(self, request, view):
         model_name = model_name_case(model, related_model)
         app_label = 'mc'
         kwargs = {'app': app_label, 'model_name': model_name}
-        perms = [perm % kwargs for perm in self.perms_map[request.method]]
+        perms = [perm % kwargs for perm in method_to_perm(request.method)]
```
pbx: The list comp, coupled with `method_to_perm` always returning a single-item list, seems unnecessary.

pbx (changes requested):
Functionally this looks good. Just a couple code style concerns.

web/impact/impact/permissions.py
mc-nathan: We could.  I didn't want to change the original behavior which was `self.perms_map[request.method]`.  I don't expect it actually matters, but it was my reasoning.  Don't have a strong opinion.  Happy to change it if you want since it should be a bit safer and I'd be surprised if anything in this code is depending on it raising an exception if the method is missing.

mc-nathan: I had trouble the last time I tried to do one replacement without affecting the other things that will be replaced later when I used format.  I expect there's some trick for it, but I gave up searching before I found it.  Do you know how to do that?  The other option would be to just use `+`, but I'm not sure that's any better that what I'm doing now.

web/impact/impact/permissions.py
mc-nathan: Good point.  Again I was just trying not to change the existing behavior, but that one seems even less controversial.  Fixed.

web/impact/impact/permissions.py
mc-nathan: Actually 'get' is not a good default since this is create a Django permission name.  For "get" the value is "view" so we could use that.  However, given that this is permission stuff, I think I'd rather throw an exception if we get something unexpected.  Won't fix (unless you make a really strong argument).

mc-nathan: Found this on SO, but I'm not lovin' it: https://stackoverflow.com/questions/11283961/partial-string-formatting

pbx: Aha yeah, I think `string.Template` is the only built-in way to do this. Not a win.
```python
>>> from string import Template
>>> t1 = Template("${app}.${action}_${model_name}")
>>> t2 = t1.safe_substitute(action="edit")
>>> print Template(t2).safe_substitute(app="admin", model_name="logentry")
admin.edit_logentry
```
Aesthetically I'd prefer concatenation to the interpolation mashup, but that preference isn't strong enough to delay approving this.

pbx (approved):

shanbady (approved):

PR #87, opened 2017-12-21
[AC-5111] - Add SSO Token Generation, Verification and Refresh to impact-api

yotammanor:
**Changes Introduced in [AC-5111](https://masschallenge.atlassian.net/browse/AC-5111):**

- add drf-jwt as dependency
- register urls for the following views:
  - obtain a jwt
  - verify a jwt
  - refresh a jwt
- configure drf-jwt in settings:
  - reasonable expiration deltas
  - a secret key (the production one is a secret travis env var.
  - work with JWTs in cookies.
- cleanup impact.urls.py (for clarity of urls.py chosen structure).

**Test:**
- make build;make dev;make load-remote-db;make grant-permissions (with v1_clients);
- update to the following [Postman Collection](https://www.getpostman.com/collections/0a95773f88df764b15ee)
- open a sample V1 call in postman, e.g. "v1 Industry list get" . Try to access it as is, get a 403.
- open sso/obtain-token request.
 - send incorrect username and password. see a 400 error
 -  send it with the correct username and password. see a 200 status, and the token
   - see that now there's a cookie added to "Cookies", with the same token.
- now reopen "v1 Industry list get":
  - send it. it should still fail (401), because you're also sending a (bad) token from our existing mechanism.
  - Go to the request "Headers" section. uncheck "Authorization". resend the request. It now succeeds.
- now create an application, and set the client_id in the "oauth token" request, and send the request (the old mechanism).
  - reselect the checkbox for Authorization in "v1 Industry list get", and send the request. it is successful. unselect, send, still successfull (the new and old mechanisms are compatible).

- in settings.py, change the value of JWT_EXPIRATION_DELTA to be very short (a few seconds). get a new token. wait that timedelta. try to use it. see that it fails (401). try to use this token in the "verify" call. see a 400 error.
- create a new token. before time is up, refresh it. see that it succeeds, and you can use it beyond the time you set - but no more than the time specified JWT_REFRESH_EXPIRATION_DELTA (set it to be short as well. Try to refresh after that time has passed. fails).

- review urlpatterns naming choices, and suggest improvements if there are any.
- review and groom https://masschallenge.atlassian.net/browse/AC-5110 and https://masschallenge.atlassian.net/browse/AC-5109 (in that order)


web/impact/impact/v1/urls.py
```diff
@@ -0,0 +1,83 @@
+from django.conf.urls import url
```
mc-nathan, line 1: +1 I like having these in separate files.

mc-nathan (approved):
Functionality looks good, Travis is happy, CodeClimate is happy, code looks good.  Approved.

web/impact/impact/settings.py
```diff
@@ -199,6 +200,18 @@ class Base(Configuration):
         'simpleuser.email_model_backend.EmailModelBackend',
         'django.contrib.auth.backends.ModelBackend',
     )
+    JWT_AUTH = {
+        'JWT_ALLOW_REFRESH': True,
+        'JWT_EXPIRATION_DELTA': datetime.timedelta(days=7),
```
shanbady, line 205: wondering if this window is too long as a default. I was thinking a few hours seems more reasonable.

web/impact/impact/urls.py
```diff
@@ -8,156 +8,50 @@
...
     if model._meta.app_label == 'impact' and not model._meta.auto_created:
         schema_router.register(model, url=model_name_to_snake(model.__name__))
 
 for model in apps.get_models('accelerator'):
-    if (
-        model._meta.app_label == 'accelerator'
-        and hasattr(model, "Meta")
-        and not model._meta.auto_created
-    ):
+    if model_is_not_auto_created(model, app_label='accelerator'):
```
shanbady, line 36: +1 for this

shanbady (approved):
giving this an approve with a non-blocking comment about defaults.

web/impact/impact/settings.py
mc-nathan: The reason we settled on this value is that I believe it will effect how long users remain logged into accelerate.  Currently I don't believe there is a time limit on that so we decided we wanted to make it as long as is reasonable.  A week between logins seems reasonable to me given what we have don't historically.  A few hours seems too short to me.

PR #90, opened 2018-01-03
[AC-5211] - Production Ready Impact-api Cluster - Register Correctly on Newrelic

yotammanor:
**Changes Introduced:**

- Added a separate environment name for the "temp" cluster.

**Test:**
- Has no effect at the moment (will have only after merge with master).
  - post-merge, I'll create a release, and then reconfigure the "temp" cluster to use this new environment. After that, New Relic should have this new Environment available and collecting data from temp.api.masschallenge.org

### DIFF
--- a/web/impact/newrelic.ini
+++ b/web/impact/newrelic.ini
@@ -210,7 +210,7 @@ app_name = MC Impact API Production
 monitor_mode = true
 
 [newrelic:temp]
-app_name = MC Impact API Temp (Production DB)
+app_name = MC Impact API Temp
 monitor_mode = true
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


web/impact/newrelic.ini
```diff
@@ -209,4 +209,8 @@ monitor_mode = true
 app_name = MC Impact API Production
 monitor_mode = true
 
+[newrelic:temp]
+app_name = MC Impact API Temp (Production DB)
```
pbx: This is utter bikeshedding (i.e. non-blocking), but I'd rather see it called just "MC Impact API Temp". I can't think of how that parenthetical will be useful, but I *can* imagine it being confusing if the `temp` host sticks around longer than this ticket's project.

pbx (approved):

web/impact/newrelic.ini
yotammanor: done

PR #91, opened 2018-01-11
[AC-5038-impact] Configure impact-api to manage models, add initial migration

jonkiparsky:
See AC-5038-mcproject for review information


yotammanor (commented):
Blocking comment here, some more below

web/impact/requirements/base.txt
```diff
@@ -36,7 +36,7 @@ djangorestframework
 djangorestframework-jwt
 drf-schema-adapter
 drf-tracking
-git+https://github.com/masschallenge/django-accelerator.git@2017.11.09#egg=django-accelerator
+git+https://github.com/masschallenge/django-accelerator.git@AC-5038-njw#egg=django-accelerator
```
yotammanor: Obviously this needs to change before we approve.

web/impact/impact/models/utils.py
```diff
@@ -8,7 +8,7 @@
 
 
 def is_managed(db_table):
-    return False
+    return True
```
yotammanor: I don't like the fact that the boolean flag is configured in settings.py in one project and is set literally in the other.

yotammanor (changes requested):
Functional review is in the mcproject branch.

Some comments:
- We should probably set codeclimate to ignore migrations somehow. I'm ok with ignoring it manually for the moment.
- We should rename `0002_auto...` to a more meaningful name.

jonkiparsky:
re: migration names, clearly we're going to have to improve `make migrations` to take the option set that `./manage.py makemigrations` allows

yotammanor:
@jonkiparsky my suggestion is that we should just have a `make manage` that allows for an arbitrary manage.py command, with its args and kwargs as needed. Here: https://masschallenge.atlassian.net/browse/AC-4598?focusedCommentId=37873&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-37873

yotammanor:
@jonkiparsky @mc-nathan closing this PR, and renaming AC-5038-impact to `modular-models-epic`

PR #94, opened 2018-01-24
[AC-5296] ripping out logging

shanbady:
in the interest of expediency (and because we may re-introduce the logging at some point) i've simply tugged out the mixin from all the views to disable logging to the database. we should get this out as soon as we possible (we will run out of disk space soon on the db prod api is pointed at)


yotammanor (approved):
Nice. MAW.

jonkiparsky (approved):
Looks fine.

PR #99, opened 2018-01-30
[AC-5266] Impact API should read from a read replica in production.

shanbady:
The changes in this branch allow and environment to specify a 'READ_REPLICA_DATABASE_URL' as an environment variable which enables the 'read-from-replica' feature (leveraging the [django-multidb-router](https://github.com/jbalogh/django-multidb-router) library. 

To try this out locally you can specify a replica database in the .dev.env file as such ```READ_REPLICA_DATABASE_URL=mysql://root:root@mysql/mc_dev_replica``` and the re-run ```make clean``` and ```make dev```. 

I currently have this branch deployed and reading from the salesforce testing database master and replica here so folks can poke at a working setup (and to make this review a bit easier): http://174.129.53.100/


web/impact/impact/routers.py
```diff
@@ -18,3 +24,31 @@ def allow_migrate(self, db, app_label, model_name=None, **hints):
...
+            model_name,
+            **hints)
+
+        if app_label in [
+                'auth',
+                'admin',
+                'sessions',
+                'simpleuser',
+                'sites',
+                'contenttypes']:
```
yotammanor, line 52: I'd prefer to see this if statement dried out. But since we're removing it soon (@jonkiparsky correct me if I'm wrong about this assumption), not blocking.

yotammanor (changes requested):
The code looks reasonable, barring for one non-blocking comment.

Functionally, I went through all of our ECS clusters and couldn't find on that has a task definition with a JSON that defines a `READ_REPLICA_DATABASE_URL`. 

Likewise, I wasn't able to create a local environment that has two databases, it is not as simple as adding the URL in .dev.env, because it seems that the `mysql` image does not support by multiple DBs out of the box - we'd need to customize the container with our own startup script, or add two containers based on the same image (with port mapping etc.).
I'm not sure it's important that we have a multi-db dev env "right away", but we should make a conscious decision about how we want to continue with this (especially re configuration of automatic deployment, but also re local testing of this behaviour. Then either do something about it, or create follow on ticket(s).

shanbady:
@yotammanor the "test" cluster has it set in the task definition: https://console.aws.amazon.com/ecs/home?region=us-east-1#/taskDefinitions/ecscompose-impact-api/1049

shanbady:
In regards to testing this locally - what i did to 'simulate' a read-replica setup was basically create a duplicate database within the same running mysql instance that had a copy of the mc_dev database. I then simply set the read-replica environment variable in .dev.env to point to this database ```READ_REPLICA_DATABASE_URL=mysql://root:root@mysql/mc_dev_replica``` - it exercises the functionality needed which is essentially - reads go to one database - writes go to another. in production of course this would be an actual read-replica.

mc-nathan (changes requested):
I was not able to get to the point of being able to demonstrate that the system was successfully accessing data from another database.

There needs to be a clear way to setup a replica on a local system for testing.

The approach I took was to create a copy of the mc_dev database in the same MySQL container named mc_dev_replica.  I then pointed it at this second database as the read replica.  This allowed me to bring up the app, but when I tried to log in it failed with an authentication error.  If switch READ_REPLICA_DATABASE_URL to point to mc_dev, then everything works fine.

http://174.129.53.100/ seems to work, but I can't figure out anyway to verify that the read replica is actually being used.

web/impact/impact/routers.py
yotammanor: Just for protocol - 
This list appears in: https://github.com/masschallenge/impact-api/pull/99/files#diff-4fbc2fef212daea86092e335e9888262R18

and in: https://github.com/masschallenge/impact-api/pull/99/files#diff-4fbc2fef212daea86092e335e9888262R46

shanbady:
@mc-nathan @yotammanor 
I now have a locally testable replica setup. I created this in a separate branch off of this one that includes those changes.

in the ```AC-5266-with-read-replica``` branch, docker-compose brings up 2 separate mysql containers - one which acts as the primary and referred to in .dev.env by DATABASE_URL (https://github.com/masschallenge/impact-api/blob/AC-5266-with-read-replica/.dev.env#L5) and also a slave which is referred to by the READ_REPLICA_DATABASE_URL environment variable (https://github.com/masschallenge/impact-api/blob/AC-5266-with-read-replica/.dev.env#L8). 

**To test this:**

1. checkout the branch  AC-5266-with-read-replica
2. run ```make clean```
3. run ```make build```
4. run ```make dev```
5. run ```make load-db GZ_FILE=db_cache/initial_schema.sql.gz```
6. open up a separate terminal instance and tail the replica mysql instance's log ```tail -f replica.out```

7. login to impact and observe the read activity in the replica logs. it should look like this:
```
2018-02-05T16:13:54.079490Z	    5 Connect	root@impactapi_web_1.impactapi_default on mc_dev using TCP/IP
2018-02-05T16:13:54.081276Z	    5 Query	SET NAMES utf8
2018-02-05T16:13:54.082607Z	    5 Query	set autocommit=0
2018-02-05T16:13:54.084190Z	    5 Query	set autocommit=1
2018-02-05T16:13:54.087197Z	    5 Query	SELECT @@SQL_AUTO_IS_NULL
2018-02-05T16:13:54.938016Z	    5 Query	SELECT VERSION()
2018-02-05T16:13:54.944034Z	    5 Quit	
2018-02-05T16:13:54.945557Z	    6 Connect	root@impactapi_web_1.impactapi_default on mc_dev using TCP/IP
2018-02-05T16:13:54.946297Z	    6 Query	SET NAMES utf8
2018-02-05T16:13:54.946995Z	    6 Query	set autocommit=0
2018-02-05T16:13:54.947660Z	    6 Query	set autocommit=1
2018-02-05T16:13:54.948421Z	    6 Query	SELECT `auth_user`.`id`, `auth_user`.`password`, `auth_user`.`last_login`, `auth_user`.`is_superuser`, `auth_user`.`username`, `auth_user`.`first_name`, `auth_user`.`last_name`, `auth_user`.`email`, `auth_user`.`is_staff`, `auth_user`.`is_active`, `auth_user`.`date_joined` FROM `auth_user` WHERE `auth_user`.`email` = 'root'
2018-02-05T16:13:54.963126Z	    6 Query	SELECT `auth_user`.`id`, `auth_user`.`password`, `auth_user`.`last_login`, `auth_user`.`is_superuser`, `auth_user`.`username`, `auth_user`.`first_name`, `auth_user`.`last_name`, `auth_user`.`email`, `auth_user`.`is_staff`, `auth_user`.`is_active`, `auth_user`.`date_joined` FROM `auth_user` WHERE `auth_user`.`username` = 'root'
2018-02-05T16:13:55.036689Z	    6 Query	SELECT `django_site`.`id`, `django_site`.`domain`, `django_site`.`name` FROM `django_site` WHERE `django_site`.`id` = 1
2018-02-05T16:13:55.229935Z	    6 Quit	
```

yotammanor (approved):

mc-nathan (approved):

mc-nathan:
Worked for me as well.  Merging!

PR #100, opened 2018-02-01
[AC-5291] API Request: Credit Code Support (probably new call)

knittingarch:
_**N.B.** The Postman Collection still needs to be created, but I wanted to get this out for review while I figure that out._

**To Test**
**Set Up the Environment**
- Run `make build; make dev`.
- In another terminal window, run `make load-remote-db DB_FILE_NAME=initial_schema.sql.gz`.
- Then run `make grant-permissions PERMISSION_USER=demoadmin@masschallenge.org PERMISSION_CLASSES=v1_clients`.

**Check Calls - List View and Detail View**
- Login in as demoadmin@masschallenge.org.
- Go to [http://127.0.0.1:8000/api/v1/credit_code/](http://127.0.0.1:8000/api/v1/credit_code/).
- You should see a LIST of refund codes with the following fields:
  - PK
  - Unique code
  - Programs ids
  - Issued to (Organization ID)
  - Discount
  - Maximum Uses
  - Notes
  - Internal
- Go to [http://127.0.0.1:8000/api/v1/credit_code/44996/](http://127.0.0.1:8000/api/v1/credit_code/44996/).
   - You should see the DETAIL view for this particular code.
   - The same fields should be present.


pbx (changes requested):
Functionally this looks fine to me, works as described!

I looked at the two Codeclimate complaints about code duplication and they seem off-base to me (i.e. ignorable).

My blocking concern is that, though [the ticket](https://masschallenge.atlassian.net/browse/AC-5291) uses the term **"Credit Code"**, the code here uses the deprecated "Refund Code" name.

If we add an endpoint to the public API that uses the deprecated naming, it will make [the renaming ticket work (AC-4027)](https://masschallenge.atlassian.net/browse/AC-4027) harder since we will have to accommodate clients using the old name.

mc-nathan:
I talked with Sarah about the CodeClimate warnings.  Those seem reasonable and not hard to address to me.

knittingarch:
@pbx I spoke with @jonkiparsky about this and he agreed that we needed to use Refund Code.  Everything currently in impact-api uses the old names...  Models will be the last thing to change.  I think it will be much more confusing to use Credit Code and Refund Code in the same files...

knittingarch:
@mc-nathan No go.  The Climate is still not appeased...

web/impact/impact/v1/helpers/model_helper.py
```diff
@@ -189,10 +198,8 @@ def save(self):
     def all_objects(cls):
         return cls.model.objects.all()
 
-    def field_name(self, field):
-        obj = getattr(self.subject, field, None)
-        if hasattr(obj, "name"):
-            return obj.name
+    def get_field_value(self, field):
```
mc-nathan: This needs a different name.  "get_" should only be used to refer explicitly to things related to "GET" calls (e.g., get_description in profile_helper.py).

web/impact/impact/v1/helpers/profile_helper.py
```diff
@@ -415,14 +415,13 @@ def functional_expertise(self):
 
     @property
     def expert_category(self):
-        return self.field_name("expert_category")
+        return self.get_field_value("expert_category").name
```
mc-nathan: This will fail if the getattr in the function returns None.  I would expect something like `self.field_property("expert_category", "name")`

web/impact/impact/v1/helpers/program_helper.py
```diff
@@ -36,8 +36,8 @@ def fields(cls):
 
     @property
     def cycle_name(self):
-        return self.field_name("cycle")
+        return self.get_field_value("cycle").name
```
mc-nathan: See above.

mc-nathan (changes requested):
Needs to go a bit further, but coming along nicely.

mc-nathan:
Hmm, I agree with @pbx and not with @jonkiparsky. I would rather this was CreditCode than RefundCode.  The API will be with us for a long time and we should name things the way we really want them named.  This is why we only have Organizations and not Startups and why we does have separate calls for Entrepreneurs and Experts.

jonkiparsky:
@sdawson I don't think I agree with that. I said we should not try to change the model name in this ticket, but we should try to use "credit" in preference to "refund" wherever possible and certainly in user-facing language.

web/impact/impact/tests/test_refund_code_detail_view.py
```diff
@@ -0,0 +1,60 @@
...
+    "unique_code",
+    "discount",
+    "maximum_uses",
+    "programs",
+]
+
+
+class TestRefundCodeDetailView(APITestCase):
+    def test_get(self):
+        code = RefundCodeFactory()
```
mc-nathan: Don't change.

web/impact/impact/tests/test_refund_code_detail_view.py
```diff
@@ -0,0 +1,60 @@
...
+        with self.login(email=self.basic_user().email):
+            url = reverse(RefundCodeDetailView.view_name,
```
mc-nathan: Change.

web/impact/impact/tests/test_refund_code_list_view.py
```diff
@@ -0,0 +1,46 @@
+# MIT License
```
mc-nathan: Change file name.

web/impact/impact/v1/helpers/__init__.py
```diff
@@ -69,6 +69,10 @@
     PROGRAM_FAMILY_FIELDS,
     ProgramFamilyHelper,
 )
+from impact.v1.helpers.refund_code_helper import (
```
mc-nathan: Change.

web/impact/impact/v1/helpers/refund_code_helper.py
```diff
@@ -0,0 +1,52 @@
+# MIT License
```
mc-nathan: Change file name.

web/impact/impact/v1/helpers/refund_code_helper.py
```diff
@@ -0,0 +1,52 @@
...
+# Copyright (c) 2017 MassChallenge, Inc.
+
+from impact.models import RefundCode
```
mc-nathan: Don't change.

mc-nathan:
I went through the PR search for 'refund'.  The only ones I wouldn't change are the impact.models.RefundCode and impact.tests.factories.RefundCodeFactory.

Added some review comments following that pattern.

mc-nathan (changes requested):

knittingarch:
@pbx @jonkiparsky @mc-nathan Roger that!  Will make the name changes.  Sorry that I wasn't clear with what I was asking before I started working on this one, @jonkiparsky!

web/impact/impact/v1/helpers/profile_helper.py
knittingarch: @mc-nathan I tried something like this first, but I got an error that I was passing in too many values to the `getattr`...  Gonna read up on it now.

pbx (approved):

pbx:
Confirmed the updated `credit_code` endpoints work.  (Updated the example links too.) 👍

mc-nathan (approved):

PR #102, opened 2018-02-08
Ac 5266 - hotfix for default (non read-replica) configuration

shanbady:
It turns out the django-multidb-router expects the SLAVE_DATABASES setting to be available as soon as the router is imported. This commit adds this variable with a default value so we dont error out.

to test:

- running make dev on the latest development and trying to login will cause a 500.
- doing the same on this branch does not.


web/impact/impact/settings.py
```diff
@@ -54,6 +54,8 @@ class Base(Configuration):
 
     DATABASE_ROUTERS = ['impact.routers.APIRouter']
 
+    SLAVE_DATABASES = ['default']
```
pbx, line 57: This is the right way to reference the `'default'` database alias, but I don't see how that alias is getting set in the main `DATABASES` configuration above. In the docs for [dj-database-url](https://pypi.python.org/pypi/dj-database-url/#usage) (which [django-configurations' DatabaseURL method](http://django-configurations.readthedocs.io/en/latest/values/#configurations.values.DatabaseURLValue) uses), the alias is assigned explicitly. Is that happening automatically here in some way I'm missing?

pbx (changes requested):

mc-nathan:
Fixes the immediate problem.  I'm curious as well about `['default']` rather than just `[]`, but happy to let @shanbady and @pbx sort that out.  Approving.

mc-nathan (approved):

web/impact/impact/settings.py
shanbady: this value is being set just to make the library happy when imported locally (in fact in reality we could omit a value here altogether). the django database url deserializes to set the database 'default' by default as well.

shanbady: also we are guarenteed to have a 'default' database since its required as per the django docs: https://docs.djangoproject.com/en/2.0/ref/settings/#std:setting-DATABASES
```The DATABASES setting must configure a default database;```

pbx: After discussing with Shankar i'm satisfied that  django-configurations' DatabaseURL method returns a full configuration dict, complete with a `default` entry. Approving.

pbx (approved):

PR #111, opened 2018-03-07
[AC-4598] New build environment

mc-nathan:
Work done:
- Complete revision of the Makefile and associated targets.
- All regular developer workflows up to release/deploy should be supported by make targets.

To test:
- Checkout AC-4598 branch for impact-api and django-accelerator repos in sibling directories
- Run `make help` to understand what the targets are and should do
- Run `make build` before you try any of the other targets.
- Run `make test` to make sure everything is working.
- Review all the other targets.


yotammanor (commented):
- *make delete-vms:* did not finish when it was done, just stayed stuck until I ^C it.

- *make run-server*: how do I see the `make dev` output I used to see on startup?

- *make grant-permissions:* I'd like to know what is the suggested alternative, since it is still needed for api local testing w. postman.

- *make load-remote-db:* I'd love it if there were still a way to enforce getting a new DB from the remote location in one lines instead of two. I also still think getting the remote should be the default behavior, although I see the costs of course.

mc-nathan:
`make delete-vms` - Not able to reproduce.  Worked fine for me.
`make run-server` - Setting debug=0 does this.  Added this to the help.
`make grant-permissions` - You should be able to do it directly from the accelerate admin interface.  If there are specific accounts we think should have these permissions all the time, then we should either change them on production or in the database cleaning process.
`make load-remote-db` - Added.

Makefile
```diff
@@ -103,130 +168,219 @@ setup:
...
+
+upload-db:
+	@if [ `wc -c < $(gz_file)` \> $(MAX_UPLOAD_SIZE) ]; \
+	then \
+	  echo Dump file exceeds $(MAX_UPLOAD_SIZE) bytes.; \
+	  echo This may indicate that this dump contains sensitive data; \
+	  false; \
+	fi
+	@echo Uploading $(gz_file) as $(s3_key)...
+	@aws s3api put-object --acl public-read --bucket public-clean-saved-db-states --key $(s3_key) --body $(gz_file)
```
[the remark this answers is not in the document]
yotammanor: Also, I want to suggest that we use the high-level api instead:
```
	aws s3 cp $(gz_file) s3://public-clean-saved-db-states/$(s3_key) --acl public-read
```
This command does exactly the same, but has a nice progress update included (and I assume a better error handling as well, although I haven't tried).

Makefile
```diff
@@ -1,95 +1,160 @@
...
+  test \
+  coverage \
+  coverage-html \
+  code-check \
+  \
+  update-schema \
+  data-migration \
+  migrations \
+  migrate \
+  models \
```
yotammanor: can't find a target with this name.. remove?

Makefile
```diff
@@ -1,95 +1,160 @@
...
+  stop-all-servers \
+  shutdown-all-vms \
+  delete-all-vms \
+  build-all \
+  \
+  bash-shell \
+  db-shell \
+  django-shell \
+  \
+  load-db \
```
yotammanor, line 32: thanks for adding back `load-remote-db`. If possible, it's better to add it to this list, as well :)

Makefile
```diff
@@ -1,95 +1,160 @@
...
+  dump-db \
+  upload-db \
+  clean-db-cache \
+  \
+  release-list \
+  release \
+  deploy \
+  \
+  update-packages \
```
yotammanor: can't find a target with this name.. remove? add a not-implemented yet msg?

web/impact/requirements/mc.txt
```diff
@@ -0,0 +1 @@
+git+https://github.com/masschallenge/django-accelerator.git@AC-4598#egg=django-accelerator
```
yotammanor, line 1: We're going to trip over the fact that we need to set the @branch in two different places. Maybe there's a way to set it once?

web/impact/requirements/travis.txt
```diff
@@ -0,0 +1,6 @@
+# MIT License
+# Copyright (c) 2017 MassChallenge, Inc.
+
+-r test.txt
+
+git+https://github.com/masschallenge/django-accelerator.git@AC-4598#egg=django-accelerator
```
yotammanor, line 6: sorry, three times.

travis.docker-compose.yml
```diff
@@ -0,0 +1,70 @@
+version: '2'
```
yotammanor, line 1: There's a lot of duplication between our different docker-compose files, which can probably be dried out by relying on https://docs.docker.com/compose/extends/

Follow on is good for the moment.

web/travis.Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM python:3.6
```
yotammanor, line 1: we can probably dry the dockerfile as well.

yotammanor (changes requested):
Got now to try all of the targets. Most are working great and looking great. A few comments though.

web/impact/requirements/mc.txt
mc-nathan: This is a deploy issue that is supposed to be resolved by Shankar's ticket.  Won't fix.

web/impact/requirements/travis.txt
mc-nathan: Again.

travis.docker-compose.yml
yotammanor: https://masschallenge.atlassian.net/browse/AC-5407

web/travis.Dockerfile
yotammanor: https://masschallenge.atlassian.net/browse/AC-5408

Makefile
```diff
@@ -191,16 +189,16 @@ coverage-report: diff_grep:=$(shell echo $(diff_sed) | tr ' ' '\n' | grep \.py |
 coverage-report:
 	@docker-compose run --rm web coverage report -i --skip-covered $(diff_grep) | grep -v "NoSource:"
 
-coverage-html:
+coverage-html-report:
 	@docker-compose run --rm web coverage html --omit="*/tests/*"
 
-coverage-html-open: coverage-html
+coverage-html: coverage
```
yotammanor: Not sure this is a correct dependency. I might want just to open it, without running it again

yotammanor (approved):
One small non-blocking comment. MAW.

Makefile
mc-nathan: Sounds like you are proposing a different target that just opens the latest coverage report.  I still think we want one that runs the coverage and then opens the result (if you like that kind of thing).  We could do that I suppose.  Personally when I want to do that I just go to File => Open in the browser.  The only reason I implemented "coverage-html" is that Shankar did it on impact.  Personally I'm fine with just opening it from the browser.

yotammanor: I guess I am too. Doesn't matter much.

knittingarch:
Whoops!  Got confused about where to put all my feedback.  Will check the `make coverage-html` target once this branch is testable again.

knittingarch (approved):
`make coverage-html` now works!  Everything else looks good!

PR #114, opened 2018-03-19
AC-5343: Build Environment Deployment Targets

shanbady:
to test this first create a new directory and then clone all three repos into it. 
```
mkdir modular-models-epic
cd modular-models-epic
git clone git@github.com:masschallenge/impact-api.git
git clone git@github.com:masschallenge/accelerate.git
git clone git@github.com:masschallenge/django-accelerator.git
```
then check out the AC-5343 branches on impact and accelerate. Also checkout the modular-models-epic branch on django-accelerator
```
cd accelerate && git checkout AC-5343
cd ../impact-api && git checkout AC-5343
cd ../django-accelerator && git checkout modular-models-epic
```

then run ```make build``` in the impact-api directory.

## testing the make release command

to test the release command:

1. cd into the impact-api project and create an empty commit via:
```git commit --allow-empty -m "test commit"```

    the release command will only create a release tag if there are changes since the last release. 

2. now run  ```make release``` to create a release in all 3 projects.

3. there should be a running travis build for the tagged release. once this completes - make note of the tag that was generated and move on to the next section.

## testing a deploy

1. to test a deploy, find a test opsworks stack and set the "custom cookbooks" branch to AC-5343:
screen shot 2018-03-19 at 3 07 03 pm

2. add the "impact_deploy::default" recipe to the "deploy" recipes section (OpsWorks -> [Stack] -> Layers -> Recipes) before the opsworks_deploy_python recipe:
screen shot 2018-03-19 at 4 30 30 pm

3. set the following settings in the stack's  custom json - you can borrow test values for this from the [test3 instance](https://console.aws.amazon.com/opsworks/home?region=us-east-1#/stack/d4a32a10-b655-44f1-9e2c-262885c56176/show):

    IMPACT_ENVIRONMENT=test
    DOCKER_REGISTRY=997386117669.dkr.ecr.us-east-1.amazonaws.com
    ECS_ACCESS_KEY_ID=(get it from test3)
    ECS_SECRET_ACCESS_KEY=(get it from test3)

4. set the mc app's deploy branch/tag to the release tag that was created in the last section.

5. in ECS make note of the currently deployed task definition's image tag. The "test" cluster is the one that we have specified via the IMPACT_ENVIRONMENT setting. make note of the currently deployed tag (ECS -> [Cluster] -> Services -> click link under "Task Defin...", scroll down on resulting page):
screen shot 2018-03-19 at 3 17 35 pm

https://console.aws.amazon.com/ecs/home?region=us-east-1#/clusters/test/services

6. deploy the opsworks stack and wait for it to finish.

7. go back to the "test" cluster in ECS and look at the currently running task definition. note that the deployed image tag should have changed.

### CODE REVIEW

File: Makefile [hunk 1]
```diff
@@ -367,8 +367,25 @@ upload-db:
...
 
 
-release-list release deploy:
-	@echo $@ not yet implemented
+release-list:
+	@git ls-remote --tags | grep -o 'refs/tags/v[0-9]*\.[0-9]*\.[0-9]*' | sort -r | grep -o '[^\/]*$$'
+
+
+install-releasedeps:
+	@if [ -d "$DIRECTORY" ]; then \
```
yotammanor: $DIRECTORY is not declared anywhere.

create_release.sh
```diff
@@ -0,0 +1,6 @@
+semantic-release --major version
```
yotammanor: not sure a major version should be the default option, I think that it should be minor by default (because anything beyond that might mean we should alert stakeholders etc).

Either way, this should be configurable.

Makefile
```diff
@@ -367,8 +367,25 @@ upload-db:
...
+		source .venv/bin/activate; \
+	else \
+		virtualenv --no-site-packages .venv && source .venv/bin/activate; \
+	fi;
+	@pip install python-semantic-release
```
yotammanor: I really don't like the fact that this requires working with virtualenv like this. Can't this happen inside a container?

At the very list, it means that the pre-requisites for this project should include installing virtualenv.

yotammanor (changes requested):
Cloning into a different directory - why is that required?

Haven't done functional review yet, but a few questions to start with.

Makefile
shanbady: fixed - wrapped the semantic-release module into a dockerfile. you will need to re-run ```make build``` to create the image before running ```make release```

Makefile
shanbady: good catch - no longer relevant since ive removed the target altogether

web/impact/impact/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = '0.1.32'
```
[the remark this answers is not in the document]
shanbady: yes - upon running make release after changes have been pushed

shanbady:
@yotammanor the "cloning into a different directory" was just to make it easier to get going since i didnt want to assume everyone had the same relative directory structure for these projects.

mc-nathan:
Functional issue:  Going through the instructions when I get to `make release` I get:
```
MC-HQ-NWILSON:impact-api nathan$ make release
Unable to find image 'semantic-release:latest' locally
docker: Error response from daemon: pull access denied for semantic-release, repository does not exist or may require 'docker login'.
See 'docker run --help'.

there is no new release to tag
```

mc-nathan:
Functional test passed.  There is a weird error about 'semantic-release:latest' at the beginning of `make build`, but it seems to sort itself out.  Here's the text of the error:
```
MC-HQ-NWILSON:impact-api nathan$ make build
Unable to find image 'semantic-release:latest' locally
docker: Error response from daemon: pull access denied for semantic-release, repository does not exist or may require 'docker login'.
...
```

yotammanor (changes requested):
Thanks for the changes introduced.
Checked functionally, working (mostly, see issue in accelerate PR), pretty cool.

A few things I noticed that are probably worth addressing:

When I run `make build` I get the following message:
```
MC-DEV-YMANOR:impact-api yotammanor$ make build
Unable to find image 'semantic-release:latest' locally
docker: Error response from daemon: pull access denied for semantic-release, repository does not exist or may require 'docker login'.
```
(Tried it after docker login, as well. same result)


- in the travis build, I see the following output (see [here](https://travis-ci.org/masschallenge/impact-api/builds/356360086#L1207)):
```
make travis-release AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY ENVIRONMENT=$DEPLOY_ENVIRONMENT DOCKER_REGISTRY=$DOCKER_REGISTRY IMAGE_TAG=$TAG
Unable to find image 'semantic-release:latest' locally
docker: Error response from daemon: pull access denied for semantic-release, repository does not exist or may require 'docker login'.
See 'docker run --help'.
tagging image v0.1.50
Incorrect Usage. flag provided but not defined: -access-key
NAME:
   ecs-cli configure - Stores a single cluster configuration.
USAGE:
   ecs-cli configure command [command options] [arguments...]
COMMANDS:
     profile  Stores a single profile.
     default  Sets the default cluster config.
     migrate  Migrates a legacy ECS CLI configuration file to the current YAML format.
OPTIONS:
   --cluster value, -c value            Specifies the ECS cluster name to use. If the cluster does not exist, it is created when you try to add resources to it with the ecs-cli up command. [$ECS_CLUSTER]
   --region value, -r value             Specifies the AWS region to use. If the AWS_REGION environment variable is set when ecs-cli configure is run, then the AWS region is set to the value of that environment variable. [$AWS_REGION]
   --config-name value                  Specifies the cluster configuration name to use for this configuration. (default: "default")
   --compose-service-name-prefix value  [Deprecated] Specifies the prefix added to an ECS service created from a compose file. Format <prefix><project-name>. (defaults to empty)
   --cfn-stack-name value               [Optional] Specifies the name of AWS CloudFormation stack created on ecs-cli up. (default: "amazon-ecs-cli-setup-<cluster-name>")
   --default-launch-type value          [Optional] Specifies the type of tasks that you would like to run. Options: EC2 or FARGATE. Defaults to empty string if none provided.
   --help, -h                           show help
...
... ## YM: omitting a few lines
...
v0.1.50: digest: sha256:44dec97ac8d555b31869d3e7b6b62ffbcbdf1d7f296b419b3d2835446117c49d size: 1571
time="2018-03-21T14:02:52Z" level=error msg="Unable to create an instance of CLIParams given the cli context" error="Set a region using ecs-cli configure command with the --region flag or AWS_REGION environment variable or --profile flag"
time="2018-03-21T14:02:52Z" level=fatal msg="Unable to create and read ECS Compose Project" error="Set a region using ecs-cli configure command with the --region flag or AWS_REGION environment variable or --profile flag"
make: *** [travis-release] Error 1
 ```

- As part of this PR, I'd love to see all OpsWorks stacks' settings updated to include the correct environment names and keys etc.

**Comment by mc-nathan** [comment]
Functional issue: Yesterday I did a successful release.  However, I did not notice that the file web/impact/impact/__init__.py had been modified in my local repository.  This morning it now conflicts with a fresh pull from the branch.  This seems like a potentially major issue.

When I was thinking about how to do this, it was exactly this sort of issue (file changes) that led me to want to do the release versioning as a git tag and to separate it from the rest of the semantic versioning (which is fine to store in __init__.py).  This is what motivated me to recommend that versions be <major>.<minor>.<patch>.<release>.  I now see that this is not technically "semantic versioning" as supported by python-semantic-versioning, but I do believe it is a reasonably common practice (e.g., Chrome => About Google Chrome).  I also note that formal semantic versioning does have an optional final spot for things like "-alpha" etc.  I think we could reasonably use that for the release number if so desired.  If we want to stick strictly with semantic versioning, then there we'll need to figure out a way to better manage the release process that can incorporate file changes in a way that does not require separate reviews etc.

**Comment by mc-nathan** [comment]
Starting from a new clean clone, I was not able to reproduce the problem with the unchecked in file.  Maybe this got fixed in some way.

File: __init__.py [hunk 5]
```diff
@@ -0,0 +1 @@
+__version__ = '0.1.34'
```
mc-nathan: What does this do?  Seems like web/impact/impact/__init__.py is the important one.

mc-nathan:
I still thing there's going to be an issue doing a release from development since pushes there aren't allowed without a review.  Is there some reason to thing that's not going to be the case?

mc-nathan (changes requested):
The code makes sense and seems reasonable, but I'm not convinced that it's actually going to work after it's merged to development.  Would probably be worth making another branch that has the same review restrictions that development has and see if it flies.  I also think there's a spare __init__ file that is not needed.

shanbady:
@mc-nathan the assumption is that the release command would be run from the master branch -post review merges so i dont think the review piece of the process will get in the way.

- Also once we get this in we will need to fiddle with permissions on the master branches. Worst case we may need to enable direct pushes to master or whatever release branch we choose but im hoping we can tie it down to only the semantic release user being able to create the release.

__init__.py
shanbady: good catch. removed

Makefile
```diff
@@ -403,6 +411,8 @@ endif
 
 
 # Deprecated targets
+deploy:
+	@echo $@ ERROR: see deployment steps for accelerate
```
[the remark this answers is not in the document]
shanbady: good point. i added a link to: https://github.com/masschallenge/standards/blob/376d290b41a202acc5c2263d7275ba4a57330ad7/create_new_release.md#deploy-to-staging for now

mc-nathan (approved):

Dockerfile.semantic-release
```diff
@@ -0,0 +1,13 @@
...
+RUN apk add --no-cache bash git
+RUN /entrypoint.sh \
+  -a git \
+  -p python-semantic-release \
+&& echo
+RUN apk add pcre git
+RUN git config --global user.email semantic-release@masschallenge.org
+RUN mkdir /app
+WORKDIR /app
+ADD setup.cfg .
```
yotammanor, line 12: would've preferred the `setup.cfg` would have a name that identifies it as config for semantic release, and not for any of the many other things that might require a config file around here :)

yotammanor (changes requested):
Changes look good on both impact-api and accelerate. I want to raise a few issues before I approve:

- Right now running `make release` triggers a git tag creation in both impact-api, accelerate and django-accelerator. It is the same tag literal, so all repos are kept in sync. @mc-nathan you said that you preferred them not to be kept in sync, so I want to make sure that we are aware of this, and that we are consciously choosing it to be this way for now.
- running `make release-list` prints out _all_ the versions, sorted alpha-numerically (v1.1, v1.2, v1.11, ...). seems like a possible follow-on, but want to make sure it is not missed.
- The default version bump is now implicitly PATCH, which is fine, but it's worth noting that the version bump is decided by parsing the latest(?) commit message according to a commit message style guide we don't follow. This might yield unexpected larger version bumps (probably not, though), and will make it hard for us to bump a MINOR or a MAJOR version if we want to. This reasonably sounds to me like a follow-on, but I want to make sure.
- I'd love to see all the version/release mess we've created as part of developing/testing this PR cleaned up.

another minor comment below.

mc-nathan:
To try to be clearer on the version issue, it is clear that we need to coordinate the release of impact-api, accelerate and django-accelerator.  This requires either that they all version in synchrony or one of them be the "master" or that we version the coordinated releases independently of any of the tools.  The current solution as I understand it leverages the third party library python-semantic-release to generate semantic versions that get stored in impact-api/web/impact/impact/__init__.py.  In my opinion this represent the middle option where impact-api is the "master".  The is also reflected in the fact that python-semantic-release is only part of the impact-api repo.

Assuming that we want to explicitly version accelerate and django-accelerator at some point (specifically such that they will have their own file with a version number in it), I do not see a reason why we would expect that version number to be the same as the version for impact-api.  The intent of semantic versioning as I understand it is to indicate changes to a specific tool that correlates with the level of semantic change in that code base.  If for example, we make a major change to impact-api (say switch to DRF 4.0 or whatever), but we don't expect there to be any change to django-accelerator, then in my view it would make sense to change the major version of impact-api, but not the major version of django-accelerator.

The application of a git tag, I see as being independent of the version of the individual tools that make up the release.  This is one of the reasons why originally I had proposed the tag be something like "release-1.2.3.4" where 1.2.3 is the current semantic version of impact-api.

Shankar has implemented this using a tag that is entirely based on the semantic version of impact-api (e.g., "v1.2.3").  This is workable as long as we assume that every release involves a change to the version of impact-api.  I'm OK with that, but it would not have been my choice.  The other problem that it creates as demonstrated by this conversation is that it make it less clear when a version reference is for a tool (impact-api) as opposed to a release (impact-api, accelerate and django-accelerator).  However, given that none of the tools currently have separate clients it doesn't practically make a difference.

mc-nathan:
Regarding the other concerns that @yotammanor raised, I agree with them all.  @shanbady please either address them in this PR or create follow on tickets.

Dockerfile.semantic-release
shanbady: unfortunately this is something that the library we use requires and is not changeable.

yotammanor (approved):

PR #115, opened 2018-03-22
AC-5238 Add Nginx/Supervisord-level Papertrail Logging to Impact API

pbx:
## [AC-5238](https://masschallenge.atlassian.net/browse/AC-5238)

You can test this locally by changing [start.sh line 16](https://github.com/masschallenge/impact-api/blob/AC-5238/web/scripts/start.sh#L16) to say `"Dev"` instead of `"Prod"` -- then logs will be sent from your dev machine.

The system name on Papertrail is [impact-api](https://papertrailapp.com/systems/impact-api/events) and I've created an [Impact API](https://papertrailapp.com/groups/8659131) group in Papertrail to hold it.

### DIFF
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -12,7 +12,7 @@ RUN apt-get update -y
 RUN apt-get install -y gettext mysql-client netcat nginx python-setuptools python2.7
 
 # Install remote_syslog2 to talk to papertrail
-RUN cd /tmp/ && curl -LO https://github.com/papertrail/remote_syslog2/releases/download/v0.20/remote-syslog2_0.20_amd64.deb
+RUN cd /tmp/ && curl -LOs https://github.com/papertrail/remote_syslog2/releases/download/v0.20/remote-syslog2_0.20_amd64.deb
 RUN dpkg -i /tmp/remote-syslog2_0.20_amd64.deb
 
 # Copy config files after installation so installers don't get tripped up
--- a/web/Dockerfile
+++ b/web/Dockerfile
@@ -3,31 +3,29 @@ FROM python:3.6
 WORKDIR /wwwroot
 
 COPY impact /wwwroot
-
 COPY scripts/start.sh /usr/bin
-
 COPY scripts/start-nodaemon.sh /usr/bin
 
 RUN apt-get update -y
 
 # python2.7 is required to run supervisor
-RUN apt-get install -y netcat mysql-client python2.7 python-setuptools nginx
+RUN apt-get install -y gettext netcat mysql-client nginx python2.7 python-setuptools
 
+# Install remote_syslog2 to talk to papertrail
+RUN cd /tmp/ && curl -LO https://github.com/papertrail/remote_syslog2/releases/download/v0.20/remote-syslog2_0.20_amd64.deb
+RUN dpkg -i /tmp/remote-syslog2_0.20_amd64.deb
+
+# Copy config files after installation so installers don't get tripped up
+COPY log_files.yml /etc/log_files.yml
 COPY nginx/nginx.conf /etc/nginx
 
 RUN pip install --upgrade pip
-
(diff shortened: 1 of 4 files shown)


web/Dockerfile
```diff
@@ -3,32 +3,30 @@ FROM python:3.6
...
-
 COPY scripts/start.sh /usr/bin
-
 COPY scripts/start-nodaemon.sh /usr/bin
 
 RUN apt-get update -y
 
 # python2.7 is required to run supervisor
-RUN apt-get install -y netcat mysql-client python2.7 python-setuptools nginx
+RUN apt-get install -y gettext netcat mysql-client nginx python2.7 python-setuptools
```
mc-nathan: Why the order change for `nginx`?  At first I thought it might be alphabetical, but then `netcat` is out of order.

mc-nathan (approved):
Looks reasonable.  One very minor non-blocking question.  Approved.

web/Dockerfile
pbx: Was intended to be alphabetical. Fixed.

yotammanor (dismissed):
- I was able to see supervisord logs in papertrail, but only after I moved the `if` statement in `start.sh` to run before `service nginx restart`. Not sure why.

- Seems like nginx.conf L28 and L29 set the logs to be to stdout instead of to the relevant log files. This seems to be wrong. It is also possible that nginx logs are not configured correctly in general, since they don't output to stdout locally. Possibly a different ticket, but preferably we should just sort it out now.

A small non-blocking comment below.

pbx:
Closing this until the remaining work (fixing the nginx log configuration issue that @yotammanor discovered) is completed.

web/Dockerfile
```diff
@@ -3,32 +3,30 @@ FROM python:3.6
...
 COPY scripts/start-nodaemon.sh /usr/bin
 
 RUN apt-get update -y
 
 # python2.7 is required to run supervisor
-RUN apt-get install -y netcat mysql-client python2.7 python-setuptools nginx
+RUN apt-get install -y gettext mysql-client netcat nginx python-setuptools python2.7
 
+# Install remote_syslog2 to talk to papertrail
+RUN cd /tmp/ && curl -LO https://github.com/papertrail/remote_syslog2/releases/download/v0.20/remote-syslog2_0.20_amd64.deb
```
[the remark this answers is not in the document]
pbx: Done

pbx:
Current state: Nginx log config has been fixed, and nginx log data is being sent to papertrail. However, a number of containers created by the build process also show up in papertrail in addition to **impact-api** (they are identified by their Docker container IDs, e.g. "cf0b168fedb9". I have put some work into trying to quiet these, but have not had success yet. Still, I wanted the PR back open so the requested work can be seen.

yotammanor:
@pbx I wasn't able to trigger an nginx log message (either from error.log or from access.log), although I did see that you got an `error` message. Any minty suggestions?

web/nginx/nginx.conf
```diff
@@ -25,8 +25,8 @@ http {
     ##
     # Logging Settings
     ##
-    access_log /dev/stdout;
-    error_log /dev/stdout info;
+    access_log /var/log/nginx/access.log;
+    error_log /var/log/nginx/access.log info;
```
yotammanor: this seems to be the wrong file name, assuming `/var/log/nginx/error.log` is not added by mistake.

pbx: :headdesk:

Fixed.

shanbady:
@yotammanor the reason you are not seeing output locally im guessing is because you are visiting http://localhost:8000 instead of https://localhost

locally - nginx is mapped to port 80 and django is mapped to 8000. if you visit http://localhost you should start seeing output in /var/log/nginx/

shanbady (approved):

PR #133, opened 2018-04-12
[AC-5457] - Deploy mechanism should automatically get the right version of django-accelerator

yotammanor:
**Changes Introduced in [AC-5457](https://masschallenge.atlassian.net/browse/AC-5457):**

- fix local dev environment, such that django-accelerator changes with the contents of the local django-accelerator directory.
- fix travis docker build such that django-accelerator version is inferred from the requested impact-api revision
- fix diff between local (Web) Dockerfile and Travis.Dockerfile.
- improve the Dockerfile structure to be more according to best practices (and utilize docker caching more efficiently, so building is usually faster)
- fix incorrect dependencies between services in docker-compose.yml
- Upgrade Docker-compose files from 2 to 2.3
- Update travis's docker compose version
- Cleanup some unused files.

**Test:**
- **Note: This might require you to upgrade your local docker engine. running `docker --version` should output `18.03.0-ce, build 0520e24`**

- Local django-accelerator behaviour:
  - on development:
    - `cd impact-api/;make delete-vms;make build;make run-server;make load-db;`
[NJW- This sequence failed for me with an error about `Service 'web' failed to build`.  When I switched to the branch, I was able to get `make build` to succeed.  However, the order of these commands seem wrong to me (and didn't work).  I would do `make load-db` before `make run-server` which seems to work correctly.]
    -  go to django-accelerator, create a new migration.
    - back in impact, run `make migrate`. See that "there's nothing to migrate".
    - delete that migration file
  - on AC-5457:
    - `cd impact-api/;make delete-vms;make build;make run-server;make load-db;`
[NJW- Again, I think this should be `make load-db; make run-server`.  It's also probably worth doing a `make clean-db-cache` before the `make load-db`.]
    -  go to django-accelerator, create a new migration.
    - back in impact, run `make migrate`. See that the new migration is executed.

- django-accelerator travis behavior:
  - in django-accelerator, on branch AC-5457, make a change that would cause an impact-api test to fail (e.g. raise error inside an __str__ function that is used)
  - trigger a build on impact-api's travis on branch AC-5457. See the expected test fail. Note that this happens without the django-accelerator branch being mentioned by its name.
  - stretch: 
    - set a travis env var DJANGO_ACCELERATOR_REVISION=development, and see that it overrides the inference mechanism. remove after testing.
    - create tags with identical names in impact and django-accelerator, and see the above behaviour works.
    - trigger an impact travis build on a branch/tag that does not exist in django-accelerator, see that it falls back to development.

- django-accelerator in deployed ECS behavior:
  - do as above, but instead of creating a breaking change, create a working noticeable change (e.g. change a field name).
  - after travis builds successfully, deploy the branch to a test machine. 
    - make sure that the opsworks_web_python's revision is set to AC-5457 (as in Test3)
    - make sure that there's an equivalent impact-api branch with that same name
  - See the change appear.

- See docker's cache improvement:
  - try to run `make build` again on the same branch. See that it works much faster than before.


mc-nathan (approved):
Code-wise this all looks very reasonable.  Still waiting for functional testing before merging.

mc-nathan:
Functional tests are good!

PR #140, opened 2018-05-03
[AC-5110] - Update accelerate to use SSO

yotammanor:
Changes Introduced in [AC-5110](https://masschallenge.atlassian.net/browse/AC-5110):

- Configure env vars to work with accelerate.
- Fix a few dev env issues

**Test:**
**IMPORTANT NOTE: If we're doing something wrong, we risk either breaking down our login process, or exposing our user data and credentials. Please test this very carefully.**

- See accelerate for functional testing.
- If you run `make delete-vms;make build;make run-server` on development, it will run forever and fail. On branch, it does not.


docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: '2.3'
 services:
   mysql:
-    image: mysql
+    image: mysql:5.7.21
```
pbx, line 4: Are we ready for 5.7? I'm not up to date on our plans on this.  I just know that Accelerate production is still at 5.5 and that [AC-4618](https://masschallenge.atlassian.net/browse/AC-4618) is still TODO.

pbx (changes requested):
My more detailed view is on the accelerate PR, but in short: **SSO on test3 didn't work for me**. I was able to log in via test3.masschallenge.org, but test3.api.masschallenge.org did not see me as authenticated.

docker-compose.yml
yotammanor: @pbx we are already on 5.7, run `make db-shell` on development and you'll see.
I believe that without this lock the mysql installed version is 8, and we're not ready for _that_ one.

yotammanor: @pbx we are not using mysql container on prod, this is only a development environment configuration at the moment. This PR doesn't change that (and again, we were already on 5.7 locally before this PR).

pbx (approved):

.dev.env
```diff
@@ -4,7 +4,8 @@ DJANGO_HIREDIS_CACHE_LOCATION=redis:6379
 DJANGO_ALLOWED_HOST=""
 DATABASE_URL=mysql://root:root@mysql/mc_dev
 EMAIL_URL=console://
-JWT_SECRET_KEY=ffffffff-ffff-ffff-ffff-ffffffffffffjwt
+JWT_AUTH_COOKIE=jwt
```
shanbady, line 7: can we make this the same name as what we use in accelerate? (JWT_COOKIE_NAME)

assets/gulpfile.js
```diff
@@ -53,7 +53,7 @@ gulp.task('build_js', function() {
       }
     }))
     .pipe(addsrc.prepend(['/gulp/js/dev/lib/*.js', '!/gulp/js/dev/lib/jquery.min.js']))
-    .pipe(addsrc.prepend('/gulp/js/dev/lib/jquery.min.js'))
+    .pipe(addsrc.prepend('/gulp/js/dev/lib/jquery.min.js', { allowEmpty: true }))
```
shanbady, line 56: why was the allowEmpty necessary?

shanbady (changes requested):

shanbady (changes requested):
only blocking comment is about the cookie name. will look tomorrow morning and get this merged.

shanbady (approved):

.dev.env
yotammanor: I was debating whether this should be the same as accelerate or the same as drf-jwt's name. We can pick that up later.

assets/gulpfile.js
yotammanor: The assets build step failed otherwise. try it and you'll see.

PR #142, opened 2018-05-18
[AC-5451] View to serve Mentor Directory

MattMSumner:
Why:

We would like to be able to deploy our shiny new Mentor Directory.

This PR:

* Adds a new page for serving the New-Mentor-Directory react app.
* Adds a new volume to `docker-compose.yml` for serving the built
  assets.

TODO:

* Currently for this to work you must have the `New-Mentor-Directory` as
  a sibling to the `impact-api` directory. You must also have run
  `bin/setup` and `yarn build` for the static assets to be generated.
  Ideally, this would be built for the first time via a docker
  container.
* Work out the story for production/staging. This is likely solved by
  building the app for the first time using docker.


MattMSumner:
This is definitely a work in progress but thought I'd throw this up here sooner rather then later.

web/impact/impact/urls.py
```diff
@@ -81,6 +82,8 @@
     url(r'^oauth/',
         include('oauth2_provider.urls', namespace='oauth2_provider')),
     url(r'^schema/$', schema_view, name='schema'),
+    url(r'^directory/$', TemplateView.as_view(template_name='directory.html'),
```
MattMSumner: I made this a dumb static view for now but I should likely change this to an authenticated view. I've not written python or Django before so went with the simplest solution to get started.

web/impact/templates/directory.html
```diff
@@ -0,0 +1,17 @@
+{% load staticfiles %}
```
MattMSumner: I'd actually rather not have this template and use the html file generated by the `New-Mentor-Directory` but couldn't work that out.

web/impact/impact/urls.py
shanbady: we should probably make the mentor directory a sub-url of directory unless we plan on serving the startup and mentor directory from a single directory.html template.

MattMSumner: We have a couple of options, we could redirect any path that starts with `/directory` and let react take over from there or we could use a subdomain. I'm not familiar enough with django but I'd presume `/directory*` will be simplier then a subdomain.

We can serve a single index page that boots the react app and that's fine.

web/impact/templates/directory.html
pbx: This template can contain whatever we want. Can you say more about what HTML you want here and what stopped you?

web/impact/impact/urls.py
pbx: Auth would be easy to add here: wrap the `TemplateView` with the `login_required` function -- [Django docs have an example](https://docs.djangoproject.com/en/1.10/topics/class-based-views/intro/#decorating-in-urlconf).

Seems like we need a design decision on the URL structure though.

docker-compose.yml
```diff
@@ -42,6 +42,7 @@ services:
     volumes:
       - ./web/impact:/wwwroot
       - ../django-accelerator:/packages/src/django-accelerator
+      - ../New-Mentor-Directory/dist:/wwwroot/static/dist
```
pbx: Since we're creating this special mount point just for the directory static files, we can give it a more expressive name, e.g. `/wwwroot/static/mentor-directory`.  Just a thought, not a blocker.

pbx (changes requested):
The work in progress looks good to me. Made a few comments. In the instructions, where you say `yarn build` did you mean `yarn start` or something else?

web/impact/templates/directory.html
shanbady: yea - the way i had this in my head is that we would merge the contents of the new mentor directory app into impact and subsequently get rid of the New Mentor Directory repo (currently this depends on having the repos side by side but im not sure how this plays into our deploys (how does the static resource get generated by ECS deployments.)

shanbady: I see - ok so the built assets are packaged into the image as a volume. i think thats sufficient for now.

docker-compose.yml
shanbady: There is a piece of this setup that is missing. Our current build/release process looks like the following:

- travis kicks off a build of the docker-compose.travis.yml file
- built images are pushed to ECR
- a deploy of accelerate kicks off a deploy of the built image to the ECS cluster

since there is no reference to the new mentor directory repo in travis yml file, the pushed impact api image would have none of the compiled assets for it. I just recently created a PR that adds a Dockerfile to the new mentor directory so that we can reference it in the docker-compose file:
https://github.com/masschallenge/New-Mentor-Directory/pull/18

shanbady: this is what the entry in the docker-compose.yml and docker-compose.travis.yml should look like once the aforementioned PR is merged:

```
  mentor_directory:
    build: ../New-Mentor-Directory
    volumes:
      - ../New-Mentor-Directory/dist:/usr/src/app/dist
```

the "web" docker-compose service should also contain a "depends_on" entry for "mentor_directory" so that the resources get built before mounting into the web service.

shanbady (changes requested):

shanbady (approved):

pbx (approved):

PR #144, opened 2018-05-24
Travis fix (AC-5652)

shanbady:
this fixes the easy_install hangup in travis by manually installing it via wget | python


pbx (approved):
Ugly that we have to do this, but it solves the problem.

PR #151, opened 2018-06-07
Add `.env` file to local dev docker-compose

MattMSumner:
Why:

We would like to add secrets to the environment but not accidentally
commit them. Currently, we have to edit `.dev.env` which is not git
ignored to have secret env variables and it might be easy to commit
these by mistake.

This PR:

Adds `.env` as an `env_file` to the local dev docker-compose file.
`.env` is already git ignored so developers can now put secrets in a
local `.env` file and be safe from accidentally committing them.


yotammanor:
@MattMSumner travis does not agree with you

MattMSumner:
@yotammanor fixed! Thanks.

shanbady (approved):

PR #150, opened 2018-06-04
[AC-5679] Add and configure CORS support

mc-nathan:
To test:
- On the branch run `make build; make run-server`
- Get a local version of accelerate running as well (`development` branch should be fine).  Login to your local version of accelerate.
- Git clone application-allocator and checkout branch AC-5679.
- Run `make run-server`
- It should report `http://localhost:1234`.  Go to that URL.
- You should see "Hello, World!" and the first name a last name of user 182.  The user info is coming from an API call.
- Log out of accelerate and refresh the application-allocator page.
- The name should disappear and the JavaScript console should report a 401.
- Log back into accelerate, do another refresh and see the name come back.

This can be merged before the corresponding PR on application-allocator, [AC-5679](https://github.com/masschallenge/application-allocator/pull/3).

PR #153, opened 2018-06-08
[AC-5689] Add JudgingRound API calls

mc-nathan:
To test these API calls should return reasonable stuff:
- http://localhost:8000/api/v1/judging_round/
- http://localhost:8000/api/v1/judging_round/?is_active=True
- http://localhost:8000/api/v1/judging_round/48/
And now with round_type!
- http://localhost:8000/api/v1/judging_round/?round_type=Online


yotammanor (changes requested):
Functional concern - is_active filter does not accept lowercase "true" or "false", and fails with a 500 for any invalid value (e.g. "foo"). It does accept 1 and 0, which is nice, I guess.

I'm also surprised that the created_at and updated_at fields do not appear, I thought it's standard to add them - plus, they are filterable for this view.

mc-nathan:
@yotammanor Did more error checking on is_active.  Also added round_type option since I'm probably going to actually need that more than is_active.

yotammanor (approved):
Looks great, thanks for picking up the changes.

Approved. Not merging because this PR is on top of #150, and this one is not approved yet.

PR #155, opened 2018-06-13
[AC-5718] - BUG - Accessing the Directory requires the user to login twice

yotammanor:
**Changes Introduced in [AC-5718](https://masschallenge.atlassian.net/browse/AC-5718):**

- Remove the login_required decorator from impact-api's directory urls

**Why:**
- The directory has it its own mechanism to check for authentication. Therefore those views should be **login exempt**, not login required.

**Test:**
- bring the servers up.
- Login on accelerate, and go to the directory. See that you are recognized as logged in, and all is working.
- Logout on accelerate (or through the logout button on the directory). Go to the directory, see that you are redirected to login in accelerate (not in the browsable api login page on impact).

**Note:**
- The value of `?next=` is within localhost:8181, which is wrong, but @shanbady says that it's addressed by making the directory appear as served from the accelerate url, so this is not handled here.

### DIFF
--- a/web/impact/impact/settings.py
+++ b/web/impact/impact/settings.py
@@ -151,6 +151,7 @@ class Base(Configuration):
     CORS_ALLOW_CREDENTIALS = True
     CORS_ORIGIN_WHITELIST = (
         'localhost:1234',
+        'localhost:8000',
         )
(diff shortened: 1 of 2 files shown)


MattMSumner (approved):
This LGTM. I've opened https://github.com/masschallenge/directory/pull/48 to address the `next` query param being set to `/accounts/logout/`.

I ran through the acceptance criteria locally and this works as expected.

PR #159, opened 2018-06-15
[AC-5729] Add .env target and connect with appropriate targets

mc-nathan:
To test:
- Move any existing .env file aside (e.g., `mv .env tmp.env`)
- On development running `make test` (or `make run-server` or `make migrate`) will fail with an error
- On branch it will just work.
- Note you will need to remove the empty .env file each time you want to test a different one of these targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -264,7 +264,7 @@ checkout:
 # Server and Virtual Machine related targets
 debug ?= 1
 
-run-server: run-server-$(debug)
+run-server: .env run-server-$(debug)
 
 run-server-0: initial-db-setup
 	@docker-compose up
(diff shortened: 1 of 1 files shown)


yotammanor (changes requested):
The following commands don't work without .env file and probably should:

- build (and build-all)
- coverage
- run-detached-server
- run-server-0
- run-all-servers
- stop-server (and stop-all-servers)
- shutdown-vms
-bash-shell, django-shell, db-shell, load-db, dumpo-db (they sort of have an implicit dependency on a working server, which is consistent if .env was never there, but inconsistent if the .env file is removed after `run-server` got executed successfully).

Changes introduced are working as expected.

mc-nathan:
Got more targets to check for .env.  Specifically, all targets that call docker-compose that don't have another dependency that checks for .env.  All the cases described by @yotammanor now work.

yotammanor (approved):
Thanks!

PR #161, opened 2018-06-15
AC-5646: allow impact api to serve the directory as a proxied view from accelerate

shanbady:
a working setup of this is on test4.masschallenge.org


pbx:
Need a little more mint

yotammanor (approved):
MAW.

PR #165, opened 2018-06-20
[AC-5535] Add graphql endpoint

MattMSumner:
Why:

We would like to try graphql as a potential format for providing data to
a frontend. There is a lot of support for this specification throughout
the react community and has been gaining in popularity for servers that
are expected to serve multiple frontends. Advantages include:

* Pushing querying capabilities to the client. This allows for clients
  to move faster when wanting to make changes around where and when data
  is displayed.
* Allowing clients to ask for exactly the data they require, sending
  less across the wire.
* Since the graphql specification is so robust, it's easy to generate a
  schema from the server that can be used to sanity check the queries a
  client is making. This effectively allows the client to "type check"
  the data they're planning to request.

This PR:

* Introduces the library `graphene-django` to expose a grpahql endpoint.
* Adds a top level `query` as the entry point into the graph.
* Adds an `expert_profile` that requires an id to fetch a particular
  expert.
* Adds the `ExpertType` and related types to expose the data required to
  be able to display an expert details page.

Testing:

* Checkout locally.
* Rebuild to install new dependencies.
* Go to `http://localhost:8000/graphql/`
* Checkout the documentation explorer on the right!
* Try out some queries and make sure you get back reasonable data!

screen shot 2018-06-20 at 3 31 50 pm

### DIFF
--- a/web/Dockerfile.travis
+++ b/web/Dockerfile.travis
@@ -29,6 +29,7 @@ RUN wget --no-check-certificate https://bootstrap.pypa.io/ez_setup.py -O - | pyt
 RUN easy_install-2.7 supervisor
 RUN useradd -s /bin/bash -u 3000 -m impact_user
 RUN pip3 install -r /wwwroot/requirements/travis.txt
+RUN pip3 install --no-dependencies -r /wwwroot/requirements/no_dependencies.txt
 
 ARG DJANGO_ACCELERATOR_REVISION=development
 RUN pip3 install git+https://github.com/masschallenge/django-accelerator.git@${DJANGO_ACCELERATOR_REVISION}#egg=django-accelerator
--- a/web/impact/impact/graphql/query.py
+++ b/web/impact/impact/graphql/query.py
@@ -1,0 +1,16 @@
+import graphene
+
+from impact.graphql.types.expert_profile_type import ExpertProfileType
+from accelerator.models.expert_profile import ExpertProfile
+
+
+class Query(graphene.ObjectType):
+    expert_profile = graphene.Field(ExpertProfileType, id=graphene.Int())
+
+    def resolve_expert_profile(self, info, **kwargs):
+        expert_profile_id = kwargs.get('id')
+
+        if expert_profile_id is not None:
+            return ExpertProfile.objects.get(pk=expert_profile_id)
+
+        return None
--- a/web/impact/impact/settings.py
+++ b/web/impact/impact/settings.py
@@ -100,7 +100,9 @@ class Base(Configuration):
         'rest_framework_swagger',
         'fluent_pages',
         'impact',
+        'graphene_django',
     ]
+
     ACCELERATOR_MODELS_ARE_MANAGED = True
 
     AUTH_USER_MODEL = 'simpleuser.User'
@@ -112,6 +114,7 @@ class Base(Configuration):
         'corsheaders.middleware.CorsMiddleware',
         'django.contrib.auth.middleware.AuthenticationMiddleware',
         'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
+        'impact.graphql.auth.middleware.CookieJSONWebTokenMiddleware',
         'django.contrib.messages.middleware.MessageMiddleware',
         'django.middleware.clickjacking.XFrameOptionsMiddleware',
         'django.middleware.security.SecurityMiddleware',
@@ -153,7 +156,8 @@ class Base(Configuration):
         'localhost:1234',
         'localhost:8000',
     )
-    CORS_ORIGIN_REGEX_WHITELIST = (r'^(https?://)?(\w+\.)?masschallenge\.org$', )
+    CORS_ORIGIN_REGEX_WHITELIST = (
+        r'^(https?://)?(\w+\.)?masschallenge\.org$', )
     ALGOLIA_INDEX_PREFIX = os.environ.get('ALGOLIA_INDEX_PREFIX', 'dev')
     ALGOLIA_INDEXES = [
         '{algolia_prefix}_mentor'.format(algolia_prefix=ALGOLIA_INDEX_PREFIX)
@@ -231,6 +235,7 @@ class Base(Configuration):
 
     AUTHENTICATION_BACKENDS = (
         'oauth2_provider.backends.OAuth2Backend',
+        'impact.graphql.auth.authentication_backend.JWTokenCookieBackend',
         'simpleuser.email_model_backend.EmailModelBackend',
         'django.contrib.auth.backends.ModelBackend',
     )
--- a/web/impact/impact/tests/test_graphql.py
+++ b/web/impact/impact/tests/test_graphql.py
@@ -1,0 +1,62 @@
+# MIT License
+# Copyright (c) 2017 MassChallenge, Inc.
+
+from django.urls import reverse
+
+from impact.graphql.middleware import NOT_LOGGED_IN_MSG
+from impact.tests.api_test_case import APITestCase
+from impact.tests.factories import ExpertFactory
+from impact.tests.utils import capture_stderr
+
+
+class TestGraphQL(APITestCase):
+    url = reverse('graphql')
+    auth_url = reverse('graphql-auth')
+
+    def test_anonymous_user_cannot_access_main_graphql_view(self):
+        user = ExpertFactory()
+        query = """query {{ expertProfile(id: {id}) {{ title }} }}
+            """.format(id=user.expertprofile.id)
+
+        with capture_stderr(self.client.post,
+                            self.url,
+                            data={'query': query}) as (response, _):
+            error_messages = [x['message'] for x in response.json()['errors']]
+
+        self.assertIn(NOT_LOGGED_IN_MSG, error_messages)
+
+    def test_anonymous_user_can_access_auth_graphql_view(self):
+        query = """mutation {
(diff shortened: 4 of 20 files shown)


MattMSumner:
There are two outstanding issues before this can be merged:

1) ~I don't know how to protect this endpoint using the JWT mechanism, @yotammanor maybe you would be able to help with that? Basically you shouldn't be able to post to `/graphql/` and receive a 200 without a valid JWT. I think I've written a failing test for this.~
 - This has been resolved by @yotammanor. Thanks!

2) In urls.py I have the following:
```
url(r'^graphql/$', csrf_exempt(GraphQLView.as_view(graphiql=True)),
        name="graphql"),
```
~The `graphiql=True` is what loads up the GUI interface for if you visit via the browser. That should only be True in development. I'm sure this is easy but I'm not sure how to do this.~
 - This has now been resolved.

web/impact/impact/graphql/expert_profile.py
```diff
@@ -0,0 +1,29 @@
+import graphene
+from graphene_django import DjangoObjectType
+from accelerator.models.expert_profile import ExpertProfile
+from impact.graphql.user import UserType
```
johnschoeman: in general should we be naming type files with the suffix `_type.py`? i.e. `from impact.graphql.user_type import UserType` (mostly for convenience w/ ctrl-p)

web/impact/impact/settings.py
```diff
@@ -153,7 +159,7 @@ class Base(Configuration):
         'localhost:1234',
         'localhost:8000',
     )
-    CORS_ORIGIN_REGEX_WHITELIST = (r'^(https?://)?(\w+\.)?masschallenge\.org$', )
+    CORS_ORIGIN_REGEX_WHITELIST = (r'^(https?://)?(\w+\.)', )
```
johnschoeman: It's not clear to me why this whitelist needs to be less restrictive?

johnschoeman (approved):
Looks good to me! (though, I'm not super familiar with python/django) A few questions on things

yotammanor:
@MattMSumner I'll look into it.

web/impact/impact/graphql/query.py
```diff
@@ -0,0 +1,16 @@
...
+
+from impact.graphql.expert_profile import ExpertProfileType
+from accelerator.models.expert_profile import ExpertProfile
+
+
+class Query(graphene.ObjectType):
+    expert_profile = graphene.Field(ExpertProfileType, id=graphene.Int())
+
+    def resolve_expert_profile(self, info, **kwargs):
+        id = kwargs.get('id')
```
yotammanor: `id` is a built-in python function, please rename

web/impact/impact/tests/test_graphql.py
```diff
@@ -0,0 +1,83 @@
...
+                        additionalIndustries {{
+                            name
+                        }}
+                        bio
+                        homeProgramFamily {{
+                            name
+                        }}
+                    }}
+                }}
+            """.format(id=user.expertprofile.id)
```
yotammanor: this string looks awful.. is there a way to prettify it?

web/impact/impact/tests/test_graphql.py
```diff
@@ -0,0 +1,83 @@
...
+            response = self.client.post(self.url, data={'query': query})
+            self.maxDiff = None
```
yotammanor: seems redundant

web/impact/impact/urls.py
```diff
@@ -82,7 +84,8 @@
     url(r'^sso/', include(sso_urlpatterns, namespace="sso")),
     url(r'^admin/', include(admin.site.urls)),
     url(r'^accounts/', include(account_urlpatterns)),
-
+    url(r'^graphql/$', csrf_exempt(GraphQLView.as_view(graphiql=True)),
```
yotammanor: ~`graphiql=True`~
`graphiql=settings.DEBUG`

yotammanor (dismissed):
Looking into the JWT thing.
A few other comments below + the graphiql in dev only.

Also, question - why are we POSTing to graphql instead of GETing?

johnschoeman:
@yotammanor : I can answer the POST question: All requests to graphql endpoints are over POST, per the specification.

web/impact/impact/tests/test_graphql.py
johnschoeman: In general graphql queries look a bit gnarly at first, they are expressing a lot of information. To my eye, this looks in line with what I would expect from a gql query

web/impact/impact/graphql/query.py
```diff
@@ -0,0 +1,16 @@
+import graphene
+
+from impact.graphql.types.expert_profile_type import ExpertProfileType
+from accelerator.models.expert_profile import ExpertProfile
+
+
+class Query(graphene.ObjectType):
+    expert_profile = graphene.Field(ExpertProfileType, id=graphene.Int())
```
MattMSumner, line 8: The id should appear as required.

web/impact/impact/tests/test_graphql.py
```diff
@@ -0,0 +1,83 @@
...
+            self.assertJSONEqual(
```
jonkiparsky: Testing every field seems excessive. Probably would be better to test that we can query an ExpertProfile and get back just the one or two fields that would confirm we were getting the right profile back.

web/impact/impact/graphql/expert_profile.py
jonkiparsky: 👍 We like file names that reflect class names.

web/impact/impact/graphql/expert_profile.py
```diff
@@ -0,0 +1,29 @@
...
+from impact.graphql.program_family import ProgramFamilyType
+from impact.graphql.startup import StartupType
+
+
+class ExpertProfileType(DjangoObjectType):
+    mentees = graphene.List(StartupType)
+
+    class Meta:
+        model = ExpertProfile
+        only_fields = (
```
jonkiparsky: Is the use of whitelist vs blacklist an intentional decision about the semantics of the two approaches? 
I note that exclude_fields would be a much shorter list.

web/impact/impact/graphql/query.py
jonkiparsky: @MattMSumner `expert_profile_id` would be fine

jonkiparsky (changes requested):
Looks good, a few small changes.

web/impact/impact/graphql/expert_profile.py
MattMSumner: I generally prefer whitelists due to the default behavior of adding a new attribute. Forgetting to whitelist an attribute means it doesn't appear in our API. Forgetting to blacklist a sensitive attribute could lead to unforeseen consequences.

web/impact/impact/tests/test_graphql.py
MattMSumner: @jonkiparsky addressed in https://github.com/masschallenge/impact-api/pull/165/commits/d65fb539a3551a1cb5a753ac6d2ffe754d034c92

web/Dockerfile.travis
```diff
@@ -29,6 +29,7 @@ RUN wget --no-check-certificate https://bootstrap.pypa.io/ez_setup.py -O - | pyt
 RUN easy_install-2.7 supervisor
 RUN useradd -s /bin/bash -u 3000 -m impact_user
 RUN pip3 install -r /wwwroot/requirements/travis.txt
+RUN pip3 install --no-dependencies -r /wwwroot/requirements/no_dependencies.txt
```
pbx, line 32: I saw the `no_dependencies.txt` file before I saw this line, and had thought maybe the file was for packages which did not have any dependencies.  Since a requirements file paired with a special pip invocation is an unusual arrangement for requirements in our Python code, I'd like to see a comment at the top of the `no_dependencies.txt` file clarifying that the packages listed will be installed with pip's `--no-dependencies` flag.

(I'm also curious why we need to install `django-graphql-jwt` in this way, but that's just curiosity)

pbx (approved):
This works great and looks fine, my only reservation is about clarity re the `no_requirements.txt` file.  I won't object if this is merged without that change but I did want to raise it. Leaving unmerged until author(s) make a call.

web/Dockerfile.travis
yotammanor: @pbx the reason is that django-graphql-jwt is supported only for django 1.11+. I did not run into any actual issue with using it in django 1.10, except for the fact that it tries to install a newer django version than we'd like. 

I'll add the comment you requested.

web/impact/impact/tests/test_graphql.py
```diff
@@ -0,0 +1,62 @@
...
+                    }
+                }"""
+        response = self.client.post(self.url, data={'query': query})
+        error_messages = [x['message'] for x in response.json()['errors']]
+        self.assertNotIn(NOT_LOGGED_IN_MSG, error_messages)
+
+    def test_query_with_expert(self):
+        with self.login(email=self.basic_user().email):
+            user = ExpertFactory()
+            query = """
```
jonkiparsky, line 41: Not a blocker, but I still don't love the pattern we're setting here. At minimum, I'd like the literal strings to be externalized. Better would be for the query and the expected result to be generated by some function, but that might be too much to ask for, considering the time frame we're in. :)

jonkiparsky (approved):

PR #169, opened 2018-06-28
[Ac 5766] Add filters for criterion and criterion_option_spec api calls

jonkiparsky:
To test: 
modify some criterion to point to a round other than 48. Append ?judging_round_id=48 to a criterion or criterion_option_spec list call and see that the appropriate objects come back
option specs can also be filtered by criterion id.

### DIFF
--- a/web/impact/impact/v1/views/criterion_list_view.py
+++ b/web/impact/impact/v1/views/criterion_list_view.py
@@ -4,7 +4,7 @@
 from impact.v1.views.base_list_view import BaseListView
 from impact.v1.views.post_mixin import PostMixin
 from impact.v1.helpers import CriterionHelper
-from impact.v1.helpers.validators import INVALID_INTEGER_ERROR
+
 
 class CriterionListView(BaseListView,
                         PostMixin):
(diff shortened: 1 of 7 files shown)


web/impact/impact/v1/views/criterion_list_view.py
```diff
@@ -11,3 +11,13 @@ class CriterionListView(BaseListView,
     view_name = "criterion"
     helper_class = CriterionHelper
     actions = ['GET', 'POST']  # Should get this from PostMixin
+
+    def get(self, request):
+        self.validate_id(request, 'judging_round_id')
+        return super().get(request)
+
+    def filter(self, qs):
+        return self._filter_by_judging_round_id(qs)
```
mc-nathan: This should be calling super() to pick up the other standard filters.  Something like:
`return self._filter_by_judging_round_id(super().filter(qs))`

web/impact/impact/v1/views/criterion_option_spec_list_view.py
```diff
@@ -11,3 +11,19 @@ class CriterionOptionSpecListView(BaseListView,
...
     view_name = "criterion_option_spec"
     actions = ['GET', 'POST']  # Should get this from PostMixin
+
+    def get(self, request):
+        self.validate_id(request, 'judging_round_id')
+        self.validate_id(request, 'criterion_id')
+        return super().get(request)
+
+    def filter(self, qs):
+        qs = self._filter_by_criterion_id(qs)
```
mc-nathan: Again this should be calling super().

mc-nathan (changes requested):
Couple missing calls to super, but otherwise looks good.

mc-nathan (approved):

PR #170, opened 2018-06-29
[AC-5666] - Expert profile page (GraphQL Server fix)

yotammanor:
**Changes Introduced:**

- Expose image_url instead of image field's default representation (which is the image's name)
- add schema.json generation in start.sh (as part of docker's build)
- add schema.json to gitignore.

### DIFF
--- a/web/scripts/start.sh
+++ b/web/scripts/start.sh
@@ -15,7 +15,7 @@ fi
 
 python3 manage.py migrate --noinput
 python3 manage.py collectstatic --noinput
-python3 manage.py graphql_schema --schema impact.graphql.schema.schema
+python3 manage.py graphql_schema --schema impact.graphql.schema.schema --indent=2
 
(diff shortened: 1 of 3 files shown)


web/scripts/start.sh
```diff
@@ -15,6 +15,7 @@ fi
 
 python3 manage.py migrate --noinput
 python3 manage.py collectstatic --noinput
+python3 manage.py graphql_schema --schema impact.graphql.schema.schema
```
MattMSumner: I'm not sure what this line is doing. Could you clarify what this line does?

MattMSumner (approved):
One question for my own interest but otherwise this looks great!

web/scripts/start.sh
yotammanor: @MattMSumner same as `gql:schema` but server side.

PR #171, opened 2018-06-29
AC-4121 Provide 500x500 thumbnails through the API

pbx:
## [AC-4121](https://masschallenge.atlassian.net/browse/AC-4121)

This feature relies heavily on a Cloudfront-based solution (detailed in the ticket comments) which includes the following:
* an endpoint through which all `startup_pics` image requests are made
* an AWS lambda function that's called by that endpoint
* an image resizing utility ("Thumbor") called by that Lambda function as needed
* a cloudfront cache where the resized images are stored

This solution is currently set up to load and process images from the `media/startup_pics` directory in the [accelerate-assets-staging](https://s3.console.aws.amazon.com/s3/buckets/accelerate-assets-staging/?region=us-east-1&tab=overview) S3 bucket.

The code changes make the `api/v0/startup_detail` endpoint change the `logo_url` values it returns, to use the resizer.

For production use we'll either update the resizer to watch the {{accelerate-assets-production}} bucket, or (if we feel that there's some use in keeping it available for staging) add a second processor that uses that production bucket.

**TO TEST**:
* Make a Postman request to the `/api/v0/startup_detail/` endppoint for a startup with a logo
* See that the `logo_url` returned uses the Cloudfront host `dl4fx6jt7wkin.cloudfront.net`

### DIFF
--- a/web/impact/impact/settings.py
+++ b/web/impact/impact/settings.py
@@ -270,6 +270,9 @@ class Base(Configuration):
 
     CMS_FILE_ROOT = '/var/www/cms-files'
 
+    IMAGE_RESIZE_HOST = "https://dl4fx6jt7wkin.cloudfront.net"
+    IMAGE_RESIZE_TEMPLATE = "/fit-in/500x500/{}"
+
 
 class Dev(Base):
     DEBUG = True
--- a/web/impact/impact/settings.py
+++ b/web/impact/impact/settings.py
@@ -271,7 +271,7 @@ class Base(Configuration):
     CMS_FILE_ROOT = '/var/www/cms-files'
 
     IMAGE_RESIZE_HOST = "https://dl4fx6jt7wkin.cloudfront.net"
-    IMAGE_RESIZE_TEMPLATE = "/fit-in/500x500/{}"
+    IMAGE_RESIZE_TEMPLATE = "/fit-in/500x500/media/startup_pics/{}"
 
 
 class Dev(Base):
--- a/web/impact/impact/settings.py
+++ b/web/impact/impact/settings.py
@@ -271,7 +271,7 @@ class Base(Configuration):
     CMS_FILE_ROOT = '/var/www/cms-files'
 
     IMAGE_RESIZE_HOST = "https://dl4fx6jt7wkin.cloudfront.net"
-    IMAGE_RESIZE_TEMPLATE = "/fit-in/500x500/media/startup_pics/{}"
+    IMAGE_RESIZE_TEMPLATE = "/fit-in/500x500/media/{}"
 
 
 class Dev(Base):
--- a/web/impact/impact/v0/views/utils.py
+++ b/web/impact/impact/v0/views/utils.py
@@ -55,8 +55,7 @@ def logo_url(startup):
     elif "startup_pics/" not in logo_field.url:
         return logo_field.url
     else:
-        # Get just the filename and drop any S3 auth params
-        filename = logo_field.url.split("startup_pics/")[1].split("?")[0]
+        filename = logo_field.name
         template = settings.IMAGE_RESIZE_HOST + settings.IMAGE_RESIZE_TEMPLATE
         return template.format(filename)
 
--- a/web/impact/impact/v0/views/utils.py
+++ b/web/impact/impact/v0/views/utils.py
@@ -45,9 +45,20 @@ def _pad(text):
 
 
 def logo_url(startup):
-    if not startup.high_resolution_logo:
+    """
+    Turns the stored *.s3.amazonaws.com URL into one that uses our
+    CloudFormation image resizer; returns empty string for None
+    """
+    logo_field = startup.high_resolution_logo
+    if not logo_field:
         return ""
-    return startup.high_resolution_logo.url
+    elif "startup_pics/" not in logo_field.url:
+        return logo_field.url
+    else:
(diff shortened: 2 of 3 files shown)


web/impact/impact/settings.py
```diff
@@ -270,6 +270,9 @@ class Base(Configuration):
 
     CMS_FILE_ROOT = '/var/www/cms-files'
 
+    IMAGE_RESIZE_HOST = "https://dl4fx6jt7wkin.cloudfront.net"
```
yotammanor, line 273: should probably be an env var

yotammanor: And I'd consider adding a Route 53 DNS record (within the cloudformation stack?) and give this link a readable name that makes sense (e.g. has staging/prod in it).

yotammanor (changes requested):
- Pretty cool approach to this problem.
- Code looks mostly good, a few small suggestions below.
- Not sure how to review any of what goes on in the lambda. I did give it a look and it seems like a mostly shelf-solution so worth noting what in particular did we configure on our own and how we configure it in the future (e.g. by adding documentation in standards)

- Also, I think we definitely have to have it configured both in staging and production, so in case we break it somehow - we know it before it hits production.

- Functionally, I saw the formatted url correctly. Tried to access one of them, https://dl4fx6jt7wkin.cloudfront.net/fit-in/500x500/logo_Ua7EIpP.png , got an internal server error. _Might_ be a local configuration error, but I'd love to see it work before we approve.
- I also noticed that this internal server error did not generate an error email with logs, which seems crucial to me if we want to support this.

web/impact/impact/v0/views/utils.py
```diff
@@ -45,9 +45,20 @@ def _pad(text):
...
+    """
+    logo_field = startup.high_resolution_logo
+    if not logo_field:
         return ""
-    return startup.high_resolution_logo.url
+    elif "startup_pics/" not in logo_field.url:
+        return logo_field.url
+    else:
+        # Get just the filename and drop any S3 auth params
+        filename = logo_field.url.split("startup_pics/")[1].split("?")[0]
```
yotammanor: much better to use logo_field.name instead of logo_field.url. It will save you the s3 params.

Also, the `startup_pics` part is required to make the path unique (within the MEDIA_ROOT). Not sure that splitting it out is a smart call, and I'm not sure where to look at to evaluate the other end of this decision.

yotammanor (dismissed):
With an extra comment that did not get submitted apparently. :)

web/impact/impact/v0/views/utils.py
```diff
@@ -45,9 +45,20 @@ def _pad(text):
...
-    if not startup.high_resolution_logo:
+    """
+    Turns the stored *.s3.amazonaws.com URL into one that uses our
+    CloudFormation image resizer; returns empty string for None
+    """
+    logo_field = startup.high_resolution_logo
+    if not logo_field:
         return ""
-    return startup.high_resolution_logo.url
+    elif "startup_pics/" not in logo_field.url:
```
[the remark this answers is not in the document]
pbx: This was a safety in case (for some reason) the URL pointed somewhere outside the bucket that the solution is configured to work with. But now I'm inclined to use your `logo_url.name` idea below, which would obviate this check.

web/impact/impact/settings.py
pbx: That's a good idea.

web/impact/impact/v0/views/utils.py
pbx: Thanks for the `logo_field.name` tip, that's perfect. I'll think about the `startup_pics` issue -- right now, this solution is handling those images and no others.

pbx:
Re the "internal server error did not generate an error email with logs" issue -- since this is totally detached from our Django stack, monitoring like that would need to be set up separately. Logs _are_ accessible via the [CloudWatch console](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logStream:group=/aws/lambda/ServerlessImageHandler-ImageHandlerFunction-N89PYZCFHZRA;streamFilter=typeLogStreamPrefix). If we want more, I'm sure AWS has a way to provide it, but I'd suggest that be a follow-on ticket rather than rolled into this.

yotammanor:
@pbx as long as we have visibility somewhere, a follow on ticket to connect cloudwatch to papertrail, or slack or email notifications sounds more than enough.

pbx:
I've made some progress on the "readable domain name that makes sense" feature (e.g. **image-resizer-staging.masschallenge.org**), but in the interest of shipping this before I leave for vacation I may defer that to a follow-on ticket as well.

web/impact/impact/settings.py
pbx: After some effort, created a ticket for this.  https://masschallenge.atlassian.net/browse/AC-5794

jonkiparsky (approved):
Looks reasonable to me. I agree with Yotam's concern, but the follow-on ticket is acceptable as a remedy. 
My only concern is that this leaves us with an undocumented dependency, and I would love to see the some words added to the standards repo to help future team understand how our resizing works and how to change it if we ever need to.

pbx:
Great, and agreed on the docs.  I cut [AC-5796](https://masschallenge.atlassian.net/browse/AC-5796) to that end. That ticket (and the one about a friendly URL for the endpoint) are linked to the ticket.

PR #172, opened 2018-07-01
[AC-5768] - Integrate The Directory To Our Build And Release Process

yotammanor:
**Changes Introduced in [AC-5768](https://masschallenge.atlassian.net/browse/AC-5768):**
- infer the Directory revision to use as django-accelerator is inferred.
- add a DIRECTORY constant in makefile

**Test:**
- `make checkout` now checks out directory as well.
- see that `make release` would do the same (I suggest not actually testing it because reverting those changes would be laborious.
- testing travis:
  - Checkout a separate branch on top of this branch.
  - In travis.yml, in line 18 (right below `before_install`, add: `- echo $MENTOR_DIRECTORY_REVISION`
  - now try the following scenarios:
    - commit and push, see that the variable gets assigned the default branch - development
    - go to Directory, create a branch with the name of the off-branch, push. Restart the impact-api build. See that the variable is now the name of the branch.
    - delete the branch. Create a tag with that same name. restart the impact-api build. See that the behaviour is the same. delete the tag. 
    - use the override mechanism - in travis, go to settings, and add an env var `DIRECTORY_REVISION` with some value. restart the build. see that this value is the one that appears. delete this env var from the settings.

**Note:**
- This ticket does not address @pbx's concerns as expressed in https://masschallenge.atlassian.net/browse/AC-5759 , and does not address further points in https://masschallenge.atlassian.net/browse/AC-5783 which need to be discussed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,6 @@ env:
     - MENTOR_DIRECTORY_REVISION=`./infer_repository_revision.sh "$BRANCH" https://$GITHUB_ACCESS_TOKEN@github.com/masschallenge/directory.git`
 
 before_install:
-- echo $MENTOR_DIRECTORY_REVISION
-- echo $DJANGO_ACCELERATOR_REVISION
 - echo -e "machine github.com\n  login $GITHUB_ACCESS_TOKEN" >> ~/.netrc
 - sudo apt-add-repository 'deb http://archive.ubuntu.com/ubuntu trusty-backports universe'
 - sudo apt-get update -qq
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,8 @@ env:
     - IS_RELEASE=$(git ls-remote origin | grep "$TRAVIS_COMMIT\s\+refs/heads/release$")
     - RELEASE_TAG=$(git tag -l --points-at HEAD | head -n1)
     - BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -11,9 +11,12 @@ env:
     - IS_RELEASE=$(git ls-remote origin | grep "$TRAVIS_COMMIT\s\+refs/heads/release$")
     - RELEASE_TAG=$(git tag -l --points-at HEAD | head -n1)
     - BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then echo $TRAVIS_BRANCH; else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
-    - DJANGO_ACCELERATOR_REVISION=`./infer_django_accelerator_revision.sh $BRANCH`
+    - DJANGO_ACCELERATOR_REVISION=`./infer_repository_revision.sh "$BRANCH" https://www.github.com/masschallenge/django-accelerator.git`
+    - MENTOR_DIRECTORY_REVISION=`./infer_repository_revision.sh "$BRANCH" https://$GITHUB_ACCESS_TOKEN@github.com/masschallenge/directory.git`
 
 before_install:
+- echo $MENTOR_DIRECTORY_REVISION
+- echo $DJANGO_ACCELERATOR_REVISION
```
shanbady: remove echo calls or add more context so its useful output.

shanbady (changes requested):

.travis.yml
yotammanor: Used it for debugging, thought I removed it already.

yotammanor:
@shanbady this is ready for re-review.

shanbady (approved):

PR #180, opened 2018-07-17
AC-5700: allocate_applications API call

mc-nathan:
** BE SURE TO MERGE https://github.com/masschallenge/django-accelerator/pull/77 IN django-accelerator BEFORE THIS **
(or immediately after if this happens to get merged first)

To test:
- Be sure to get the AC-5700 branch from django-accelerator
- Load a special database that gets a judging round setup properly.  E.g., `make load-db s3_key=AC-5700.sql.gz`.  You are expected to get warnings about "changes that are not yet reflected in a migration" that can be safely ignored.
- Start both the accelerate server and the impact-api server.
- Login to your local accelerate server as a confirmed judge for an active judging round (e.g., 291-user@example.com for the PULSE 2018 Round 1 (id=56))
- You should see the /panels page with a message at the bottom that says "Thank You! Can you review more?"
- Still logged in as the judge go to http://localhost:8000/api/v1/allocate_applications/56/291/
- This should return a list of application ids that they have been assigned.
- Reload the /panels page and see the assigned applications.
- If you run http://localhost:8000/api/v1/allocate_applications/56/291/ again it should give an error.
- Extra credit for the interested developer: Go through the review process for this judge, e.g. mark everything as "not judged conflict".  Mark the panel as done.
- You should now get a new set of application if you run http://localhost:8000/api/v1/allocate_applications/56/291/ again.


web/impact/impact/tests/test_allocate_applications_view.py
```diff
@@ -0,0 +1,116 @@
...
+            url = reverse(AllocateApplicationsView.view_name,
+                          args=[judging_round.id, judge_id])
+            response = self.client.get(url)
+            assert response.status_code == 200
+
+    def test_get_judging_round_inactive(self):
+        context = JudgeFeedbackContext()
+        judging_round = context.judging_round
+        judging_round.is_active = False
+        judging_round.save()
```
jonkiparsky: this seems like it should be a simple modification to the Context

web/impact/impact/tests/test_allocate_applications_view.py
```diff
@@ -0,0 +1,116 @@
...
+            response = self.client.get(url)
+            assert response.status_code == 200
+            commitment.refresh_from_db()
+            assert commitment.capacity > 0
+            assert commitment.current_quota > 0
+
+    def test_get_by_judge_succeeds(self):
+        context = JudgeFeedbackContext()
+        judging_round = context.judging_round
+        judge = ExpertFactory()
```
jonkiparsky: Not a blocker, but would prefer to see a one-liner to get a loggable-in expert.

web/impact/impact/tests/test_allocate_applications_view.py
```diff
@@ -0,0 +1,116 @@
...
+        judge.set_password("password")
+        judge.save()
+        with self.login(email=judge.email):
+            url = reverse(AllocateApplicationsView.view_name,
+                          args=[judging_round.id, judge.id])
+            response = self.client.get(url)
+            assert response.status_code == 200
```
jonkiparsky, line 108: Anything we can test about the content? status_code == 200 is a pretty wide net.

web/impact/impact/tests/test_allocate_applications_view.py
```diff
@@ -0,0 +1,116 @@
...
+        context = JudgeFeedbackContext()
+        judging_round = context.judging_round
+        judge = ExpertFactory()
+        judge.set_password("password")
+        judge.save()
+        with self.login(email=judge.email):
+            url = reverse(AllocateApplicationsView.view_name,
+                          args=[judging_round.id, ExpertFactory().id])
+            response = self.client.get(url)
+            assert response.status_code == 403
```
jonkiparsky, line 120: Likewise, I'd like to test the content here, not just the status_code.

jonkiparsky (changes requested):
Generally looks good. A few minor quibbles.

web/impact/impact/v1/views/allocate_applications_view.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def __init__(self):
+        self.errors = []
+
+    def get(self, request, round_id, judge_id):
+        self.judging_round = JudgingRound.objects.get(pk=round_id)
+        self.judge = User.objects.get(pk=judge_id)
+        self.scenario = self._find_scenario()
+        self._validate()
+        if self.errors:
```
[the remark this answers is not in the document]
mc-nathan: Good idea, but I'm pushing back for this ticket.  Would you mind making a ticket for it?

web/impact/impact/v1/views/allocate_applications_view.py
```diff
@@ -0,0 +1,137 @@
...
+    def _make_panel(self, choices):
+        self._enable_additional_assignments(len(choices))
+        panel = Panel(status=ACTIVE_PANEL_STATUS)
+        panel.save()
+        jpa = JudgePanelAssignment(
+            judge=self.judge,
+            panel=panel,
+            scenario=self.scenario,
+            assignment_status=ASSIGNED_PANEL_ASSIGNMENT_STATUS)
+        jpa.save()
```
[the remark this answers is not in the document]
mc-nathan: Ah, I was missing `.objects`.  I tried JPA.create and that failed.

web/impact/impact/tests/test_allocate_applications_view.py
mc-nathan: Looked at this for a bit and I didn't see a quick way to do it.  I tried just adding `password="password"` which looks like it should work based on the code, but it doesn't and it smells like a rabbit hole.  Left a comment.

web/impact/impact/tests/test_allocate_applications_view.py
mc-nathan: Pushing back.  All I'm testing is that the judge can access the page so this seems like the right test to me.  We are testing content elsewhere.  Exactly what it means to run this API call as the judge vs. as a staff member could at least in theory change.  Note that this is the same check that we have for test_get.

web/impact/impact/tests/test_allocate_applications_view.py
mc-nathan: Pushing back. Again, it seems to me that the thing we care about here actually is the status code.  No error message is sent because it's caught by the permission handling mechanism (see allocate_applications_permissions.py).

web/impact/impact/v1/views/allocate_applications_view.py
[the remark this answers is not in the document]
jonkiparsky: ✅

jonkiparsky (approved):

PR #181, opened 2018-07-17
[AC-5818] Add weight and count to analyze_judging_round call

jonkiparsky:
Trivial modification to existing call.


mc-nathan (approved):
Pretty straight forward.  I could ask for a test, but I think I won't.

PR #183, opened 2018-07-24
[AC-5836] Add API call for cloning criteria

jonkiparsky:
To test: Create a new judging round and get its id, call this TARGET. Find a judging round for which criteria exist and get its id, call this SOURCE. 
hit http://localhost:8000/api/v1/clone_criteria/SOURCE/TARGET to clone criteria

http://localhost:8000/api/v1/criterion/?judging_round=SOURCE
should return the original criteria
http://localhost:8000/api/v1/criterion/?judging_round=TARGET
should return the new criteria

### DIFF
--- a/web/impact/impact/v1/views/clone_criteria_view.py
+++ b/web/impact/impact/v1/views/clone_criteria_view.py
@@ -31,6 +31,8 @@ ANALYZE_JUDGING_ROUND_FIELDS = {
     "remaining_capacity": READ_ONLY_INTEGER_FIELD,
 }
 
+ROUND_DOES_NOT_EXIST_ERROR = "Judging Round {} does not exist"
+
 
 class CloneCriteriaView(ImpactView):
     view_name = "clone_criteria"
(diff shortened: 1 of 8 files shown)


web/impact/impact/v1/views/clone_criteria_view.py
```diff
@@ -0,0 +1,40 @@
...
+
+class CloneCriteriaView(ImpactView):
+    view_name = "clone_criteria"
+    model = JudgingRound
+
+    @classmethod
+    def fields(cls):
+        return {}
+
+    def get(self, request, source_pk, target_pk):
```
shanbady: since we are creating a new resource this should be a 'post' method rather than a GET since get requests should not alter state.

web/impact/impact/v1/views/clone_criteria_view.py
```diff
@@ -0,0 +1,40 @@
...
+        self._validate_judging_round_exists(source_pk)
+        self._validate_judging_round_exists(target_pk)
```
shanbady: not sure if the permission checking is handled upstream somewhere but does this protect against non-staff users making calls?

shanbady (changes requested):

shanbady:
@jonkiparsky looks like there is aconflict to resolve. otherwise looks good.

jonkiparsky:
@shanbady conflict resolved, added a test, back to you

shanbady (approved):

PR #185, opened 2018-07-26
AC-5703: Implement the dynamic matrix algorithm

mc-nathan:
To test:
- Check out the AC-5703 branch for django-accelerator.  (PR: https://github.com/masschallenge/django-accelerator/pull/81)
- Start your local accelerate and impact-api servers.
- Login to accelerate as the 291-user@example.com.  Note that it goes to the /panels page and no applications are listed in the interface.
- As that user in a separate window go to http://localhost:8000/api/v1/allocate_applications/56/291/
- This should respond with a list of applications.
- Reloading the /panels page on the first tab should now show applications.
- Ensuring that these are really the right applications is hard.  So far it has primarily been tested through the automated tests which establish some basic cases that we would want to be true.
- To review I would start with the tests and consider adding some more tests for other interesting cases.  A few suggestions appear in the code.

### DIFF
--- a/web/impact/impact/tests/test_allocate_applications_view.py
+++ b/web/impact/impact/tests/test_allocate_applications_view.py
@@ -1,11 +1,6 @@
 # MIT License
 # Copyright (c) 2017 MassChallenge, Inc.
 
-# TODO: Coverage, tests that make reasonable selections
-# Need to account for completed feedback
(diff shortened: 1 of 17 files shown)


jonkiparsky (approved):
Looks good!

PR #188, opened 2018-08-07
[AC-5778] Add current mentees and previous mentees to expert profile object type

mkiterian:
#### Changes added:
- Add a StartupMentorRelationshipType to represent the previous and current mentee fields and add it to the ExpertProfileType

#### Test:
- ~Ensure you have pulled and checkout out into the `AC-5778` branch of the `django-accelerator` to have new properties~ This is no longer needed. Current implementation is done without adding the related_name ref to the mentor field on the StartupMentorRelationship model.
Please discard changes on the `django-accelerator` `AC-5778` branch cc @jonkiparsky 
- Run the impact api with `make run-server`
- Open graphiql by going to the graphql endpoint in the browser e.g on localhost go to `http://localhost:8000/graphql/`
- Create a query to retrieve an expert profile and specify currentMentees and previousMentees as fields e.g

```
{
  expertProfile(id:ID){
    user{
      firstName
    }
    currentMentees{
      startupId
      startupName
      startupHighResolutionLogo
      startupShortPitch
      programLocation
      programYear
      programStatus
    }
    previousMentees{
      startupId
      startupName
      startupHighResolutionLogo
      startupShortPitch
      programLocation
      programYear
      programStatus
    }
  }
}
```

### DIFF
--- a/web/impact/impact/graphql/types/expert_profile_type.py
+++ b/web/impact/impact/graphql/types/expert_profile_type.py
@@ -13,7 +13,9 @@ from accelerator_abstract.models import (
     ENDED_PROGRAM_STATUS,
 )
 from impact.graphql.types.industry_type import IndustryType  # noqa: F401
-from impact.graphql.types.startup_mentor_relationship_type import StartupMentorRelationshipType  # noqa: E501
+from impact.graphql.types.startup_mentor_relationship_type import (
+    StartupMentorRelationshipType,
+)
 from impact.graphql.types.program_family_type import ProgramFamilyType  # noqa: F401, E501
 from impact.graphql.types.user_type import UserType  # noqa: F401
 from impact.graphql.types.functional_expertise_type import FunctionalExpertiseType  # noqa: F401, E501
@@ -86,26 +88,10 @@ class ExpertProfileType(DjangoObjectType):
                     mentor_id=self.user.id))
 
     def resolve_current_mentees(self, info, **kwargs):
-        mentee_filter = compose_filter([
-            'startup_mentor_tracking',
-            'program',
-            'program_status'
-        ], ACTIVE_PROGRAM_STATUS)
-        return self.user.startup_mentor_relationships.filter(
-            status=CONFIRMED_RELATIONSHIP,
-            **mentee_filter
-            )
+        return _get_mentees(self.user, ACTIVE_PROGRAM_STATUS)
 
     def resolve_previous_mentees(self, info, **kwargs):
-        mentee_filter = compose_filter([
-            'startup_mentor_tracking',
-            'program',
-            'program_status'
-        ], ENDED_PROGRAM_STATUS)
-        return self.user.startup_mentor_relationships.filter(
-            status=CONFIRMED_RELATIONSHIP,
-            **mentee_filter
-            )
+        return _get_mentees(self.user, ENDED_PROGRAM_STATUS)
 
 
 def _get_user_programs(user):
@@ -118,3 +104,15 @@ def _get_user_programs(user):
     )
     return Program.objects.filter(
         programrole__in=user_program_roles_as_participant).distinct()
+
+
+def _get_mentees(user, program_status):
+    mentee_filter = compose_filter([
+        'startup_mentor_tracking',
+        'program',
+        'program_status'
+    ], program_status)
+    return user.startup_mentor_relationships.filter(
+        status=CONFIRMED_RELATIONSHIP,
+        **mentee_filter
+    )
--- a/web/impact/impact/graphql/types/expert_profile_type.py
+++ b/web/impact/impact/graphql/types/expert_profile_type.py
@@ -6,18 +6,26 @@ from accelerator.models import (
     UserRole,
     ProgramRole,
     Program,
+    CONFIRMED_RELATIONSHIP,
+)
+from accelerator_abstract.models import (
+    ACTIVE_PROGRAM_STATUS,
+    ENDED_PROGRAM_STATUS,
 )
 from impact.graphql.types.industry_type import IndustryType  # noqa: F401
-from impact.graphql.types.startup_type import StartupType
+from impact.graphql.types.startup_mentor_relationship_type import StartupMentorRelationshipType  # noqa: E501
 from impact.graphql.types.program_family_type import ProgramFamilyType  # noqa: F401, E501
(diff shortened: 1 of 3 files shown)


web/impact/impact/graphql/types/expert_profile_type.py
```diff
@@ -88,3 +98,10 @@ def _get_user_programs(user):
     )
     return Program.objects.filter(
         programrole__in=user_program_roles_as_participant).distinct()
+
+
+def _get_mentees(expert_id):
+    user_id = ExpertProfile.objects.get(id=expert_id).user_id
+    relationships = StartupMentorRelationship.objects.filter(
+        mentor_id=user_id).all()
+    return [r.startup_mentor_tracking.startup for r in relationships]
```
yotammanor: A more django-esque way to write this function would be:

```
def resolve_previous_mentees(self, info, **kwargs):
    return _get_mentees(self.user, ENDED_PROGRAM_STATUS)

def resolve_current_mentees(self, info, **kwargs):
    return _get_mentees(self.user, ACTIVE_PROGRAM_STATUS)

def _get_mentees(user, program_status): # pass in self.user
    return Startup.objects.filter(startupmentortrackingrecord__startupmentorrelationship__mentor=self.user, startupmentortrackingrecord__program__program_status=program_status)
```

Advantages of this approach:
- You return a queryset instead of a list, which keeps you in the django world (which is good because e.g. lazy evaluation, extra filtering later on if needed, and just generally more consistent interfaces).
- you filter for the program_status you want in the database layer instead of python, which is more efficient.
- you filter for the program_status directly on the mentor_tracking object, which gives you the status of the relevant program, instead of the first program for that startup, and this avoids errors.
- you return startups objects right away, so no need to iterate over the results once again.

yotammanor (changes requested):
Some suggested changes on how to resolve the data correctly.

web/impact/impact/graphql/types/expert_profile_type.py
```diff
@@ -77,6 +85,28 @@ def resolve_office_hours_url(self, info, **kwargs):
...
+            'startup_mentor_tracking',
+            'program',
+            'program_status'
+        ], ACTIVE_PROGRAM_STATUS)
+        return self.user.startup_mentor_relationships.filter(
+            status=CONFIRMED_RELATIONSHIP,
+            **mentee_filter
+            )
+
+    def resolve_previous_mentees(self, info, **kwargs):
```
jonkiparsky: This function and the previous one differ only in the PROGRAM_STATUS that they select on. I would like to see them combined into a single function, taking the PROGRAM_STATUS as an argument. 

EDIT: unless they are called by graphql code and need to be separate functions with these arg structures. In that case, I'd want to factor out the common code into a helper function, and the `resolve_previous_mentees` and `resolve_current_mentees` functions turned into simple proxies calling that helper with the appropriate PROGRAM_STATUS.

web/impact/impact/graphql/types/expert_profile_type.py
```diff
@@ -77,6 +85,28 @@ def resolve_office_hours_url(self, info, **kwargs):
                 '?mentor_id={mentor_id}'.format(
                     mentor_id=self.user.id))
 
+    def resolve_current_mentees(self, info, **kwargs):
```
jonkiparsky: `info` and `kwargs` are not used.

web/impact/impact/graphql/types/expert_profile_type.py
jonkiparsky: `info` and `kwargs` are not used.

web/impact/impact/tests/test_graphql.py
```diff
@@ -5,7 +5,7 @@
 
 from impact.graphql.middleware import NOT_LOGGED_IN_MSG
 from impact.tests.api_test_case import APITestCase
-from impact.tests.factories import ExpertFactory
+from impact.tests.factories import ExpertFactory, StartupMentorRelationshipFactory  # noqa: E501
```
jonkiparsky: MC style has imports on separate lines:
```
from impact.tests.factories import (
    ExpertFactory, 
    StartupMentorRelationshipFactory,
)
```

Note trailing comma on last line, to facilitate addition of new imports. 
Also note that components are sorted alphabetically within each import.

web/impact/impact/tests/test_graphql.py
```diff
@@ -56,6 +58,60 @@ def test_query_with_expert(self):
...
+                    }
+                }
+            )
+
+    def test_requested_fields_for_startup_mentor_relationship_type(self):
+        with self.login(email=self.basic_user().email):
+            obj = StartupMentorRelationshipFactory()
+            startup = obj.startup_mentor_tracking.startup
+            program = obj.startup_mentor_tracking.program
+            query = """
```
jonkiparsky: This long literal string is a little awkward. Not a blocker, but I would prefer to have it at least externalized as a constant. Ideally, we'd write some code to generate this stuff, but that's not something we should hold this ticket up for.

web/impact/impact/graphql/types/expert_profile_type.py
```diff
@@ -6,18 +6,26 @@
...
     Program,
+    CONFIRMED_RELATIONSHIP,
+)
+from accelerator_abstract.models import (
+    ACTIVE_PROGRAM_STATUS,
+    ENDED_PROGRAM_STATUS,
 )
 from impact.graphql.types.industry_type import IndustryType  # noqa: F401
-from impact.graphql.types.startup_type import StartupType
+from impact.graphql.types.startup_mentor_relationship_type import StartupMentorRelationshipType  # noqa: E501
```
jonkiparsky: long imports can be broken up with parentheses:
```
from impact.graphql.types.startup_mentor_relationship_type import (
    StartupMentorRelationshipType,
)
```

jonkiparsky (changes requested):
A few style changes to work on while I look at the functionality.

jonkiparsky:
Reviewed functionality. Looks good there.

jonkiparsky (approved):

jonkiparsky:
@yotammanor Are you good with the fixes that you asked for? If so, please approve and merge.

yotammanor (approved):
Approved, merge when Directory PR is ready.

PR #191, opened 2018-08-09
[AC-5495] - Allow for custom tagging when using `make release`

VincentHokie:
Changes Introduced in [AC-5495](https://masschallenge.atlassian.net/browse/AC-5495):

- allow use of `TAG` argument when using `make release` for custom tagging

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -142,7 +142,9 @@ target_help = \
   '\tRelease name defaults release-<version>.<number> where <version> is' \
   '\tthe first line of impact-api/VERSION and <number> is the next unused' \
   '\tnon-negative integer (0,1,2,...).' \
-  '\t$$(release_name) overrides the entire release name.' \
+  '\t$$(TAG) overrides the entire release name.' \
+  '\tThe TAG can be set by adding an optional TAG="custom-tag" to the command' \
+  '\te.g. make release TAG="XX-XXXX" ' \
   'deploy - Deploy $$(release_name) to a $$(target).' \
   '\tValid targets include "staging" (the default), "production",' \
   '\t "test-1", and "test-2"' \
@@ -403,8 +405,6 @@ release-list:
 
 
 release:
-	@git commit --allow-empty -m "generating a new release"
-	@git push
 	@bash create_release.sh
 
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -403,8 +405,6 @@ release-list:
 
 
 release:
-	@git commit --allow-empty -m "generating a new release"
-	@git push
```
shanbady: why was this removed? there was a reason for this change - without it you cannot tag repos without changes in them. for example - if you wanted to release some changes in accelerate but impact and directory should point to the latest master

shanbady (changes requested):

create_release.sh
```diff
@@ -1,7 +1,20 @@
-docker run --rm  -v"$(pwd)":/app  -ti semantic-release  -- semantic-release version
-export pwd=$(pwd)
-export TAG=$(docker run --tty=false --rm  -v$pwd:/app  -i semantic-release -- semantic-release version --noop | grep "Current version: " | cut -d ' ' -f 3 | sed -e "s/\r//")
+
+# if a tag hasn't been passed in as a parameter, create a semantic one
+if [ ! -n "$TAG" ]; then
+    git commit --allow-empty -m "generating a new release"
+	git push
```
shanbady: spacing issue

shanbady (changes requested):

create_release.sh
```diff
@@ -1,7 +1,20 @@
-docker run --rm  -v"$(pwd)":/app  -ti semantic-release  -- semantic-release version
-export pwd=$(pwd)
-export TAG=$(docker run --tty=false --rm  -v$pwd:/app  -i semantic-release -- semantic-release version --noop | grep "Current version: " | cut -d ' ' -f 3 | sed -e "s/\r//")
+
+# if a tag hasn't been passed in as a parameter, create a semantic one
+if [ ! -n "$TAG" ]; then
+    git commit --allow-empty -m "generating a new release"
```
shanbady: i see - this was moved here. it may be necessary regardless if we use custom tags or not

shanbady (changes requested):

Makefile
VincentHokie: put this back here, it makes sense that both with an without a tag, we should be doing these.

shanbady (approved):

PR #195, opened 2018-08-15
[AC-5889] Make the new person profiles reference User ID and not Expert ID

mkiterian:
#### Changes introduced in [AC-5889](https://masschallenge.atlassian.net/browse/AC-5889):
- Retrieve `user id` from API request instead of the `expert id` and query the Expert Profile using this user id

#### How to test

- Run the graphiql interface on `http://localhost:8000/graphql`
- Run this query:
```
{
  expertProfile(id:user_id){
    user{
      firstName
    }
    bio
  }
}
```
If the user id is valid and belongs to an expert you should get the correct response. Otherwise the response should be null

### DIFF
--- a/web/impact/impact/graphql/query.py
+++ b/web/impact/impact/graphql/query.py
@@ -14,7 +14,6 @@ class Query(graphene.ObjectType):
 
     def resolve_expert_profile(self, info, **kwargs):
         user_id = kwargs.get('id')
-
         User = get_user_model()
 
         try:
--- a/web/impact/impact/graphql/query.py
+++ b/web/impact/impact/graphql/query.py
@@ -1,12 +1,7 @@
 import graphene
-from django.contrib.auth import get_user_model
-from django.core.exceptions import ObjectDoesNotExist
-from graphql import GraphQLError
 
 from impact.graphql.types.expert_profile_type import ExpertProfileType
-from accelerator_abstract.models.base_base_profile import (
-    EXPERT_USER_TYPE,
-)
+from accelerator.models.expert_profile import ExpertProfile
 
 
 class Query(graphene.ObjectType):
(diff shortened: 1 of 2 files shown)


web/impact/impact/graphql/query.py
```diff
@@ -1,16 +1,29 @@
...
     def resolve_expert_profile(self, info, **kwargs):
-        expert_profile_id = kwargs.get('id')
+        user_id = kwargs.get('id')
+        User = get_user_model()
 
-        if expert_profile_id is not None:
-            return ExpertProfile.objects.get(pk=expert_profile_id)
+        try:
+            user = User.objects.get(id=user_id)
+            profile = user.get_profile()
```
yotammanor: It's a good habit to keep `try` blocks as small as possible, so you know where things fail, and also keep the code flatter.
So I'd suggest this one to be written:
```
try:
    user = User.objects.get(id=user_id)
except ObjectDoesNotExist:
    raise GraphQLError(USER_NOT_FOUND_MESSAGE)
# ... rest of code
```

That being said, as long as there's no good reason to differentiate the case of non-existing user and an existing user is not a mentor (there might be such reason, but maybe not at the moment), I'd stick with my suggestion from earlier:
```
def resolve_expert_profile(self, info, **kwargs):
        user_id = kwargs.get('id')
        return ExpertProfile.objects.filter(user_id=user_id).first()
```

This will always return None for both cases mentioned above, and also has the benefit of being a simpler function and about a quarter the amount of queries executed :)

mkiterian: This ok for now since we are only getting expert profiles. I used `get_profile` to get the profile of whichever type of user the id belongs to since this will eventually be refactored to get all user profiles not just for experts.  I'll refactor to only handle experts for now and remove the error handling until we've figured that out

mkiterian: @yotammanor using this `return ExpertProfile.objects.filter(user_id=user_id).first()` does not raise an exception hence the response does not indicate an error happened and sends back null  which causes the Profile page to try and render null since no error was received. I'm reverting to using `ExpertProfile.objects.get`

pbx (approved):
Looks good, works as described. Fixed a typo in a method name.

PR #205, opened 2018-08-29
[AC-5853] Add tests covering some allocator edge cases

jonkiparsky:
These tests were requested in the original review of the allocator, but deferred since Nathan did not have time to get them in before he left. 
Note that a context change in django-accelerator is required for this code to run. (https://github.com/masschallenge/django-accelerator/pull/91)


pbx (approved):
Looks good, tests pass.

[Oops -- have a blocking question that came up after I merged; see the other PR.]

PR #207, opened 2018-09-05
AC-5961 Add high level Application information to Salesforce API

mkiterian:
Changes introduced in [AC-5961](https://masschallenge.atlassian.net/browse/AC-5961):
- Add application list view to impact V1
- Add application detail view to impact V1

How to test (on localhost):
- Make sure you are logged in as a user who belongs to the `v1_clients` user group. (You can add the user to the group here `http://localhost:8181/admin`)
-Enter the url `http://localhost:8000/api/v1/application/` in the browser to view the results of a GET request to that route
- -Enter the url `http://localhost:8000/api/v1/application/{application_id}` in the browser to view the results of a GET request to that route

### DIFF
--- a/web/impact/impact/tests/test_application_detail_view.py
+++ b/web/impact/impact/tests/test_application_detail_view.py
@@ -1,0 +1,61 @@
+# MIT License
+# Copyright (c) 2017 MassChallenge, Inc.
+
+import json
+from jsonschema import Draft4Validator
+
+from django.urls import reverse
+from impact.tests.factories import ApplicationFactory
+from impact.tests.api_test_case import APITestCase
+from impact.tests.utils import assert_fields
+from impact.v1.views import ApplicationDetailView
+
+APPLICATION_GET_FIELDS = [
+    "id",
+    "created_at",
+    "updated_at",
+    "application_status",
+    "cycle",
+    "startup",
+    "application_type",
+]
+
+
+class TestApplicationDetailView(APITestCase):
+    def test_get(self):
+        application = ApplicationFactory()
+        with self.login(email=self.basic_user().email):
+            url = reverse(ApplicationDetailView.view_name,
+                          args=[application.id])
+            response = self.client.get(url)
+            assert response.data["created_at"] == application.created_at
+            assert response.data["updated_at"] == application.updated_at
+            assert (response.data["application_status"] ==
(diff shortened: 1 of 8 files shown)


web/impact/impact/v1/views/application_list_view.py
```diff
@@ -0,0 +1,10 @@
+# MIT License
+# Copyright (c) 2017 MassChallenge, Inc.
+
+from impact.v1.views.base_list_view import BaseListView
```
pbx: We should sort these (as above)

web/impact/impact/tests/test_application_detail_view.py
```diff
@@ -0,0 +1,61 @@
+# MIT License
+# Copyright (c) 2017 MassChallenge, Inc.
+
+import json
+from jsonschema import Draft4Validator
+
+from django.urls import reverse
+from impact.tests.factories import ApplicationFactory
```
pbx, line 8: Should be a line break between Django imports and Accelerate imports

web/impact/impact/tests/test_application_detail_view.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class TestApplicationDetailView(APITestCase):
+    def test_get(self):
+        application = ApplicationFactory()
+        with self.login(email=self.basic_user().email):
+            url = reverse(ApplicationDetailView.view_name,
+                          args=[application.id])
+            response = self.client.get(url)
+            assert response.data["created_at"] == application.created_at
```
pbx, line 31: I know there are many other tests which use raw `assert a == b`, but I prefer that we  use `assertEqual` wherever possible. One benefit is that, when it fails, it tells you what the values are. Plain `assert` does not do that, producing more work. It's acceptable for "this will never fail" preconditions not directly tied to the object of the test, but that's about it.

web/impact/impact/v1/urls.py
```diff
@@ -1,8 +1,10 @@
 from django.conf.urls import url
 
 from impact.v1.views import (
+    ApplicationDetailView,
```
pbx: These need sorting

pbx (approved):
Looks good, works as advertised. I made a few small changes needed (sorting imports).

PR #208, opened 2018-09-05
AC-5970 - fix impact-api dependencies

pbx:
## [AC-5970](https://masschallenge.atlassian.net/browse/AC-5970)

Two of our unpinned dependencies got version bumps in the past few days that broke compatibility with Django 1.10. This PR locks them to compatible versions.

(The diff is noisy because I also sorted the reqs I was editing.)

To test: check out the branch and run `make test`, or look at Travis output. You should see no errors and no failures.

To reproduce the errors that this PR fixes: check out impact API at 7fc4974 or earlier (e.g. v1.0.44), run `make build && make test`, and see 240+ errors.

### DIFF
--- a/web/impact/requirements/base.txt
+++ b/web/impact/requirements/base.txt
@@ -32,7 +32,7 @@ django-embed-video
 django-filter # Filtering support
 django-oauth-toolkit<1.1
 django-redis-cache
-django-registration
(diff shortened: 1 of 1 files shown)


jonkiparsky (approved):
Changes look sane, tests pass

PR #210, opened 2018-09-07
[Ac 5901] Reduce number of queries for analyze_judging_round

jonkiparsky:
Plucked a bunch of low-hanging fruit, made follow-on tickets for the remaining (hard) work. 
AC-5981 and AC-5982 are the follow-ons

To test:
* check out development. 
* load clean db
* working as a user with correct permissions (v1 group, unless permissions PR has merged) hit http://localhost:8000/api/v1/analyze_judging_round/48/
* check out AC-5901
* hit the same endpoint in a new tab
* compare the two: results should be identical, and debug toolbar should show 153 queries on development and 82 on the branch. Load time,  CPU time, and DB time on the branch should be reduced relative to development.

### DIFF
--- a/web/impact/impact/tests/test_analyze_judging_round_view.py
+++ b/web/impact/impact/tests/test_analyze_judging_round_view.py
@@ -11,9 +11,6 @@ from accelerator.tests.contexts import AnalyzeJudgingContext
 from impact.tests.api_test_case import APITestCase
 from impact.tests.utils import assert_fields
 from impact.v1.views import AnalyzeJudgingRoundView
-from impact.v1.helpers.matching_program_criterion_helper import (
-    MatchingProgramCriterionHelper,
-)
 
 
 class TestAnalyzeJudgingRoundView(APITestCase):
@@ -115,8 +112,6 @@ class TestAnalyzeJudgingRoundView(APITestCase):
                                         options=[""])
         # This is where I gave up and just hard coded it!
         dists = {context.program.program_family.name: {0: 1, 1: 1}}
-        # clear cache since we added a new program family!
-        MatchingProgramCriterionHelper.program_families = None
         self.assert_option_distributions(context, dists)
 
     def assert_option_distributions(self, context, dists):
--- a/web/impact/impact/tests/test_analyze_judging_round_view.py
+++ b/web/impact/impact/tests/test_analyze_judging_round_view.py
@@ -11,6 +11,9 @@ from accelerator.tests.contexts import AnalyzeJudgingContext
 from impact.tests.api_test_case import APITestCase
 from impact.tests.utils import assert_fields
 from impact.v1.views import AnalyzeJudgingRoundView
+from impact.v1.helpers.matching_program_criterion_helper import (
+    MatchingProgramCriterionHelper,
+)
 
 
 class TestAnalyzeJudgingRoundView(APITestCase):
@@ -112,6 +115,8 @@ class TestAnalyzeJudgingRoundView(APITestCase):
                                         options=[""])
         # This is where I gave up and just hard coded it!
         dists = {context.program.program_family.name: {0: 1, 1: 1}}
+        # clear cache since we added a new program family!
+        MatchingProgramCriterionHelper.program_families = None
         self.assert_option_distributions(context, dists)
 
     def assert_option_distributions(self, context, dists):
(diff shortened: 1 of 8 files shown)


web/impact/impact/v1/helpers/matching_program_criterion_helper.py
```diff
@@ -12,13 +12,21 @@
...
+    program_families = None
 
     def __init__(self, subject):
         super().__init__(subject)
         self._program_name_cache = None
 
+    @classmethod
+    def _program_family_map(cls):
+        if cls.program_families is None:
+            cls.program_families = cls.cache_instances_by_name(ProgramFamily)
```
shanbady: how does this behave if this class is loaded in memory and we were to add/remove program families?

web/impact/impact/v1/helpers/matching_program_criterion_helper.py
```diff
@@ -12,13 +12,21 @@
...
+        return cls.program_families
+
     def app_ids_for_feedbacks(self, feedbacks, option_name, applications):
-        target = ProgramFamily.objects.filter(name=option_name).first()
+        target = self.__class__._program_family_map()[option_name]
```
shanbady: if you are accessing a class attribute use MatchingProgramCriterionHelper._program_family_map()

web/impact/impact/v1/helpers/matching_industry_criterion_helper.py
```diff
@@ -12,13 +12,20 @@
...
 
+    @classmethod
+    def _industry_map(cls):
+        if cls.industries is None:
+            cls.industries = cls.cache_instances_by_name(Industry)
+        return cls.industries
+
     def app_ids_for_feedbacks(self, feedbacks, option_name, applications):
-        target = Industry.objects.filter(name=option_name).first()
+        target = self.__class__._industry_map()[option_name]
```
shanbady, line 28: use MatchingIndustryCriterionHelper. _industry_map() instead

web/impact/impact/tests/test_analyze_judging_round_view.py
```diff
@@ -112,6 +115,8 @@ def test_get_with_program_criterion(self):
                                         options=[""])
         # This is where I gave up and just hard coded it!
         dists = {context.program.program_family.name: {0: 1, 1: 1}}
+        # clear cache since we added a new program family!
+        MatchingProgramCriterionHelper.program_families = None
```
shanbady: I see that you are having to clear the cache manually here in the test but how/where does this occur in the application? ( if we were to add a new program family in admin)

shanbady (changes requested):

web/impact/impact/v1/helpers/matching_program_criterion_helper.py
jonkiparsky: Sadly, that leads to a ridiculously long line

web/impact/impact/v1/helpers/matching_industry_criterion_helper.py
jonkiparsky: Sadly, that leads to a ridiculously long line

web/impact/impact/tests/test_analyze_judging_round_view.py
jonkiparsky: That is not a scenario we would expect to encounter. Planning for it is serious YAGNI

web/impact/impact/v1/helpers/matching_program_criterion_helper.py
jonkiparsky: It would break. Fortunately, we do not add program families in the course of normal workflows. Could make an attempt to be robust about this, but YAGNI.

web/impact/impact/v1/helpers/matching_program_criterion_helper.py
shanbady: im not seeing your point - you should still be able to use MatchingProgramCriterionHelper._program_family_map() and make it pass pep8

web/impact/impact/v1/helpers/matching_industry_criterion_helper.py
shanbady: you should still be able to get this pass pep8

web/impact/impact/tests/test_analyze_judging_round_view.py
shanbady: 3 questions:
- why do we expect to never add a program family?
- how would one clear the cache if we did change/add a program family?
- is there some cheap/efficient way to invalidate the cache in this scenario

web/impact/impact/v1/helpers/matching_program_criterion_helper.py
jonkiparsky:
```
>>> len("        target = MatchingProgramCriterionHelper._program_family_map()[option_name]")
82
>>> 
```

Can you suggest your preferred change? I don't see why `self.__class__` is an issue

shanbady: @pbx i would be interested to see what you think about this.

shanbady (approved):

PR #213, opened 2018-09-20
[AC-6021] Fix dependencies breaking impact-api build (django-registration, graphene-django)

pbx:
## [AC-6019](https://masschallenge.atlassian.net/browse/AC-6019)

The versions of **django-registration** and **graphene-django** that we were getting were not compatible with the version of Django that impact-api uses, v1.10.

graphene-django was formerly an unpinned dependency; it had a recent update that hadn't been noticed yet.

The too-new version of django-registration was an error from a previous ticket.

To test: run `make build` on this branch in impact-api and see that you do not get compatibility warnings of the type recorded on the ticket.

### DIFF
--- a/web/impact/requirements/base.txt
+++ b/web/impact/requirements/base.txt
@@ -37,7 +37,7 @@ django-oauth-toolkit<1.1
 django-paypal
 django-polymorphic==1.3
 django-redis-cache
-django-registration==2.5.2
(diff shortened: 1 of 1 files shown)


VincentHokie:
Does the thing and code changes are verified and pretty straightforward.

VincentHokie (approved):

PR #215, opened 2018-09-24
AC-5993 Improve handling of clearance-style permission checks on impact-api

pbx:
## [AC-5993](https://masschallenge.atlassian.net/browse/AC-5993)

This ticket arose because impact-api had a local, buggy check for clearance level (it failed if the user's permission was higher *or* lower than requested, not just if it was lower). 

Per the "ideal solution" described in the ticket, the changes here (across three repos) relocate the Clearance manager code to django-accelerator, and uses that for the impact-api `global_operations_manager_check`.

(Also see [accelerate PR](https://github.com/masschallenge/accelerate/pull/1894) and [django-accelerator PR](https://github.com/masschallenge/django-accelerator/pull/96))

**To test:**
* Log into as a user with `CLEARANCE_LEVEL_EXEC_MD` clearance
* Load a view that uses the `global_operations_manager_check` (e.g. [CloneCriteriaView](https://github.com/masschallenge/impact-api/blob/AC-5993/web/impact/impact/v1/views/clone_criteria_view.py)
* Confirm that you are allowed to see the page (i.e. you don't get a 403)


pbx:
Closing temporarily while I fix something.

jonkiparsky (approved):
I'm baffled as to why this worked, when my very similar attempt to do the same thing failed. I suppose it's because Paul has the special magic. In any case, it seems to work - clearances work on accelerate, and analyze_judging_view passes for users with GOM and exec/md clearance. Approving, though mystified.

shanbady (approved):

PR #223, opened 2018-10-11
AC-5994 - Allocator judging round selector needs to be optionally aware of user's clearance

pbx:
## [AC-5994](https://masschallenge.atlassian.net/browse/AC-5994)

This code changes the `JudgingRoundListView.filter` method to, by default, filter its returned values according to the user's clearances. 

(The control for this is a request parameter, so the user _could_ get the list of all program families -- including the ones they're not cleared for -- by manually adding that parameter to a request. This does not seem like a security hole to me, given that the exposed data is only program family names and IDs, but I thought it worth mentioning.)

**To test:** 
1. In **impact-api**, `make run-server`, log in as demoadmin
1. Add demoadmin to **v1_clients** group
1. Give demoadmin clearance for BOS ProgramFamily (see below)
1. In **application-allocator**, `yarn && yarn start`
1. go to http://localhost:1234/app-allocator-setup
1. Look at the choices in the judging round menu
1. Via admin or console, give demoadmin clearance for MX (see below)
1. Refresh the page, check the menu, and confirm that the ProgramFamily you added .a clearance for now appears. (You can also directly check the [output of the API call](http://localhost/api/v1/judging_round/?round_type=Online) used to build the menu)

```python
# To add demoadmin to v1_clients group
from accelerator.models import User
from django.contrib.auth.models import Group
Group.objects.get(name='v1_clients').user_set.add(User.objects.get(email__startswith="demoadmin"))

# To give demoadmin user a new clearance
from accelerator.models import *
Clearance.objects.create(
    user=User.objects.get(email__startswith="demoadmin"), 
    level=CLEARANCE_LEVEL_EXEC_MD, 
    program_family=ProgramFamily.objects.get(pk=1)) # BOS: 1; MX: 5
```

### DIFF
--- a/web/impact/impact/tests/test_judging_round_list_view.py
+++ b/web/impact/impact/tests/test_judging_round_list_view.py
@@ -7,10 +7,11 @@ from jsonschema import Draft4Validator
 from django.urls import reverse
 
 from accelerator.models import (
+    CLEARANCE_LEVEL_GLOBAL_MANAGER,
     IN_PERSON_JUDGING_ROUND_TYPE,
     ONLINE_JUDGING_ROUND_TYPE,
 )
-
+from accelerator.tests.factories.clearance_factory import ClearanceFactory
 from impact.tests.factories import JudgingRoundFactory
 from impact.tests.api_test_case import APITestCase
 from impact.tests.test_judging_round_detail_view import (
@@ -100,3 +101,24 @@ class TestJudgingRoundListView(APITestCase):
             response = self.client.get(self.url + "?round_type=bogus")
             assert response.status_code == 401
             assert response.data == [INVALID_ROUND_TYPE_ERROR.format("bogus")]
+
+    def test_clearance_enforced(self):
+        the_round = JudgingRoundFactory.create(
+            round_type=ONLINE_JUDGING_ROUND_TYPE)
+        url = self.url + "?round_type={}".format(ONLINE_JUDGING_ROUND_TYPE)
+        user = self.basic_user()
+        # This user does not have a clearance for the ProgramFamily
+        with self.login(email=user.email):
+            response = self.client.get(url)
+            results = response.data["results"]
+            round_ids = [item["id"] for item in results]
+            self.assertFalse(the_round.id in round_ids)
+        # Give the user a clearance for the ProgramFamily and check again
+        ClearanceFactory(level=CLEARANCE_LEVEL_GLOBAL_MANAGER,
+                         user=user,
+                         program_family=the_round.program.program_family)
+        with self.login(email=user.email):
+            response = self.client.get(url)
+            results = response.data["results"]
+            round_ids = [item["id"] for item in results]
+            self.assertTrue(the_round.id in round_ids)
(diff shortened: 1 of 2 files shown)


VincentHokie:
I'm 404ing on this url, `http://localhost:1234/app-allocator-setup`, am I missing something?

web/impact/impact/tests/test_judging_round_list_view.py
```diff
@@ -100,3 +110,41 @@ def test_get_round_type_cannot_be_bogus(self):
...
+        user = self.basic_user()
+        # This user does not have a clearance for the ProgramFamily
+        with self.login(email=user.email):
+            response = self.client.get(url)
+            results = response.data["results"]
+            round_ids = [item["id"] for item in results]
+            self.assertFalse(the_round.id in round_ids)
+        # Give the user clearance for relevant ProgramFamily and check again
+        _add_clearance(user, the_round)
+        with self.login(email=user.email):
```
VincentHokie, line 133: I'd be more comfortable if these test cases were split into 2 separate tests eg `test_user_without_clearances_doesnt_see_judging_round_of_corresponding_program_families` and `test_user_with_clearances_sees_judging_rounds_of_corresponding_program_families`

VincentHokie (changes requested):

VincentHokie:
I haven't been able to look at this locally due to the 404ing, but I had a thought concerning the possible security hole you mentioned. Given I have an active MC account and I add the `ignore_clearance` flag to the URL, would I be able to infer sensitive information given the UI functionality? (which again, I cannot see. yet :) )

Also, why are we implementing this functionality if it's so easy to get around? If it's important, I think we should add a permissions check (if not there) to whatever action the user can do while on the UI this is meant for. To ensure users can't perform any follow-on actions using the information they get from the dropdown.

pbx:
@VincentHokie - thanks for the review, I'll update with some changes. My instructions on how to get the local instance running were incomplete.

Also, to answer your questions:
1. Sensitive information is not exposed to non-clearance-holding users (unless we decide that names and IDs of ProgramFamily objects are sensitive, e.g. "secret" programs?)
2. I believe we are implementing this to improve user experience, so that they don't select judging rounds that they can't work with, and don't have a menu full of judging rounds they aren't concerned with.

web/impact/impact/tests/test_judging_round_list_view.py
```diff
@@ -58,7 +62,10 @@ def test_options_against_get(self):
     def test_get_is_active(self):
         is_active = JudgingRoundFactory.create(is_active=True)
         is_not_active = JudgingRoundFactory.create(is_active=False)
```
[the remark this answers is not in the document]
pbx: I agree, and will update this. I probably would push back against changing existing code if I weren't making other updates to this test file right now though.

web/impact/impact/tests/test_judging_round_list_view.py
pbx: Even though I like the way the test case reads as a single method with "before and after", breaking it up as you suggest is the right thing to do. I'll do that.

VincentHokie:
Ah, I see. Thanks. So no damage can be done if a wrong Judging Round is selected 👍

VincentHokie:
Thanks, just waiting on travis now.

VincentHokie:
It does the thing!
Was able to test successfully after creating a `.env` file with the impact url and `yarn; yarn clean`ing.

VincentHokie (approved):

PR #224, opened 2018-10-12
AC-6074: Entrants should not be able to see mentors on `/directory/`

VincentHokie:
**Changes introduced in [AC-6074](https://masschallenge.atlassian.net/browse/AC-6074)**:
- introduce the `AC-6023-fix` that ensures we correctly assume the alumni program is active
- add Directory permissions that stop entrants from accessing the directory

**How to test**:
- Ensure impact-api and the directory are both on `AC-6074`
- register as an entrepreneur on `accelerate`
- visit the directory with that user
- you should get a 403

*Previously entrants were able to access the directory


VincentHokie:
NOTE: For this build to pass, it will need logic in [this PR](https://github.com/masschallenge/django-accelerator/pull/105)

jonkiparsky (changes requested):
Please eliminate references to `is_staff` in algolia_api_key_view.py - behavior should not change based on that flag. (see @EcoLek for more specs if needed)

VincentHokie:
Tis fixed @jonkiparsky , @EcoLek .

web/impact/impact/tests/test_algolia_api_key_view.py
```diff
@@ -53,10 +81,24 @@ def test_unauthenticated_user_is_denied(self):
...
+            with self.login(email=user.email):
+                response = self.client.get(self.url)
+                self.assertTrue(response.status_code, 403)
+
+    def test_user_with_staff_role_grant_sees_all_mentors(self):
         user = self.basic_user()
-        user.is_staff = True
-        user.save()
+
+        named_group = NamedGroupFactory()
```
jonkiparsky: I see this (96-100) repeated a few times, which suggests that it should be a function.

jonkiparsky: The fact that you're adding the program with an explicit `mentor_program_group` suggests that the `mentor_program_group` is important, but it's not clear to me how since it doesn't show up in your code changes. Can you clarify? What happens if it's not there, and is there a test about that?

web/impact/impact/permissions.py
```diff
@@ -73,6 +77,19 @@ def has_permission(self, request, view):
             name=settings.V0_API_GROUP).exists()
 
 
+class DirectoryAccessPermissions(BasePermission):
```
jonkiparsky: It has been an expectation in our code that each class should go in a file named for the class - by that standard, this should be in a file called `directory_access_permissions.py`. However, I see that that standard has been allowed to slip. I also see that there is an AllocateApplicationsPermissions class in v1/classes/. It seems like we should probably make some consistent treatment of these things. 

It seems that the right thing to do here would be to make permissions.py into a directory and extract the random classes into correctly-named files, then collect AllocateApplicationsPermissions into that directory.

web/impact/impact/tests/test_algolia_api_key_view.py
```diff
@@ -53,10 +81,24 @@ def test_unauthenticated_user_is_denied(self):
                 response_data['detail'] == 'Authentication credentials '
                                            'were not provided.')
 
-    def test_staff_user_gets_empty_filters(self):
+    def test_entrepreneur_that_never_had_a_finalist_role_gets_403(self):
```
jonkiparsky, line 95: how is this different from the test at line 38?

web/impact/impact/tests/test_algolia_api_key_view.py
```diff
@@ -128,7 +174,7 @@ def test_alumni_user_only_sees_mentors_of_alumni_programs(self):
 
                 for program in programs:
                     self.assertNotIn(program.name, response_data["filters"])
-                self.assertIn(other_program.name, response_data["filters"])
+                self.assertIn(alumni_program.name, response_data["filters"])
```
jonkiparsky, line 181: The name of this test is `test_user_with_staff_role_grant_sees_all_mentors`. I'm not sure I see how this relates to that name. It's a very confusing test and I'm not sure in fact what it's meant to be testing at all. It seems to create a lot of supporting objects, but very little of that is used in the asserts. 
Can it be cleaned up so the next person has a chance of understanding what's going on here?

web/impact/impact/tests/test_algolia_api_key_view.py
```diff
@@ -159,32 +205,32 @@ def test_alumni_user_who_is_also_finalist_sees_mentors_of_both_programs(
...
-                              response_data["filters"])
+                self.assertEqual(response_data["filters"], [])
 
     def _create_entrepreneur(self):
         ent_user = EntrepreneurFactory()
         ent_user.set_password("password")
         ent_user.save()
         return ent_user
 
+    def _create_expert(self):
```
jonkiparsky: It's not clear why this is an instance method, since it does not use `self`.

This duplicates the previous function, differing only in the choice of factory. Can refactor as 
```
def _create_expert():
  return _create_user(ExpertFactory)

def _create_entrepreneur():
  return _create_user(EntrepreneurFactory)

def _create_user(factory):
  user = factory()
  user.set_password("password")
  user.save()
  return user
```

web/impact/impact/views/algolia_api_key_view.py
```diff
@@ -58,20 +62,22 @@ def get(self, request, format=None):
...
 
 def _get_filters(request):
-    if request.user.is_staff:
+
+    # an empty filter i.e. [], means the user sees all mentors
+    if is_employee(request.user):
         return []
     participant_roles = [UserRole.AIR, UserRole.STAFF, UserRole.MENTOR]
 
-    participant_roles = _entrepreneur_specific_alumni_filter(
```
jonkiparsky: this change means we're beating up on the `participant_roles` list in the function and implicitly changing its state in a way that's not obvious at the point of call. This seems to me like a loss of clarity. If we want to take this route, we should make the statefulness explicit by making it a field of the object, and make the filter functions into instance methods.

jonkiparsky (changes requested):

web/impact/impact/views/algolia_api_key_view.py
VincentHokie: I understand your point, I'll revert.

web/impact/impact/tests/test_algolia_api_key_view.py
VincentHokie: Uuuu nice, very clean. I like it!

web/impact/impact/tests/test_algolia_api_key_view.py
VincentHokie: This one tests an Entrant user i.e. they do not have a RoleGrant to anything, the other tests for an Entrepreneur whose RoleGrants are all for ended programs. I was testing both use cases of this condition in the `DirectoryAccessPermissions` class

https://github.com/masschallenge/impact-api/blob/95f528908b7c4775f517eac2a0dc2a3e20b93802/web/impact/impact/permissions.py#L87

web/impact/impact/tests/test_algolia_api_key_view.py
VincentHokie: The test is `test_alumni_user_only_sees_mentors_of_alumni_programs`, and we create a bunch of ended programs, give the user role grants to one of them, as well as to an active alumni program. Making sure they only see alumni things.

web/impact/impact/tests/test_algolia_api_key_view.py
jonkiparsky: Okay, but they're identical. Not sure that we need both.  :)

```
    def test_logged_in_user_with_no_role_grant_gets_403(self):
        user = self._create_entrepreneur()
        with self.settings(
                ALGOLIA_APPLICATION_ID='test',
                ALGOLIA_API_KEY='test'):
            with self.login(email=user.email):
                response = self.client.get(self.url)
                self.assertTrue(response.status_code, 403)

def test_entrepreneur_that_never_had_a_finalist_role_gets_403(self):
        user = self._create_entrepreneur()
        with self.settings(
                ALGOLIA_APPLICATION_ID='test',
                ALGOLIA_API_KEY='test'):
            with self.login(email=user.email):
                response = self.client.get(self.url)
                self.assertTrue(response.status_code, 403)
```

web/impact/impact/tests/test_algolia_api_key_view.py
VincentHokie: there's this [logic](https://github.com/masschallenge/impact-api/blob/95f528908b7c4775f517eac2a0dc2a3e20b93802/web/impact/impact/views/algolia_api_key_view.py#L88) where we do this

```
program_groups = Program.objects.filter(
        programrole__in=user_program_roles_as_participant).values_list(
        'mentor_program_group', flat=True).distinct()
    program_families = ProgramFamily.objects.filter(
        programs__mentor_program_group__in=program_groups).prefetch_related(
        'programs').distinct()
    facet_filters = _facet_filters(program_families)
```

I've been meaning to get some context from Shanks about `_facet_filters` but it looks like `mentor_program_group` is used in a grander idea to get the most recent program of a program family to grant the user access to that programs mentors.

web/impact/impact/tests/test_algolia_api_key_view.py
VincentHokie: ah, those two..I was comparing the wrong two. I'll get rid of the first. Thanks.

web/impact/impact/permissions.py
VincentHokie: It looks like the class you mean here `v1/classes/` is inheriting from the `permissions.py`, which leads me to think it may be intentional. No?

jonkiparsky: well, yeah, it's inheriting from permissions.py - but probably should be with the rest of the permissions. 
I'm not going to insist on this cleanup now, but in general it's a lot better to clean up messes as you find them than to try to schedule tickets.

web/impact/impact/tests/test_algolia_api_key_view.py
jonkiparsky: okay, still seems kind of messy but I'll let it go

web/impact/impact/tests/test_algolia_api_key_view.py
VincentHokie: Sorted this one, by moving to a function i.e. `_create_batch_program_and_named_group`

web/impact/impact/permissions.py
VincentHokie: Doing it now.

VincentHokie: Tis done, permissions cleaned up on the directory!

VincentHokie: on **impact!

jonkiparsky (approved):

PR #225, opened 2018-10-16
[AC-6065] allocate_applications_view fails if user quota is None

pbx:
## [AC-6065](https://masschallenge.atlassian.net/browse/AC-6065)

Apologies for the mint-free PR. I plead 9:40PM.


jonkiparsky (changes requested):
Sorry, probably my bad for not getting it into the title, but the Done also calls for handling capacity of None (which is also a thing that happens, and will also fail)
I don't see anything else that I could object to here.

pbx:
@jonkiparsky - see my comments on the ticket from last night. `capacity` is a non-nullable field. If the schema needs to change, yeah, that definitely needs to be on the ticket.  Also repro steps for "thing that happens". Thanks.

jonkiparsky (approved):

PR #234, opened 2018-11-14
AC-6171: Migration 0009 fails if criteria already exist

VincentHokie:
**Changes introduced in [AC-6171](https://masschallenge.atlassian.net/browse/AC-6171)**:
- explicitly use the `mc_dev` database in case none is specified e.g. when using `load-remote-db`

**How to test**
- run `make load-remote-db`
- your data should have been successfully loaded in the local

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -301,9 +301,13 @@ shutdown-vms: .env
 
 REMOVE_IMAGES = no
 remove_images ?= $(REMOVE_IMAGES)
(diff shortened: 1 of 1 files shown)


jonkiparsky (approved):

jonkiparsky:
Note: this does NOT address any concerns with migration 0009 itself. However, it's likely that this fixes the underlying issue which surfaced those issues, so I'm willing to go ahead and merge it.

PR #244, opened 2019-01-04
AC-6277: make run-all-servers does not start directory

VincentHokie:
**Changes introduced in [AC-6277](urlhttps://masschallenge.atlassian.net/browse/AC-6277)**:
- create make targets for starting and stopping our frontend
- expose our frontend to port 1234 automatically

**How to test**:
1 - run `make stop-all-servers`
2 - run `docker ps -a | grep mentor_directory` and get the container_id of `impact-api_mentor_directory` and do `docker rm <id>`
3 - run `docker images | grep mentor_directory` and do `docker rmi <image-id>`

- now checkout AC-6277 on both impact and the directory
- while on impact, run `make run-server`
- wait for the text "FRONTEND BUILD COMPLETE - visit http://localhost:1234 in a browser"
- a this point the directory should be available on localhost:1234 and should reload on code change

@masschallenge/dev-team please note that when the AC-6277 PRs are merged, we will all need to do steps 1 - 3 to ensure you're using these changes.

Steps 1 - 3 can also be used to switch context (i.e. reverse the changes of this PR) after testing this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -265,19 +265,31 @@ checkout:
 	  echo; \
 	  done
 
+watch-frontend: process-exists=$(shell ps -ef | grep "yarn.js watch" | grep -v "grep" | awk '{print $$2}')
+watch-frontend:
+	@if [ -z "$(process-exists)" ]; then \
+		cd $(DIRECTORY) && nohup bash -c "yarn watch &" && cd $(IMPACT_API); \
+	fi;
+
+stop-frontend: process-exists=$(shell ps -ef | grep 'yarn.js watch' | grep -v 'grep' | awk '{print $$2}')
+stop-frontend:
+	@if [ -n "$(process-exists)" ]; then \
+		kill $(process-exists); \
+	fi;
+
 # Server and Virtual Machine related targets
 debug ?= 1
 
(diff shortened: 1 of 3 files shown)


mkiterian:
QA Notes:

- [x] Make run-all-servers starts the directory
- [x] Changes on the directory are reflected without having to restart

pbx (dismissed):
The basic functionality is definitely there (i.e. the `run-server` output is as described and the directory server is responsive). 

However the console output is a bit misleading. Specifically the impact-api web server (port 8000) was not responsive at the time the "WEB BUILD COMPLETE" message was printed.

Also, while the directory did start up, the page displayed "Oh no! <script> unable to load due to an `error` event on it" -- is this expected?

VincentHokie:
That language can be improved, on it.

Definitely not expected. When you run the directory as you usually do, do you get this error? Any chance there's a local change on the directory causing this?

jonkiparsky (approved):
This worked for me, after a few tries, and the code looks reasonable. I'm happy to go forward with it.

mkiterian (approved):

PR #246, opened 2019-01-11
[AC-6262] Address pathological performance of AllocateApplicationsView when criteria are engaged

jonkiparsky:
https://masschallenge.atlassian.net/browse/AC-6262

To review: 
- on clean db, find an online JudgingRound and modify it so that 
 -- round is active
 -- end date is in the future
 -- allow_dynamic_allocation is True

- count the number of active Applications for that round: `Application.objects.filter(cycle=judging_round.program.cycle, application_status="submitted").count()`
- find a Criterion whose `type` is industry, set its judging_round_id to point to your chosen JudgingRound

- Find a confirmed judge for that round, delete all of their JudgePanelAssignments, and become that user

hit the allocate applications endpoint on impact:
http://localhost:8000/api/v1/allocate_applications/[JUDGING ROUND ID]/[USER ID]/

On development branch, this will take a while. On the feature branch, this will be much faster

Note the SQL tab in the debug toolbar. On the development branch, you will see (among others) 5 queries for each Application - one query for each top-level industry. On the feature branch, those will be gone. 

You can also test for the other criteria using a similar technique. I found it useful to reduce the number of active applications:
```
ids = list(Application.objects.filter(cycle=judging_round.program.cycle, application_status="submitted").values_list('id', flat=True))
Application.objects.filter(cycle=judging_round.program.cycle, application_status="submitted", id__in=ids[50:]).update(application_status="discarded")
```
This will ensure that you're down to 50 applications, which should be more manageable. 

The feature branch should load with reasonable performance and db usage with all criteria engaged for your judging round. ("reasonable db usage": there are still some duplicated queries, but no N+1 performance)

Note: for repeated testing, you can clear out existing assignments like so:
```
panel = JudgePanelAssignment.objects.filter(judge_id=YOUR_JUDGE_ID).panel
panel.applicationpanelassignment_set.all().delete()
panel.judgepanelassignment_set.all().delete()
```

### DIFF
--- a/web/impact/impact/tests/test_criterion_helper.py
+++ b/web/impact/impact/tests/test_criterion_helper.py
@@ -2,9 +2,10 @@ from impact.tests.api_test_case import APITestCase
 from impact.tests.factories import CriterionOptionSpecFactory
 from impact.v1.helpers import CriterionHelper
 
+
 class TestCriterionHelper(APITestCase):
     def test_find_helper_caches_by_criterion(self):
-        
+
         criterion = CriterionOptionSpecFactory().criterion
         helper_instance = CriterionHelper.find_helper(criterion)
         assert CriterionHelper.find_helper(criterion) == helper_instance
--- a/web/impact/impact/tests/test_criterion_helper.py
+++ b/web/impact/impact/tests/test_criterion_helper.py
@@ -1,0 +1,10 @@
+from impact.tests.api_test_case import APITestCase
+from impact.tests.factories import CriterionOptionSpecFactory
+from impact.v1.helpers import CriterionHelper
+
+class TestCriterionHelper(APITestCase):
+    def test_find_helper_caches_by_criterion(self):
+        
+        criterion = CriterionOptionSpecFactory().criterion
+        helper_instance = CriterionHelper.find_helper(criterion)
+        assert CriterionHelper.find_helper(criterion) == helper_instance
--- a/web/impact/impact/v1/helpers/criterion_helper.py
+++ b/web/impact/impact/v1/helpers/criterion_helper.py
@@ -40,12 +40,16 @@ class CriterionHelper(ModelHelper):
 
     @classmethod
     def find_helper(cls, criterion):
-        helper = cls.helper_instances.get(criterion.name)
+        helper = cls.helper_instances.get((criterion.name))
         if helper is None:
             helper = cls.specific_helpers.get((criterion.type, criterion.name),
                                               cls)(criterion)
-            cls.helper_instances[criterion.name] = helper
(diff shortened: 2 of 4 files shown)


kitaka:
@jonkiparsky I have needed to limit the number of applications to 10 to get a baseline on development. However, whereas the changes in the branch reduce the number of queries by nearly 50%, the amount of time taken to execute all the queries seems to remain the same. Is this expected behaviour?

jonkiparsky:
@kitaka You should be seeing substantial improvements in the time required. This might be obscured by the fact that you've cut the application set down to 10. Even a slightly higher number - 30, say - might show the difference more strongly. 
Also, make sure you just have one criterion (Industry, for example) engaged.

jonkiparsky:
Closing this PR in favor of a ticket to do this work in a better way.

jonkiparsky:
Re-opening per @shanbady 
The plan is to merge this for use in the upcoming round (which starts 24 January) and to address the possible collisions (see follow-on ticket AC-6342) at our leisure.

VincentHokie:
A LOT better performance on my first test!
For a judging round with all criteria activated and 50 applications.
Queries went from 1003 -> 58
Time went from 4529ms to 363ms

VincentHokie:
Tried
- clearing all pyc files and
- restarting servers to try and see if caching from my development tests was causing better performance on the branch.

and found
- Query count remained the same as previous test
- Time trippled to 1016ms but still A LOT better than the 4529ms from the development branch
- Subsequent allocate calls went down to ~600ms

VincentHokie:
As a side-note, what worked when re-allocating to the same judge while testing was the following

    In [1]: from mc.models import *
    In [2]: jr = JudgingRound.objects.filter(id=<judging-round-id>)[0]
    In [3]: scenarios = Scenario.objects.filter(judging_round=jr, is_active=True)
    In [4]: jpa = JudgeApplicationFeedback.objects.filter(panel__judgepanelassignment__scenario__in=scenarios).all()
    In [5]: judgeOur = jpa[6].judge
    In [6]: judgeOur.judgepanelassignment_set.filter(scenario__judging_round=jr, assignment_status='ASSIGNED').delete()

VincentHokie (approved):
I'm confident in these changes.

The follow-on - [AC-6342](https://masschallenge.atlassian.net/browse/AC-6342) - will address my concerns about thread-safety as well as (almost simultaneous) allocations across judging rounds affecting each other.

kitaka:
@jonkiparsky Initially I began my tests on development and it kept timing out and that led me to reducing the number of applications. From further testing starting with this feature branch, the api appears to respond very fast while development still times out.

kitaka (approved):

PR #247, opened 2019-01-14
AC-5981: Resolve repeated queries stemming from JudgeCriteria lookups

VincentHokie:
#### Changes introduced in [AC-5981](https://masschallenge.atlassian.net/browse/AC-5981):
- remove repeated queries
- use new logic designed to get aggregates for analysis

#### How to test:

- check out AC-5981 on `impact` and `django-accelerator`
- ensure all your servers are running
- Download [this dump](https://s3.console.aws.amazon.com/s3/object/db-dump-sante-test/analysis-dump-2019-01-15.sql.gz?region=us-east-1&tab=overview&prefixSearch=ana) from S3 that has a judging round's commitments all made INCOMPLETE
- load it locally with `make load-db gz_file=<path-to-gz-dump>`

##### To validate the analysis data
- run `make django-shell` in accelerate and do the following
  - `from mc.models import *`
  - `d = User.objects.get(email__startswith="ahokie@")` (alternatively, use your email address here and give yourself the clearance needed)
  - `Clearance.objects.create(user=d, level=CLEARANCE_LEVEL_GLOBAL_MANAGER, program_family=ProgramFamily.objects.first())` (if this fails, do the following)
  - `Clearance.objects.filter(user=d, program_family_id=1).update(level=CLEARANCE_LEVEL_GLOBAL_MANAGER)`
  - `d.is_staff = d.is_superuser = True`
  - `d.save()`
  - `scenarios = Scenario.objects.filter(judging_round=48, is_active=True)`
  - `jpa = JudgeApplicationFeedback.objects.filter(panel__judgepanelassignment__scenario__in=scenarios).all()`
- at this point you can start doing interesting things e.g.
  - `first_ten = jpa[0:10]`
  - `for item in first_ten:`
  - `item.feedback_status = 'COMPLETE'`
  - `item.save()`

- if you gave `ahokie` clearances, you will have to mask as him i.e. http://localhost:8181/admin/simpleuser/user/60464/masquerade/, otherwise log in as the user you gave clearances to
- then go to the allocator at `localhost:1234/allocator` and select `2017-04 BOS CH IL Round 1` from the judging round dropdown in the top left
- notice that we have 10 reads among other analyses displayed in the gender, role, program and industry sections.
- to undo the reads, you could run the above loop and set `feedback_status` to `INCOMPLETE`

##### To see duplicate query comparisons
- from `web/impact/impact/v1/views/analyze_judging_round_view.py` do the following
  - put the following import at the top `from django.db import connection`
  - then put this in the get method before returning the response --> `print(connection.queries)`
  - this will print all the queries executed in the console, which can then be moved to a file for analysis
  - I did all this and expect we'll all have similar results. The results are the following which you can download. [Queries](https://s3.amazonaws.com/db-dump-sante-test/AC-5981/old_queries.json) on development, [queries](https://s3.amazonaws.com/db-dump-sante-test/AC-5981/new_queries.json) on AC-5981.
- to see how many times each query is executed, run
  - `sort <path-to-file> | uniq -c > [old/new]_query_count.json`
- to see the queries that have been repeated, run
  - `sort <path-to-file> | uniq -c | grep -v '^ *1 ' > [old/new]_duplicate_query_count.json`
- comparing these files will show you that we have reduced the number of duplicated queries

##### Alternate approach to review:
- check out development and hit http://localhost:8000/api/v1/analyze_judging_round/48/ 
- check out this branch, open a new tab, and hit that url again
- compare the json
- compare the queries in the django debug toolbar (toolbar does not currently show the actual query, but it does show what tables the query is hitting, what it's retrieving, and where it's called, which is generally sufficient to understand what's happening)

#### Note:
- As a bonus, we also have gender statistics now.
- the numbers that come from the old and new logic defer in the industry analyses but the results on this branch have been confirmed as correct.


jonkiparsky:
Side note: In general prefer `qs[:10].update(field=val)` to 
```
for item in qs[:10]:
  item.field=val
  item.save()
```

See https://docs.djangoproject.com/en/dev/topics/db/queries/#limiting-querysets for slicing querysets and db performance

VincentHokie:
thanks! definitely handy knowledge to have

jonkiparsky:
Note also that when you hit the http://localhost:8000/api/v1/analyze_judging_round/[ROUND_ID] link, the django debug toolbar gives you an SQL tab which identifies and groups similar queries. Due to recent update breakage, the queries themselves are not available in the toolbar, but you can see the line of code that triggered the query, and the explain tab will show you information about what table(s) were hit. Obviously the `connection.queries` information is useful in nailing down the actual form of the query.

VincentHokie:
My toolbar actually shows me the queries, maybe I don't have the latest update. But the reason I took the `connection.queries` approach is to get confirmation about what queries are being duplicated. My suspicions were actually confirmed, the toolbar showed me queries that were apparently duplicated but the `connection.queries` didn't confirm this.

web/impact/impact/v1/helpers/criterion_helper.py
```diff
@@ -32,15 +32,30 @@ class CriterionHelper(ModelHelper):
...
+        if cls.helper_instances.get(criteria_tuple) is None:
+            cls.helper_instances[criteria_tuple] = cls.specific_helpers.get(
+                criteria_tuple,
+                cls
+            )(criterion)
+        return cls.helper_instances[criteria_tuple]
+
+    @classmethod
+    def clear_cache(cls):
+        cls.helper_instances = {}
```
shanbady: does self. app_count_cache need to be reset here?

shanbady (dismissed):
@VincentHokie - im running into permissions issues on the db dump referenced in the testing instructions. (also be aware that this repo is not private like the others)

VincentHokie:
Sorry, I didn't see your comment Shankar..added instructions for adding clearance permissions

VincentHokie:
Going through logic to make sure getting development changes has not broken anything

web/impact/impact/v1/helpers/criterion_helper.py
VincentHokie: this cache is at the instance level, so it get's destroyed at the end of the request by Djangos garbage collector

web/impact/impact/v1/classes/option_analysis.py
```diff
@@ -96,38 +139,161 @@ def calc_needs(self, option_name):
...
-                                                    applications=self.apps)
-        app_counts = Counter(app_ids)
-        counts = Counter(app_counts.values())
-        unread_count = (self.helper.app_count(self.apps, option_name) -
-                        len(app_counts))
+        app_counts = self.application_criteria_read_state(
+            self.completed_feedbacks,
+            option_name=option_name,
+            applications=self.apps)
+        counts = {}
```
jonkiparsky: Looks like this wants to be a `defaultdict(int)`

web/impact/impact/v1/classes/option_analysis.py
```diff
@@ -96,38 +139,161 @@ def calc_needs(self, option_name):
...
+        for count in app_counts.values():
+            total = count[self.option_spec.criterion.name].get(option_name, 0)
```
jonkiparsky: @VincentHokie I'm getting a fail at this line, `'int' object has no attribute 'get'`

Possibly an unpushed commit?

VincentHokie: I tested this against a JR without reads as a criteria, it's fixed now

web/impact/impact/v1/classes/option_analysis.py
VincentHokie: just read about it, definitely what we need for this, updating now

VincentHokie:
@kaymation , I thought this was a problem with my data. Some of the objects won't get updated *you could cross-check). I didn't give much focus to that, knowing that one object didn't update still made the calculations sane.

web/impact/impact/v1/classes/option_analysis.py
```diff
@@ -24,38 +25,81 @@
...
-            criterion_helpers)
+        self.cycle = self.judging_round.program.cycle
+        self.helper = None
         self.apps = apps
         self.app_ids = app_ids
         self.application_counts = application_counts
+        self.criterion_total_capacities = {}
+        self.judge_to_capacity_cache = None
+        self.criterion_total_functions = {
+            ("judge", "gender"): {
```
jonkiparsky: Information specific to criteria and options should live on the appropriate helpers.

web/impact/impact/v1/classes/option_analysis.py
```diff
@@ -96,38 +140,164 @@ def calc_needs(self, option_name):
...
+        key_exists = self.criterion_total_capacities[option_name].get(option)
+        return 0 if not key_exists else self.criterion_total_capacities[
+            option_name][option]
+
+    def remaining_capacity(self, assignment_counts, option_spec, option):
+        if self.judge_to_capacity_cache is None:
+            self.judge_to_capacity_cache = JudgeRoundCommitment.objects.filter(
+                judging_round=self.judging_round).values(
+                    "judge_id",
+                    self.industry_judge_field,
```
jonkiparsky: Again, OptionAnalysis should get criterion-specific information from the criterion helpers. Note that this will make this query more precisely targeted to the criteria in force for the judging round - if, for example, we're analyzing a judging round that doesn't care about industry, we would not be getting the values for industry.

web/impact/impact/v1/classes/option_analysis.py
```diff
@@ -96,38 +140,164 @@ def calc_needs(self, option_name):
...
             "remaining_capacity": remaining_capacity,
         }
 
+    def general_criterion_total_capacity(self, option):
+        option_name = self.option_spec.criterion.name
+        field = self.criterion_total_functions[
+            (self.option_spec.criterion.type, option_name)
+        ]['filter_field']
+
+        if option_name == "gender":
```
jonkiparsky: logic about specific criteria belongs in the appropriate criterion helpers

web/impact/impact/v1/classes/option_analysis.py
```diff
@@ -96,38 +140,164 @@ def calc_needs(self, option_name):
...
+            self.judge_to_capacity_cache = JudgeRoundCommitment.objects.filter(
+                judging_round=self.judging_round).values(
+                    "judge_id",
+                    self.industry_judge_field,
+                    self.program_judge_field,
+                    self.gender_judge_field,
+                    self.role_judge_field,
+                    "capacity"
+                ).annotate(
+                    industry=F(self.industry_judge_field),
```
jonkiparsky: These can come from criterion helpers. In this case (and in the similar case below) we'd do this by getting a dict of `{field_name: field}` from each criterion helper and updating a kwargs dict, which we pass to the `annotate` function. Something like
```
fields = ["judge_id"] + [helper.field() for helper in self.criterion_helpers]
annotations = {criterion_helper.annotation() for criterion in self.criterion_helpers}
self.judge_to_capacity_cache = JudgeRoundCommitment.objects.filter(
    judging_round=self.judging_round).values(*fields).
       annotate(**annotations)
```

web/impact/impact/v1/classes/option_analysis.py
```diff
@@ -24,38 +25,81 @@
...
+
+        active_scenarios = Scenario.objects.filter(
+            judging_round=judging_round, is_active=True)
+        self.jpa = JudgeApplicationFeedback.objects.filter(
+            application__in=apps,
+            panel__judgepanelassignment__scenario__in=active_scenarios)
+        self.completed_feedbacks = self.jpa.filter(
+            feedback_status=JUDGING_FEEDBACK_STATUS_COMPLETE)
+        self.application_criteria_read_state_cache = {}
+        self.gender_match_dict = dict(
```
jonkiparsky: This seems to be here because some criteria were set up wrong and have options "female" and "male" instead of "f" and "m". We should just use the db values ("f" and "m") and fix the erroneous criteria options rather than writing the mistake into the code.

jonkiparsky (changes requested):
See various comments (I didn't realize this would turn into a full-fledged review!)

web/impact/impact/v1/classes/option_analysis.py
```diff
@@ -144,8 +144,12 @@ def calc_needs_distribution(self, option_name):
             option_name=option_name,
             applications=self.apps)
         counts = {}
+        criterion_name = self.option_spec.criterion.name
         for count in app_counts.values():
-            total = count[self.option_spec.criterion.name].get(option_name, 0)
+            total = (
+                count[criterion_name].get(option_name, 0)
+                if criterion_name != "reads" else count[criterion_name]
```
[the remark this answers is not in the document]
VincentHokie: ~ugh~.

Got it, will fix.

web/impact/impact/v1/classes/option_analysis.py
```diff
@@ -96,38 +140,164 @@ def calc_needs(self, option_name):
...
+                judge[option_name] == gender_option
+            ):
+                result += max(0, judge['capacity'] - assignment_counts.get(
+                    judge['judge_id'], 0))
+        return result
+
+    def application_criteria_read_state(self, feedbacks, option_name):
+        criterion_name = self.option_spec.criterion.name
+
+        if not self.application_criteria_read_state_cache:
```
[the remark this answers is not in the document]
VincentHokie: this whole function is the object-caching step

VincentHokie:
I'm not a huge fan of merge commits, so I cherry picked your cleanup commits and rebased with dev.

VincentHokie:
> It looks like the issues with the industry and program may be due to you collecting all of the reads that a startup has got in that criterion, when we should only be collecting the reads relevant to that particular startup. (ie, if startup is in Health Care, we don't care how many reads they got by High Tech judges)

Handeld this. Now I'm working on the failing tests.

VincentHokie:
Also, I've handled most of the "move logic to helper" concerns. A bit more could be done, but it's worth another pass.

VincentHokie:
Also, @jonkiparsky , please check out the failing errors. They seem to do with allocation but I'm not sure what change could have caused this failure. Will probably have a look with fresh eyes in a couple of hours.

web/impact/impact/tests/test_analyze_judging_round_view.py
```diff
@@ -96,7 +100,7 @@ def test_get_with_unread_application(self):
                     context.needed_reads())
 
     def test_get_with_gender_criterion(self):
-        genders = ["f", "m"]
+        genders = ["female", "male"]
```
jonkiparsky: Huh. This should fail, since we never store "female" or "male" in the db.

VincentHokie: ~this actually creates female and male criteria objects in the db through the AnalyzeJudgingContext which is why it still works~

jonkiparsky: Looked into this. It actually makes sense: if we specify bad criteria, it just doesn't register any reads for them which is what the test ultimately asserts. 
Restoring these to the expected values, leaving it alone otherwise.

jonkiparsky (approved):

VincentHokie:
Before this is merged, ensure [this PR](https://github.com/masschallenge/django-accelerator/pull/129) is merged. This one depends on logic from django-accelerator

kaymation (approved):
Qa'ing locally using method 2. It went from 83 to 82 queries. 
I'm not sure if this is sucfficient, so i will approve to be unblocking, but raise this as a concern

pbx:
I gave this a go, but even after loading the provided database dump, logged in as `demoadmin` (with all  the user settings above e.g. `CLEARANCE_LEVEL_GLOBAL_MANAGER`) I see no choices in the Online Judging Rounds menu when I load http://localhost:1234/allocator

It's possible that somewhere in my build process (which was in several stages due to interruptions) I ended up with the default `initial_schema`, but in any case I don't have a setup that will let me test per the instructions.  

Also, re "print all the queries executed in the console, which can then be moved to a file for analysis" -- I wasn't clear what actions on my part were supposed to be generating the output. Perhaps if I got farther than the empty Online Judging Allocation menu this would have become clear.

A lighter-weight setup technique than "load this DB dump" would help too.

Sorry not to have been able to provide useful feedback yet.

PR #248, opened 2019-01-16
[AC-6217] Allocation analysis should not break if option_spec has `option` field set

jonkiparsky:
https://masschallenge.atlassian.net/browse/AC-6217

The deleted lines from `OptionAnalysis.find_options` lines short-circuit matching criterion lookup if the option_spec has an option value set, causing dict lookup failures.

Note that the deleted lines are the default behavior for `CriteriaHelper.options`. The MatchingCriteriaHelpers override this behavior, and the ReadsCriterion and the various JudgeCriteria do not override it, so we get this behavior through the Helper method. 

To test:
on clean db, Boston 2017 Round 1 has criteria set up. in the shell, the following command will create the state that should fail on development and succeed on the feature branch:
```
mc.CriterionOptionSpec.objects.filter(criterion__type="matching").update(option="asdfasdf")
```

No other behavior should change.


jonkiparsky:
Discovered another caching-on-the-class issue while writing a test for this. Closing while I look into that.

jonkiparsky:
Addressed caching-on-the-class, reopening

web/impact/impact/v1/helpers/matching_industry_criterion_helper.py
```diff
@@ -18,14 +18,13 @@ def __init__(self, subject):
         super().__init__(subject)
         self._top_level_id_cache = None
 
-    @classmethod
-    def _industry_map(cls):
-        if cls.industries is None:
-            cls.industries = cls.instances_by_name(Industry)
-        return cls.industries
+    def _industry_map(self):
+        if self.industries is None:
```
shanbady, line 22: what is the use of this method if its now an instance method? the _industry_map method seems to only be referenced once so i dont think it needs to be cached.

shanbady (changes requested):

web/impact/impact/v1/helpers/matching_industry_criterion_helper.py
jonkiparsky: This instance is persistent, so I suspect it's hit multiple times in the course of a given analysis. (once for each of the top-level industries)

shanbady (changes requested):

web/impact/impact/v1/helpers/matching_industry_criterion_helper.py
shanbady: we can set self.industry_map on the __init__ method and getting rid of this method

VincentHokie:
Just to be clear, a couple of things on the ticket don't hold anymore i.e. you can create a criterion option with the "Option" left empty from the admin panel.

Also, I haven't been able to reproduce this with the data I have locally (not from `load-db`).

Doesn't that suggest that it's a state issue rather than a bug that requires code change?

jonkiparsky:
@VincentHokie Repro steps are included in the testing notes. It's a straightforward matter to see that this is a code issue. 
Also note that the tests fail when run on development, but pass on the branch.

web/impact/impact/v1/helpers/matching_industry_criterion_helper.py
jonkiparsky: Per our standards, we avoid db activity in the __init__

web/impact/impact/tests/test_analyze_judging_round_view.py
```diff
@@ -125,6 +125,16 @@ def test_get_with_industry_criterion(self):
         dists = _calc_industry_dists(options, judge_key)
         self.assert_option_distributions(context, dists)
 
+    def test_option_spec_option_field_can_be_nonempty(self):
```
VincentHokie: if im using the other test for reference, this should probably be `test_industry_option_field_can_be_nonempty`

web/impact/impact/v1/helpers/matching_industry_criterion_helper.py
VincentHokie: What I wonder here is how this code change relates to the solution itself, I understand everything else but this as far as the ticket. I believe it's fine as it was.

jonkiparsky: @VincentHokie 
> Discovered another caching-on-the-class issue while writing a test for this. Closing while I look into that.

As written this code made it impossible to write more than one test for any given matching criterion.

VincentHokie: ah! because it was stored as a class variable and not on the object the cache was available cross-test..I see

VincentHokie (changes requested):
Waiting on [this](https://github.com/masschallenge/impact-api/pull/248/files#r249171723) for a green light from me

VincentHokie (approved):

shanbady (approved):

PR #249, opened 2019-01-18
AC-6350: Pin mysqlclient dependancy on impact

VincentHokie:
#### Changes introduced in [AC-6350](https://masschallenge.atlassian.net/browse/AC-6350)
- pin mysqlclient dependancy

#### How to test
- my failing impact was fixed by this, and having Paul's env build should be good enough for this PR


pbx (approved):
Success!

PR #256, opened 2019-01-25
[AC-6263] Fix New Relic on impact-api

pbx:
## [AC-6263](https://masschallenge.atlassian.net/browse/AC-6263)

**Summary:** The [New Relic docs for the Python agent](https://docs.newrelic.com/docs/agents/python-agent/installation/python-agent-advanced-integration) note that the `newrelic.agent` import (which happens in our `wsgi.py`) should be first; in accelerate it was, but in impact-api it was not. 

**Testing:** Not feasible locally. My suggestion is that we ship this once we agree that it looks reasonable, and make a microticket for checking the fix: turn off the NR agent on staging, deploy the release containing this change to staging, and confirm that the agent starts and reports as expected.


VincentHokie (approved):
agree with Paul that this is a merge and see type of change, approving so that we can get it tested with the next release

PR #259, opened 2019-02-08
[AC-6226] Remove the job board

mkiterian:
### Changes introduced in [AC-6226](https://masschallenge.atlassian.net/browse/AC-6226):
- Remove Job Post views and associated urls from `/api/v0` and `/api/accelerator`

### How to test
- Attempt to access `/api/job_posting/` or`/api/accelerator/job_posting/`
 from the api should result in a LookupError


web/impact/impact/urls.py
```diff
@@ -35,14 +35,18 @@
...
 accelerator_router = routers.DefaultRouter()
 simpleuser_router = routers.DefaultRouter()
 simpleuser_router.register('User', GeneralViewSet, base_name='User')
 
 for model in apps.get_models('accelerator'):
-    if model._meta.app_label == 'accelerator' and not model._meta.auto_created:
-        schema_router.register(model, url=model_name_to_snake(model.__name__))
+    if (model._meta.app_label == 'accelerator' and not
+            model._meta.auto_created and
+            model.__name__ not in EXCLUDE_MODELS):
```
kitaka: @mkiterian instead of the setting, for consistency, couldn't we just create a meta option that flags whether the model should be excluded?

kitaka:
@mkiterian It would be good to have full URLs in the instructions.

web/impact/impact/views/general_view_set.py
```diff
@@ -22,9 +27,12 @@ def model(self):
         model = snake_to_model_name(self.kwargs.get('model', ''))
         related_model = self.kwargs.get('related_model', '')
         model_name = model_name_case(model, related_model)
-        return apps.get_model(
+        model = apps.get_model(
             app_label=self.kwargs['app'],
             model_name=model_name)
+        if model.__name__ in EXCLUDE_MODELS:
+            raise Http404
```
kitaka: @mkiterian why do we raise a 404? Why not raise a `LookupError` just like the `get_model` method? https://github.com/django/django/blob/402c0caa851e265410fbcaa55318f22d2bf22ee2/django/apps/config.py#L180

Lastly, why do we exclude the model from both the `GeneralViewSet` and the `urls.py`, are the endpoints generated twice?

kitaka (changes requested):

web/impact/impact/views/general_view_set.py
mkiterian: The exclusion in urls.py is to make sure that it's not listed here: `http://localhost:8000/api/accelerator/`

web/impact/impact/urls.py
mkiterian: Thanks for the suggestion @kitaka . I've looked into the [meta options](https://docs.djangoproject.com/en/1.11/ref/models/options/) and there only a predefined set none of which we can use. Custom options cannot not be created [ref](https://code.djangoproject.com/ticket/5793).

kaymation (approved):
Works like the can says

kitaka (approved):

PR #267, opened 2019-02-28
AC-6428: Break our local environment into logical pieces working together

VincentHokie:
#### Changes introduced in [AC-6428](https://masschallenge.atlassian.net/browse/AC-6428)
- create mysql docker-compose file
- remove mysql service from impact docker-compose file
- use v3 in old docker-compose file
- connect old docker-compose file to impact-api_default network
- remove .gitkeep from dir we dont use anymore
- add ensure-mysql and kill-exited-containers target to makefile
- ensure targets run services that they depend on

#### How to test
to do a more targeted test
- while on development
- run `make shutdown--all-vms`
- checkout this branch on impact and accelerate
- run `make run-all-servers`
- now run tests and see improved test times

to do more general testing, this may be a bit tedious but may be the best way to test env changes i.e. using the changes in your daily workflow
- while on development
- run `make shutdown--all-vms`
- checkout this branch on impact and accelerate
- run `git reset --soft HEAD~1; git reset HEAD .` on both impact and accelerate
   - the above will allow you to keep the changes in this branch as you do your daily work
- now when doing your own work, you will have to 
   - `git stash`
   - `git checkout <the-branch-you-want-to>`
   - `git stash pop`
- also you will have to be a lot more selective when adding changes to commit i.e. you can't do `git add .` since this will also add changes from this branch.

Feel free to choose the approach that works best for you

#### Note
This PR and its [sibling](https://github.com/masschallenge/accelerate/pull/2072) are tightly coupled, merge only when both have been approved

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -228,7 +228,6 @@ DIRECTORY = ../directory
 SEMANTIC = ../semantic-ui-theme
 REPOS = $(ACCELERATE) $(DJANGO_ACCELERATOR) $(DIRECTORY) $(SEMANTIC) $(IMPACT_API) 
 
-
 # Database migration related targets
 
 data-migration migrations: .env
@@ -282,16 +281,16 @@ stop-frontend:
 # Server and Virtual Machine related targets
 debug ?= 1
 
-run-server: run-server-$(debug) watch-frontend
+run-server: run-server-$(debug)
 
-run-server-0: .env initial-db-setup watch-frontend
+run-server-0: .env initial-db-setup watch-frontend kill-exited-containers ensure-mysql
 	@docker-compose up
 
-run-detached-server: .env initial-db-setup watch-frontend
+run-detached-server: .env initial-db-setup watch-frontend ensure-mysql kill-exited-containers
 	@docker-compose up -d
 	@docker-compose run --rm web /usr/bin/mysqlwait.sh
 
-run-server-1: run-detached-server watch-frontend
+run-server-1: run-detached-server watch-frontend ensure-mysql kill-exited-containers
 	@docker-compose exec web /bin/bash /usr/bin/start-nodaemon.sh
 
 dev: run-server-0
@@ -307,11 +306,9 @@ initial-db-setup:
 
 stop-server: .env stop-frontend
 	@docker-compose stop
-	-@killall -9 docker-compose || true
 
 shutdown-vms: .env stop-frontend
 	@docker-compose down
-	@rm -rf ./mysql/data
 	@rm -rf ./redis
 
 REMOVE_IMAGES = no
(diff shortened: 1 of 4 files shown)


jonkiparsky:
Per kaizen, I'll take the "shepherd" role and focus on QA.

jonkiparsky (changes requested):
Blocking review to ensure that this PR and its sibling merge simultaneously. This review may be dismissed iff https://github.com/masschallenge/accelerate/pull/2072 is ready to merge.

jonkiparsky (commented):
From the QA side, this seems to be working. If there are any introduced bugs, I haven't been able to find them. Will approve and merge once both PRs have another engineer's approval.

Makefile
```diff
@@ -492,5 +489,16 @@ comp-messages: .env
 messages: .env
 	@docker-compose exec web python manage.py makemessages -a
 
-lint: .env
+lint: .env ensure-mysql
 	@docker-compose run --rm web pylint impact
+
+ensure-mysql: kill-exited-containers
```
pbx, line 496: Seven of the targets naming `ensure-mysql` as a dependency also name `kill-exited-containers` as a dependency; two don't.

This is a little bikesheddy I admit, but in the interest of reducing verbosity and repetition could we add a target for use by the two that don't (`load-remote-db` and `lint`), like `ensure-mysql-but-keep-containers`, so we can drop the explicit `kill-exited-containers` from the other seven?

pbx (changes requested):
Looks good, works locally.  I have one thought about cleaning up the Makefile slightly.

Also, I saw this warning:
> WARNING: Found orphan containers (mc-mysql) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up.

Can we squelch that (if appropriate), or clarify it so the engineer knows whether there's something they need to take action on or not?

Makefile
VincentHokie: good eye!
- the `ensure-mysql` on `load-remote-db` is actually redundant since the target it executes i.e. `load-db` also executes `ensure-mysql`
- `lint` actually also needs to `kill-exited-containers` just to be sure it doesn't restart 'bad' containers.

but thanks for this, removed the `kill-exited-containers` redundancy

VincentHokie:
Unfortunately I have no control over that warning @pbx , its a misinterpretation by docker-compose that something does not belong. An environment variable [was added](https://github.com/docker/compose/pull/4992) that controls whether this warning is seen or not, but I haven't gotten this to work for me.

For now we may have to ignore it.

pbx (approved):

jonkiparsky (approved):

PR #269, opened 2019-03-08
[AC-6483] Handle case where no submitted applications of a judging rounds application type exist

kaymation:
[JIRA](https://masschallenge.atlassian.net/browse/AC-6483)

to test
* load a clean db with `make load-db`
* login with demo admin, give yourself exec/md clearance for all program families using the following in django-shell
```python
from accelerator.models import *
d = User.objects.get(email__startswith="demoadmin")
for pf in ProgramFamily.objects.all():
    Clearance.objects.create(user=d, level=CLEARANCE_LEVEL_EXEC_MD, program_family=pf)
```
* check out `development`
* judging round 48 has existing criteria, so it's a good test case on clean db. run the following in the django-shell to set the applications to the state that causes the error
* first, hit the following endpoint, it should return useful data, note that the 'needy_apps' field in each result is a nonzero number
http://localhost/api/v1/analyze_judging_round/48
* next, unsubmit the existing applications for that round
```
jr = JudgingRound.objects.get(pk="48")
Application.objects.filter(application_type=jr.application_type).filter(application_status="SUBMITTED").update(application_status='hidden')
```
* hit the analyze endpoint again, it should 500 
http://localhost/api/v1/analyze_judging_round/48
* check out this branch and refresh
* the api should respond without error, the 'needy_apps' field on each result should be '0'
* finally, reverse the unsubmit and see that the view still works:
```
Application.objects.filter(application_type=jr.application_type).filter(application_status='hidden').update(application_status='SUBMITTED')
```

There is a [Django Accelerator PR here](https://github.com/masschallenge/django-accelerator/pull/150) to support the unit test

### DIFF
--- a/web/impact/impact/tests/test_analyze_judging_round_view.py
+++ b/web/impact/impact/tests/test_analyze_judging_round_view.py
@@ -210,6 +210,18 @@ class TestAnalyzeJudgingRoundView(APITestCase):
                     assert commitment.capacity - 1 == result[
                         'remaining_capacity']
 
+    def test_get_when_no_applications_for_round(self):
+        context = AnalyzeJudgingContext(add_application=False)
+        judging_round_id = context.criterion.judging_round.id
+        email = self.global_operations_manager(program_family).email
+        with self.login(email=email):
+            url = reverse(AnalyzeJudgingRoundView.view_name,
+                          args=[judging_round_id])
+            response = self.client.get(url)
+            results = response.data["results"]
+            for result in results:
+                assert result["needy_applications"] == 0    
+
 
 def _industry_options(context):
     return [app.startup.primary_industry.name
--- a/web/impact/impact/tests/test_analyze_judging_round_view.py
+++ b/web/impact/impact/tests/test_analyze_judging_round_view.py
@@ -213,6 +213,7 @@ class TestAnalyzeJudgingRoundView(APITestCase):
     def test_get_when_no_applications_for_round(self):
         context = AnalyzeJudgingContext(add_application=False)
         judging_round_id = context.criterion.judging_round.id
+        program_family = context.program.program_family
         email = self.global_operations_manager(program_family).email
         with self.login(email=email):
             url = reverse(AnalyzeJudgingRoundView.view_name,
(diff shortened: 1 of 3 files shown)


jonkiparsky (changes requested):
We need a test for this change.

web/impact/impact/tests/test_analyze_judging_round_view.py
```diff
@@ -210,6 +210,18 @@ def test_get_with_commitment(self):
                     assert commitment.capacity - 1 == result[
                         'remaining_capacity']
 
+    def test_get_when_no_applications_for_round(self):
+        context = AnalyzeJudgingContext(add_application=False)
+        judging_round_id = context.criterion.judging_round.id
+        email = self.global_operations_manager(program_family).email
```
jonkiparsky, line 217: `program_family` does not seem to be defined here. (Travis will probably complain about this)

kaymation: shoot, this didnt even error on me when i ran `make test`!! 😠

jonkiparsky (changes requested):
This cannot merge until https://github.com/masschallenge/django-accelerator/pull/150#pullrequestreview-213068524 has merged.

jonkiparsky (approved):
Looks good to me. (Code pass, this review does not include QA)

pbx:
@kaymation I may make it through before you revisit this, but the instructions on this PR need more [mint](https://github.com/masschallenge/standards/blob/master/kaizens/code-review/mint-on-the-pillow.md).

* Re "pick a judging round" -- I need to know what criteria I should use.  Picking some random round does not appear to work.  Providing a `JudgingRound` PK from the `initial_schema` data would be helpful for instance.
* Re "the api should respond without error" -- it would be good to know what the expected response content is.
* The "hit the following endpoint" URL you give does 500 -- on both `development` and the feature branch in fact -- because it lacks a JudgingRound PK.

pbx (changes requested):
Unfortunately, I am not able to get the `analyze_judging_round` call to respond -- every call to http://localhost/api/v1/analyze_judging_round/48/ results in a **"504 Gateway Time-out"**.  Other API calls return responses normally.

* Data: fresh `initial_schema` (just ran `make load-db`)
* User: demoadmin, with `CLEARANCE_LEVEL_EXEC_MD` on all program families

pbx (approved):

PR #278, opened 2019-04-12
[AC-6604] updated version number to match latest git tag

pbx:
## See [AC-6604](https://masschallenge.atlassian.net/browse/AC-6604)


VincentHokie:
Quick update, @pbx , @geofrocker tried this out and it didn't work unfortunately. I'm going to fall to option B that's on the ticket and try that out, seeing as BOS is out today.

 There will unfortunately be no PR for it.

Will report progress here/ on the ticket.

VincentHokie (approved):
✅

geofrocker (approved):

PR #280, opened 2019-04-15
AC-6617: Merge master into development to resolve merge conflict

VincentHokie:
#### Why this ticket
When creating a PR from development to master for the release per the [AC-6604 ticket](https://masschallenge.atlassian.net/browse/AC-6604) we run into a merge commit on the [PR](https://github.com/masschallenge/impact-api/pull/279). The fix is to merge master into development, leading o this.

#### Reason for change history
Development has never had changes from master (as far as I know).

We have been releasing off master since the beginning of time.
The AC-6604 commit changed a file on development that we have always changed in master and not development.

What this PR is reflecting is every commit that is on master and not development.

I'll leave this open until the team gets back to discuss this PR.

To not have this happen again, I propose after every release, we merge master into development.

#### Note
This PR is currently blocking a release [PR to master](https://github.com/masschallenge/impact-api/pull/279)


shakirandagire (approved):
The changes made in this process look good to me

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
pbx: This file never should have been committed. It would be a kindness if you could `git rm` it as part of this PR.

pbx (approved):
Good to go, one cleanup request.

.vscode/settings.json
VincentHokie: Yeah, I `git rm --cached` this file. So it should not show up in our history after this.

pbx: OK cool, my mistake.

PR #282, opened 2019-04-16
AC-6332: "make checkout branch" command fails if branch has not been checked out locally

VincentHokie:
#### Changes introduced in [AC-6332](https://masschallenge.atlassian.net/browse/AC-6332)
- ensure remote branches can be checked out by 'make checkout'
- add syntax highlighting to catch users attention on important text

#### How to test
There is currently a **AC-6332** branch on the accelerate and impact remote
- while on the development branch on impact
- run `make checkout branch=AC-6332`, note that it fails to checkout the branch on accelerate (and impact)
- now checkout this branch on impact
- try and run `make checkout branch=AC-6332` again and notice we have now checked out the AC-6332 branch on accelerate which was not previously possible.

As a bonus, text colour has been added to catch a users attention on important text.

#### Note
Whoever does the merge of this PR, please cleanup the AC-6332 remote on accelerate i.e. run `git push --delete origin AC-6332` on accelerate.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,6 +163,8 @@ no_release_error_msg = RELEASE must be set.  E.g., 'make deploy RELEASE=1.2.3.4'
 
 .PHONY: $(targets) $(deprecated_targets)
 
+YELLOW=\033[0;33m
+NO_COLOR=\033[0m
 
 .env:
 	@touch .env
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -256,11 +258,17 @@ status:
 checkout:
 	@for r in $(REPOS) ; do \
 		cd $$r; \
-		git show-ref --verify --quiet refs/heads/$(branch); \
+		git fetch 2>/dev/null; \
+		if [ $$? -ne 0 ]; then \
+			echo "$(YELLOW)Fetching the latest from the remote failed, you may not be able to checkout an existing remote branch.$(NO_COLOR)"; \
```
jonkiparsky: I think we should not introduce visual noise as a side-effect to a PR. 
If we want to use color in our process we should probably introduce this more intentionally (particularly since two people on engineering/product staff have color blindness)

jonkiparsky (changes requested):
Tested, works as advertised, please remove the colorization.

Makefile
VincentHokie: ah, I did not know this. Removed. Thanks.

jonkiparsky (approved):

mkiterian (approved):
Works!

PR #283, opened 2019-04-16
[AC-6535] Allocator analysis view: Judge criteria are not displaying remaining capacity

jonkiparsky:
https://masschallenge.atlassian.net/browse/AC-6535

To test:
- load clean-db

- fix gender criteria, which were created wrongly on clean db:
```
from accelerator.models import CriterionOptionSpec
CriterionOptionSpec.objects.filter(option="female").update(option="f")
CriterionOptionSpec.objects.filter(option="male").update(option="m")
```
- start servers, visit http://localhost:1234/allocator and select `2017 BOS CH IL`
- note that you should see "Available" values for roles and genders (except Lawyer, which we actually did run out of in that round)
- hit `http://localhost:8000/api/v1/analyze_judging_round/48/` to see the API bundle returned


VincentHokie:
taking code quality

VincentHokie:
the [directory.html](https://github.com/masschallenge/impact-api/pull/283/files#diff-ddd3c2c434dd8fef7e2de4d49887c1de) shouldn't be committed) file should not be committed. If changes are showing up it's because it was once committed by mistake. it's already on the .gitignore, a `git rm --cached` will make sure changes on it don't show up.

web/impact/impact/v1/helpers/matching_criterion_helper.py
```diff
@@ -14,6 +14,7 @@ class MatchingCriterionHelper(CriterionHelper):
     A matching criterion's weight and count cannot be set independently for
     different "options", since there are no distinct options. 
     '''
+    cache_key = ""
```
VincentHokie, line 17: missing new line here

VincentHokie:
`code-check` is needed on your branch, I'm getting several violations

```
web/impact/impact/v1/classes/option_analysis.py:164:80: E501 line too long (87 > 79 characters)
web/impact/impact/v1/classes/option_analysis.py:168:49: W291 trailing whitespace
web/impact/impact/v1/classes/option_analysis.py:169:80: E501 line too long (81 > 79 characters)
web/impact/impact/v1/helpers/criterion_helper.py:24:15: W291 trailing whitespace
web/impact/impact/v1/helpers/judge_criterion_helper.py:9:67: W291 trailing whitespace
web/impact/impact/v1/helpers/judge_criterion_helper.py:13:69: W291 trailing whitespace
web/impact/impact/v1/helpers/judge_gender_criterion_helper.py:12:1: W293 blank line contains whitespace
web/impact/impact/v1/helpers/matching_criterion_helper.py:11:73: W291 trailing whitespace
web/impact/impact/v1/helpers/matching_criterion_helper.py:13:48: W291 trailing whitespace
web/impact/impact/v1/helpers/matching_criterion_helper.py:15:62: W291 trailing whitespace
web/impact/impact/v1/helpers/matching_criterion_helper.py:18:5: E301 expected 1 blank line, found 0
web/impact/impact/v1/views/analyze_judging_round_view.py:50:1: W293 blank line contains whitespace
```

VincentHokie:
Also, I recall us finding a single line logic change to fix this with some failing tests. Did you find that the tests are well written and therefore the single line change is not valid?

jonkiparsky:
@VincentHokie Thanks for that review. 
Regarding the one-liner, moving to an explicitly consistent cache key seemed like a more solid fix going forward. I believe this does not affect the allocation process, and I have some experimental evidence to support this, but if you have the time and want to double-check that aspect of things, that would be a big help

VincentHokie:
Okay, I'll dig deep into it soon as I can. I was curious about the approach is all, but I see your perspective now.

web/impact/impact/v1/helpers/judge_criterion_helper.py
```diff
@@ -1,25 +1,31 @@
...
 
     def analysis_fields(self):
         return [
             self.judge_field,
         ]
+
+    def analysis_annotate_fields(self):
+        return {
+            self.cache_key: F(self.judge_field),
+            }
```
VincentHokie, line 31: seeing as this `analysis_annotate_fields` function is defined in both the children of this class i.e. `JudgeGenderCriterionHelper` and `JudgeRoleCriterionHelper`, I think this can go

web/impact/impact/v1/helpers/judge_criterion_helper.py
```diff
@@ -1,25 +1,31 @@
...
-    example, "one female judge and one male judge" would require two 
+    example, "one female judge and one male judge" would require two
     OptionSpecs. Weight and count can be set independently for each option.
     '''
 
     def option_for_field(self, field):
         return field
 
     def judge_matches_option(self, judge_data, option):
-        return option == judge_data.get(self.cache_judge_field)
```
VincentHokie, line 20: There has been a bit of a misunderstanding, coz I'm lost. I can understand the logic and all but the consistency is not right I think.

This field, the `cache_judge_field` was created for a single use-case, that is when querying data from the `JudgeDataCache`. Now the new field `cache_key` is being used across the logic as the key used on the `populate_criterion_total_capacities_cache`.

web/impact/impact/v1/helpers/judge_role_criterion_helper.py
```diff
@@ -8,10 +8,12 @@
 class JudgeRoleCriterionHelper(JudgeCriterionHelper):
     cache_judge_field = "expertprofile__expert_category__name"
     judge_field = "judge__" + cache_judge_field
+    cache_key = cache_judge_field
 
     def analysis_annotate_fields(self):
         return {
             "role": F(self.judge_field),
+            self.cache_key: F(self.judge_field),
```
VincentHokie, line 16: also, the logic were adding here for example is querying the exact same data but with a different name

web/impact/impact/v1/helpers/judge_criterion_helper.py
jonkiparsky: The problem was that having two different names for the entry in the cache caused the judge_matches_option check to fail sometimes. So the fix was to use the same name throughout. 
I feel like there's more refactor to get to here, the question is, is now the right time to do it?

VincentHokie: Okay, we could leave this as is but I'll put comments elsewhere so that were not duplicating some of our logic

web/impact/impact/v1/helpers/judge_role_criterion_helper.py
jonkiparsky: I'm not sure yet what purpose "role" serves here. I'd love to be able to remove it, but this is a piece of the complexity that I have not yet been able to reduce

VincentHokie: ah, okay. so in all of these `analysis_annotate_fields` functions. The keys are used in the `analysis_tally` functions to retrieve the value from this [cache](https://github.com/masschallenge/impact-api/blob/0063180e4f8b6a85bdd82d7d5a1f8434907df0f7/web/impact/impact/v1/classes/option_analysis.py#L185)

pbx:
Taking QA.

web/impact/impact/v1/classes/option_analysis.py
```diff
@@ -158,13 +161,19 @@ def calc_capacity(self, option_name, helper, spec_helper):
...
+                                                  cache_key):
+        '''Query total capacity data and cache it.
+        Use annotation to ensure that cache keys are consistent'''
         if self.criterion_total_capacities.get(option_name) is None:
-            capacities = JudgeRoundCommitment.objects.filter(
-                judging_round=self.judging_round).values(field).annotate(
-                    total=Sum("capacity"))
-            result = {cap[field]: cap["total"]
+            capacities = JudgeRoundCommitment.objects.annotate(
+                **{cache_key: F(field)}).filter(
```
pbx: Can we do this query without using `**`? It impairs readability for sure. If I have to fix a bug involving these lines, I will have to really slow down to figure out what this means.

If for some reason it must be used, it would be a kindness to include a brief comment  explaining how the query works.

pbx (changes requested):
QA: pass.

pbx (approved):

PR #288, opened 2019-05-02
AC 6573 Integrate MassChallenge semantic customizations into React/Front-end

freshprincekla:
**Changes introduced in [AC-6573](https://masschallenge.atlassian.net/browse/AC-6573)**
 - Added service to build and compile the custom semantic ui theme
 - Added commands to tag the semantic repo

**How to test**
- ensure that you have pulled the changes in this branch, semantic-ui-theme and front-end and are on branch AC-6573 in all these repos.
- run `make run-all-servers` to run both accelerate and front-end.
- make changes in the semantic-ui-theme repo and then changes will be seen in accelerate and in front-end

**For example**
- in semantic-ui-theme repo, navigate to the `message.overrides` file. 
  `src > themes > masschallenge > collections > message.overrides`
- insert the following property `* { background: #000}` and save.
- this will change the background color in accelerate and directory to black


VincentHokie (approved):
Code looks good ✅

PR #290, opened 2019-05-14
[AC-6077] Remove capacity update logic

jonkiparsky:
Per discussion with Jeff, determined that the correct thing is to simply remove the capacity update logic and associated tests, as it is not needed. 

Changed behavior: 
when a judge self-allocates in a dynamically-allocated round, their capacity is not changed even if they go above their originally stated capacity. 

Capacity is reflected on the user's /panels page, as in the screen shot. 

Testing hints: 
pick a judging round, say pk=48
set it to allow_dynamic_allocation
update is_active to True and end_date_time to > now 
pick a JudgeRoundCommitment for that round and update its capacity to zero
sign in as that judge and navigate to /panels and allocate yourself a panel
note that capacity does not change

Screen Shot 2019-05-14 at 2 50 05 PM

### DIFF
--- a/web/impact/impact/tests/test_allocate_applications_view.py
+++ b/web/impact/impact/tests/test_allocate_applications_view.py
@@ -94,19 +94,6 @@ class TestAllocateApplicationsView(APITestCase):
             assert [NO_APP_LEFT_FOR_JUDGE.format(context.judge.email)
                     in response.data]
 
-    def test_get_adds_capacity_and_quota(self):
-        context = AnalyzeJudgingContext()
-        judging_round = context.judging_round
-        commitment = context.judges[0].judgeroundcommitment_set.first()
-        with self.login(email=self.basic_user().email):
-            url = reverse(AllocateApplicationsView.view_name,
-                          args=[judging_round.id, context.judge.id])
-            response = self.client.get(url)
-            assert response.status_code == 200
-            commitment.refresh_from_db()
-            assert commitment.capacity > 0
-            assert commitment.current_quota > 0
-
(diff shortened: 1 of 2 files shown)


VincentHokie:
taking QA

shakirandagire:
Am doing code review.

VincentHokie (changes requested):
The commitment remains the same as expected, however I noticed an odd thing. With my commitment at 13, when I request applications I only get 10. Is this the expected behaviour?
I noticed it's because we give applications that are equal to the [Scenario panel_size](https://github.com/masschallenge/impact-api/blob/43712c259f70636c3339a07721c6b788da941ff5/web/impact/impact/v1/views/allocate_applications_view.py#L166-L177) and when we [create the Scenario](https://github.com/masschallenge/impact-api/blob/43712c259f70636c3339a07721c6b788da941ff5/web/impact/impact/v1/views/allocate_applications_view.py#L230-L231), we don't specify a panel_size and so the [default](https://github.com/masschallenge/django-accelerator/blob/a4123586f3fa90278b222efffc6a944af97540e9/accelerator_abstract/models/base_scenario.py#L12) is taken.

shakirandagire (approved):
Looks like it is only deletions.
LGTM

jonkiparsky:
@VincentHokie We allocate in batches of panel_size. The user's capacity does not and should not affect this.

VincentHokie:
Got it @jonkiparsky ! Thanks, I thought that was the number of applications allocated was equal to the capacity. Reminder that correlation != causation.

VincentHokie (approved):
✅

PR #291, opened 2019-05-15
AC-6715: Silence logging when running tests on impact-api

VincentHokie:
#### Changes introduced in [AC-6715](https://masschallenge.atlassian.net/browse/AC-6715)
- stop verbose logging when running tests

#### How to test
- while on development
- run `make test` in impact and notice the logging that drowns out basically everything
- now checkout AC-6715
- run `make test` again and see our tests are back to cleanliness 🏆

### DIFF
--- a/web/impact/impact/settings.py
+++ b/web/impact/impact/settings.py
@@ -319,8 +319,7 @@ class Test(Base):
     DATABASE_ROUTERS = []
     DEBUG = False
     LANGUAGE_CODE = 'en'
-
-    logging.disable(logging.CRITICAL)
+    logging.disable(logging.INFO)
 
 
 class Prod(Base):
--- a/web/impact/impact/settings.py
+++ b/web/impact/impact/settings.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2017 MassChallenge, Inc.
 
 import os
-
+import logging
 import datetime
 from configurations import (
     Configuration,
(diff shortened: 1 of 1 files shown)


shakirandagire:
Doing QA

shakirandagire (approved):
QA looks good and clean

web/impact/impact/settings.py
```diff
@@ -320,6 +320,8 @@ class Test(Base):
     DEBUG = False
     LANGUAGE_CODE = 'en'
 
+    logging.disable(logging.CRITICAL)
```
jonkiparsky: Per ticket, we'd like to see WARNINGs. This looks like we'd see no logging at all.

jonkiparsky (changes requested):

web/impact/impact/settings.py
VincentHokie: the log messages that show up are [warning messages](https://github.com/masschallenge/django-accelerator/blob/a4123586f3fa90278b222efffc6a944af97540e9/accelerator/managers/member_profile_manager.py#L17) @jonkiparsky , allowing warning shows those as well

jonkiparsky: Okay, then perhaps the issue is that that's not the right log level for that situation. After all, if it's something that happens twice with every test we run, it's probably not a thing we need to be WARNed about, is it? Probably INFO would be sufficient, no?

We certainly don't want to silence warnings on our tests, since warnings indicate things we need to fix.

VincentHokie: that works for me, this will need another PR. Raising in a few

VincentHokie: PR raised [here](https://github.com/masschallenge/django-accelerator/pull/183), @jonkiparsky

jonkiparsky (approved):
Looks good, approving.

PR #292, opened 2019-05-15
AC-6533: Add 'keepdb' option to Makefile test targets

VincentHokie:
#### Changes introduced in [AC-6533](https://masschallenge.atlassian.net/browse/AC-6533)
- add keepdb option to 'make test' to allow a user opt out of keeping the db while running tests

#### How to test
- checkout this branch
- run `time make test tests=impact.tests.test_authentication_views` twice, and notice the second time, the test db is not recreated (the first time, the db is actually created)
- now run `time make test tests=impact.tests.test_authentication_views keepdb=no` and notice a prompt appears asking you to destroy the existing db.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,8 +188,9 @@ build: shutdown-vms delete-vms setup
 
 tests ?= $(TESTS)  # Backwards compatibility
 test: setup
-	@docker-compose run --rm web \
-		python3 manage.py test --configuration=Test $(tests)
+	@@if [ -z $$keepdb ]; then keepdb="--keepdb"; else keepdb=""; fi; \
+	docker-compose run --rm web \
+		python3 manage.py test $$keepdb --configuration=Test $(tests)
 
 coverage: coverage-run coverage-report coverage-html-report
 
--- a/Makefile
+++ b/Makefile
@@ -69,8 +69,9 @@ lower_target_help = \
 target_help = \
   'help - Prints this help message.' \
   ' ' \
-  'test - Run tests with no coverage. Run just those specified in $$(tests)' \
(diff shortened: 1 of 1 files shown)


pbx (changes requested):
This needs an update to `target_help` explaining the `keepdb` option.

VincentHokie:
thanks, added

pbx (approved):

kitaka:
@VincentHokie @pbx 
Given that people have been accustomed to running tests with keepdb off, why don't we maintain that as the default and making keepdb on as the option?

kitaka:
Also, running `make test` the first time - passes but on the second run given the reused DB - it fails. That's not good behaviour to have in a default state.

pbx:
@kitaka - the default is a matter of taste, and I'm fine with what the team prefers, but personally I prefer having `--keepdb` _on_ by default. It's what I am going to want 9 out of 10 times (if not more) when I am running tests locally. 

Re the errors on subsequent runs, I've seen that but don't have a good recipe for recreating _or_ fixing. Can you share repro steps?

jonkiparsky:
Drive-by comment: Errors on subsequent runs indicate that a test is not sufficiently isolated. This is useful information and we should use it to fix any tests that fail in this manner.

kitaka:
> Drive-by comment: Errors on subsequent runs indicate that a test is not sufficiently isolated. This is useful information and we should use it to fix any tests that fail in this manner.

You raise a good point. If the cost is acceptable, then I have no reservations.

VincentHokie:
haven't been able to reproduce that failing test, how do I repro?

I tried running `make test` twice.

PR #294, opened 2019-05-22
AC-6678: Allow a merge-to-deploy workflow to an AWS Stack e.g pre-staging

VincentHokie:
#### Changes introduced in [AC-6678](https://masschallenge.atlassian.net/browse/AC-6678)
- add logic that triggers an an ecs deploy when development is built

#### How to test
- rerunning the last [impact-api build](https://travis-ci.org/masschallenge/impact-api/builds/536757005?utm_source=github_status&utm_medium=notification) for this branch a deploy will be triggered at the end
- you can "see" the deployment in action on [ECS](https://console.aws.amazon.com/ecs/home?region=us-east-1#/clusters/pre-staging/services/testing-impact-service/tasks). you will notice that the re are now 2 tasks running showing that a deployment is in progress. The way deployments work, is a new task is created (to make a total of 2) and once that is healthy the old one is destroyed leaving the new task with the new code.
- also on the [events](https://console.aws.amazon.com/ecs/home?region=us-east-1#/clusters/pre-staging/services/testing-impact-service/events) section on ECS, you will notice a sequence of events that show a deployment took place. i.e.

```
service testing-impact-service has reached a steady state.
service testing-impact-service has stopped 1 running tasks: task a5d5b70b-f767-42e9-a444-97cdc367ad99.
service testing-impact-service has begun draining connections on 1 tasks.
service testing-impact-service deregistered 1 targets in target-group impact-pre-staging-target-group
service testing-impact-service registered 1 targets in target-group impact-pre-staging-target-group
service testing-impact-service has started 1 tasks: task 822f382f-f5b7-4606-8fc6-8ff4fb3c6eb3.
service testing-impact-service has reached a steady state.
```

ps, the logs are read from bottom to top


Makefile
```diff
@@ -472,12 +463,34 @@ endif
...
+	@if [ "$$IMAGE_TAG" != "development" ] && [ "$$IMAGE_TAG" != "AC-6678" ]; then exit 1; fi;
+	@pip install --upgrade certifi pyopenssl requests[security] ndg-httpsclient pyasn1 pip botocore
+	@curl -s https://raw.githubusercontent.com/silinternational/ecs-deploy/master/ecs-deploy | sudo tee /usr/bin/ecs-deploy
+	@sudo chmod +x /usr/bin/ecs-deploy
+	@git clone https://$$GITHUB_ACCESS_TOKEN@github.com/masschallenge/infrastructure -b pre-staging
+	@wget https://releases.hashicorp.com/terraform/0.11.11/terraform_0.11.11_linux_amd64.zip -q
+	@sudo unzip terraform_0.11.11_linux_amd64.zip -d /usr/bin/
+	@sudo echo "deb http://us.archive.ubuntu.com/ubuntu vivid main universe" >> /etc/apt/sources.list
+	@cd infrastructure/environment/test/static-secrets; \
+	terraform init; \
```
kitaka: @VincentHokie why do we need terraform? My understanding was that Terraform is used for infrastructure automation and with that in mind, this doesn't seem to be the right application.
Can't we trigger some sort of refresh with just the `aws cli`?

VincentHokie: there's a cli trigger that is done [here](https://github.com/masschallenge/impact-api/pull/294/files#diff-b67911656ef5d18c4ae36cb6741b7965R487) incase the infrastructure changes themselves don't lead to a deploy.

if e.g. we edit our [task definition](https://github.com/masschallenge/infrastructure/blob/development/super_stack/ecs-cluster/impact/task-def.json) (with say a new secret), terraform would do 2 things.
- update the infrastructure &
- perform a deploy on ECS

Adding terraform here has the benefit of us being able to test continuous changes to our infrastructure as well.

kitaka: In the spirit of separation of concerns. Would it not make more sense to have the terraform on the infrastructure repo and keeping this to the simple task of triggering a deploy? 
The simpler, the easier it will be to maintain by people without terraform experience.

VincentHokie: sure, makes sense. done.

VincentHokie: note though, that were going to have several stacks up built off e.g master. So a single travis build on the infrastructure repo updating > 1 AWS Stack won't be an easy thing to accomplish.

But then again, infra changes aren't regular. So we'll see how it goes.

kitaka: Do we really need to have continuous deployment on infrastructure? A bug in Accelerate code might not be as disruptive as one in the Infrastructure code. I wonder what the best practices are in the infrastructure as code community.

VincentHokie: I think we can hold off on CD for infra for now.

kitaka:
@VincentHokie I think a way to test is necessary. Is it possible to have some sort of `ping repo` that works exactly like development - in that when something is merged to it, the deploy happens?

Also, I believe a new task should be created on ECS upon a successful deployment with a start date that should match the time of deployment.

VincentHokie:
A task definition does not `need` to be created since nothing has changed. The one on staging/prod does this because it uses a different tool. and that tool did not correctly create the task definition on deployment, which is what led to terraform.

About the testing 😩 forgot to update this repo. The rest have this already. Updating now.

VincentHokie:
Updated.

pbx:
Tagged "Don't merge" based on this: "Before this PR is merged, I will need to stop AC-6678 from performing a continuous deploy situation"

kitaka:
> Also, I believe a new task should be created on ECS upon a successful deployment with a start date that should match the time of deployment

I meant it as a way of verifying that the deployment works!

kitaka (approved):

VincentHokie:
> I meant it as a way of verifying that the deployment works!

ah 🤦‍♂

Makefile
```diff
@@ -474,12 +465,14 @@ endif
...
 deploy:
-	@echo $@ ERROR: see deployment steps for accelerate.
-	@echo see: https://github.com/masschallenge/standards/blob/376d290b41a202acc5c2263d7275ba4a57330ad7/create_new_release.md#deploy-to-staging
+	@if [ "$$IMAGE_TAG" != "development" ] && [ "$$IMAGE_TAG" != "AC-6678" ]; then exit 1; fi;
+	@pip install --upgrade certifi pyopenssl requests[security] ndg-httpsclient pyasn1 pip botocore
+	@curl -s https://raw.githubusercontent.com/silinternational/ecs-deploy/master/ecs-deploy | sudo tee /usr/bin/ecs-deploy
+	@sudo chmod +x /usr/bin/ecs-deploy
+	@ecs-deploy -c $$PRE_STAGING_ECS_CLUSTER -n $$PRE_STAGING_ECS_SERVICE -i $$DOCKER_REGISTRY/impact-api:$$IMAGE_TAG --force-new-deployment
 
+# Deprecated targets
```
shanbady, line 475: this may still be relevant but im not going to block on it

Makefile
```diff
@@ -474,12 +465,14 @@ endif
 	@ecs-cli compose -f docker-compose.prod.yml up
 
 
-
-# Deprecated targets
 deploy:
-	@echo $@ ERROR: see deployment steps for accelerate.
-	@echo see: https://github.com/masschallenge/standards/blob/376d290b41a202acc5c2263d7275ba4a57330ad7/create_new_release.md#deploy-to-staging
+	@if [ "$$IMAGE_TAG" != "development" ] && [ "$$IMAGE_TAG" != "AC-6678" ]; then exit 1; fi;
```
shanbady, line 469: is this hard-coded branch name needed once merged?

Makefile
```diff
@@ -474,12 +465,14 @@ endif
...
+	@pip install --upgrade certifi pyopenssl requests[security] ndg-httpsclient pyasn1 pip botocore
+	@curl -s https://raw.githubusercontent.com/silinternational/ecs-deploy/master/ecs-deploy | sudo tee /usr/bin/ecs-deploy
```
shanbady, line 471: could we/should we store this script on our own/this repo?

web/impact/impact/settings.py
```diff
@@ -161,7 +161,8 @@ class Base(Configuration):
         'localhost:8181',
     )
     CORS_ORIGIN_REGEX_WHITELIST = (
-        r'^(https?://)?(\w+\.)?masschallenge\.org$', )
+        r'^(https?://)?(\w+\.)?masschallenge\.org$',
+        r'^(https?://)?(\w+)(-)?(\w+)?.masschallenge\.org$', )
```
shanbady: does the "." just before "masschallenge" need to be escaped?

shanbady (changes requested):

web/impact/impact/settings.py
VincentHokie: pre-staging works without escaping the `.`

so I'd say, no. We don't need to escape this

Makefile
VincentHokie: the hard-coded `AC-6678` can go..and there's a [ticket](https://masschallenge.atlassian.net/browse/AC-6754) to clean this up once merged.

Makefile
VincentHokie: We can, I just think in case AWS changes its API for some reason..that overhead would fall on us to update. Pulling it from the remote allows us to always have a "valid" script

web/impact/impact/settings.py
shanbady: while it will validate on the cases we care about without the escape this regex matches any domain so long as it ends in "masschallenge.org" (ex: https://maliciousdomainmasschallenge.org)

VincentHokie: ohh, I see what you mean! good catch! fixing now

VincentHokie: fixed.

shanbady (approved):

PR #302, opened 2019-06-20
[AC-6761] API endpoint for Office hours that's filterable by Users

mkiterian:
### Changes introduced in [AC-6761](https://masschallenge.atlassian.net/browse/AC-6761):
Add an office hours endpoint filterable by mentor id and/or finalist id

### How to Test

- Run impact on this branch and visit: http://localhost:8000/api/v1/office_hours/ and use `mentor_id`, `mentor_name`, `finalist_id` and `finalist_name` as url parameters to do the filtering
- `mentor_name` and `finalist_name` should also return results for partial matches


VincentHokie:
taking code

web/impact/impact/v1/views/mentor_program_office_hour_list_view.py
```diff
@@ -0,0 +1,60 @@
...
+        filter_values = {
+            key: value for key, value in param_items
+            if key in ID_FIELDS and value.isdigit()}
+        if filter_values:
+            return qs.filter(**filter_values)
+        return qs.none()
+
+    def _has_participant_filter(self, fields):
+        return any(
+            key in self.request.query_params.keys() for key in fields)
```
VincentHokie: you could probably get this with
```
subtracted_list = list(set(fields) - set(self.request.query_params.keys()))
return len(subtracted_list) < len(fields)
```

but not a blocker for me, just a suggestion.
I think we should avoid looping through the keys

web/impact/impact/v1/views/mentor_program_office_hour_list_view.py
```diff
@@ -0,0 +1,60 @@
...
+    def _filter_by_participant_name(self, qs):
+        params = self.request.query_params
+        mentor_name = params.get('mentor_name', None)
+        finalist_name = params.get('finalist_name', None)
+
+        if mentor_name:
+            return self._generate_name_query(qs, 'mentor', mentor_name)
+        if finalist_name:
+            return self._generate_name_query(qs, 'finalist', finalist_name)
+        return qs.none()
```
VincentHokie: this function implies we can't filter by both `mentor_name` **and** `finalist_name` since the function terminates if there's a mentor name. is it expected that we only filter by one of them?

web/impact/impact/v1/helpers/mentor_program_office_hour_helper.py
```diff
@@ -0,0 +1,61 @@
...
+    "start_time": OPTIONAL_STRING_FIELD,
+    "end_time": OPTIONAL_STRING_FIELD,
+    "date": OPTIONAL_STRING_FIELD,
+    "mentor_id": OPTIONAL_INTEGER_FIELD,
+    "mentor_name": OPTIONAL_STRING_FIELD,
+    "mentor_email": OPTIONAL_STRING_FIELD,
+    "finalist_id": OPTIONAL_INTEGER_FIELD,
+    "finalist_name": OPTIONAL_STRING_FIELD,
+    "finalist_email": OPTIONAL_STRING_FIELD,
+}
```
VincentHokie, line 27: I believe the `reservation status` is missing in this list..it's specified on the ticket
i.e. Open vs Reserved

web/impact/impact/v1/views/mentor_program_office_hour_list_view.py
mkiterian: Yes that is the case. It wasn't spec'ed out in the ticket. I do not envision a use case for filtering by both `mentor_name` and `finalist_name`. Remember the use case here would be searching. In cases where we'd like to retrieve office hours for  a specific mentor and finalist combination, their ids would be the most appropriate. cc @shanbady, @hfreund

web/impact/impact/tests/test_mentor_program_office_hour_list_view.py
```diff
@@ -0,0 +1,164 @@
...
+    def _create_user_office_hours(self, mentor=True):
+        if mentor:
+            user = ExpertFactory()
+            office_hours = MentorProgramOfficeHourFactory.create_batch(
+                USER_COUNT, mentor=user)
+        else:
+            user = EntrepreneurFactory()
+            office_hours = MentorProgramOfficeHourFactory.create_batch(
+                USER_COUNT, finalist=user)
+        return user, office_hours
```
VincentHokie, line 194: you could probably get rid of the conditional by passing the factory as a function argument

web/impact/impact/tests/test_mentor_program_office_hour_list_view.py
```diff
@@ -0,0 +1,164 @@
...
+from impact.tests.factories import MentorProgramOfficeHourFactory
+from impact.tests.api_test_case import APITestCase
+from impact.v1.helpers.mentor_program_office_hour_helper import (
+    OFFFICE_HOUR_FIELDS,
+)
+from impact.tests.utils import assert_fields
+from impact.v1.views import MentorProgramOfficeHourListView
+
+USER_COUNT = 3
+NON_USER_COUNT = 4
```
VincentHokie: from the names it's not clear what `USER_COUNT` and `NON_USER_COUNT`..and I'm not quite sure what they are from the tests either

web/impact/impact/tests/test_mentor_program_office_hour_list_view.py
```diff
@@ -0,0 +1,164 @@
...
+        self.login(email=self.basic_user().email)
+        return self.client.get(self.url, params)
+
+    def _compare_ids(self, office_hours, response):
+        office_hour_ids = [
+            office_hour.id for office_hour in office_hours]
+        result_ids = [
+            result['id'] for result in response.data["results"]
+        ]
+        return all([id in result_ids for id in office_hour_ids])
```
VincentHokie: there's a world where you have 2 rather than 3 comprehensions here i.e.

```
    def _compare_ids(self, office_hours, response):
	        result_ids = [
	            result['id'] for result in response.data["results"]
	        ]
	        return all([office_hour.id in result_ids for office_hour in office_hours])
```

web/impact/impact/tests/test_mentor_program_office_hour_list_view.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def test_get(self):
+        _, office_hours = self._create_user_office_hours()
+        response = self._get_response_as_logged_in_user()
+        total_count = USER_COUNT + NON_USER_COUNT
+        self.assertEqual(response.data["count"], total_count)
+        self.assertTrue(self._compare_ids(office_hours, response))
+        self.assertTrue(self._compare_ids(self.test_office_hours, response))
+
+    def test_options(self):
```
VincentHokie: this test name could be better, where testing that the options do what? and which options?

VincentHokie (changes requested):
Left some requests for changes, some suggestions

web/impact/impact/v1/views/mentor_program_office_hour_list_view.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def _generate_name_query(self, qs, user, name_value):
```
kitaka: @mkiterian to avoid introducing more code to maintain, you can leverage that parent `_filter_by_name` method.

web/impact/impact/tests/test_mentor_program_office_hour_list_view.py
mkiterian: we wouldn't avoid the conditional as the arguments for `create_batch` correspond to fields in the MentorProgramOfficeHour model. We'd still have to either set the `finalist` or `mentor` fields

web/impact/impact/v1/views/mentor_program_office_hour_list_view.py
```diff
@@ -0,0 +1,60 @@
...
+        first_name_field = '{}__first_name'.format(user)
+        last_name_field = '{}__last_name'.format(user)
+        result = qs.annotate(
+            full_name=Concat(
+                first_name_field, V(' '), last_name_field)).filter(
+                    full_name__icontains=name_value)
+        return result
+
+    def _filter_by_participant_id(self, qs, param_items):
```
kitaka: @mkiterian this code looks very unreadable for a method that shouldn't be that complex. Consider leveraging the `filter_by_field` parent method.

kitaka (changes requested):

web/impact/impact/v1/views/mentor_program_office_hour_list_view.py
```diff
@@ -0,0 +1,60 @@
...
+            return self._generate_name_query(qs, 'mentor', mentor_name)
+        if finalist_name:
+            return self._generate_name_query(qs, 'finalist', finalist_name)
+        return qs.none()
+
+    def _generate_name_query(self, qs, user, name_value):
+        first_name_field = '{}__first_name'.format(user)
+        last_name_field = '{}__last_name'.format(user)
+        result = qs.annotate(
+            full_name=Concat(
```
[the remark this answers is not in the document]
mkiterian: `_generate_name_query ` is necessary because we do not have a `full_name` field for the mentors and finalists, we do have `first_name` and `last_name` fields but would like to perform filtering by the users full names

web/impact/impact/v1/views/mentor_program_office_hour_list_view.py
mkiterian: `filter_by_field` wouldn't work in this case coz we'd like to filter by `mentor_id` and/or `finalist_id`

web/impact/impact/v1/views/mentor_program_office_hour_list_view.py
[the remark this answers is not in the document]
mkiterian: `_filter_by_name` wouldn't work here. There is additional complexity because we are filtering by `full name` which does not exist as a field of the related model where it is to retrieved hence the concatenation of the fields while performing the query

web/impact/impact/v1/views/mentor_program_office_hour_list_view.py
mkiterian: WIll refactor to use filter_by_field

web/impact/impact/v1/views/mentor_program_office_hour_list_view.py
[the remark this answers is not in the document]
mkiterian: rename method to sth meaningful

web/impact/impact/v1/views/mentor_program_office_hour_list_view.py
hfreund: Herman's position seems reasonable to me.

jonkiparsky: #drive-by 
It seems like it would not be hard to make this a bit more flexible now, which would save us a lot of complaining later on when we decide we actually do want to have that flexibility. 
(when I say "not hard", I think the function actually gets simpler, since it can be a loop or comprehension)

web/impact/impact/v1/views/mentor_program_office_hour_list_view.py
```diff
@@ -0,0 +1,63 @@
...
+        if not self.request.query_params.keys():
+            return qs
+
+        if self._has_participant_filter(NAME_FIELDS):
+            return self._filter_by_participant_name(qs)
+
+        if self._has_participant_filter(ID_FIELDS):
+            return self._filter_by_ids(qs)
+
+    def _filter_by_ids(self, qs):
```
kitaka, line 38: @mkiterian there appears lots of replication; consider introducing a new method
```python
def _xxx(param_id):
  value = params.get('mentor_id', None)
  if value and value.isdigit():
    return self.filter_by_field('mentor_id', qs)
```

web/impact/impact/v1/views/mentor_program_office_hour_list_view.py
```diff
@@ -0,0 +1,63 @@
...
+        mentor_id = params.get('mentor_id', None)
+        finalist_id = params.get('finalist_id', None)
+
+        if mentor_id and mentor_id.isdigit():
+            qs = self.filter_by_field('mentor_id', qs)
+        if finalist_id and finalist_id.isdigit():
+            qs = self.filter_by_field('finalist_id', qs)
+        return qs
+
+    def _filter_by_participant_name(self, qs):
```
kitaka: @mkiterian just as above, the replication can be further reduced.

kitaka (changes requested):

web/impact/impact/tests/test_mentor_program_office_hour_list_view.py
VincentHokie: we could probably achieve that by passing in a dict that we expand within the function e.g.
```
    def _create_user_office_hours(self, Factory, user_dict):
        user = Factory()
        office_hours = MentorProgramOfficeHourFactory.create_batch(
            USER_OFFICE_HOUR_COUNT, **user_dict)
        return user, office_hours
```

then we would use it like so
```
user, office_hours = self. _create_user_office_hours(ExpertFactory, {"mentor": user})
```

web/impact/impact/v1/views/mentor_program_office_hour_list_view.py
VincentHokie: thanks for the change..one more thing here. I believe the condition now as is is `if the mentor has name A and the menteee has name B then get that entry`, I think the condition should instead be `or` to make this a bit more flexible

web/impact/impact/tests/test_mentor_program_office_hour_list_view.py
mkiterian: @sante in this case we'd still have to instantiate the user object from outside and pass it in as an argument

web/impact/impact/v1/views/mentor_program_office_hour_list_view.py
mkiterian: done

web/impact/impact/tests/test_mentor_program_office_hour_list_view.py
VincentHokie: oh I see, that code snippet actually doesn't make sense 😆

VincentHokie (approved):
✅  I foresee some tinkering with this once we know the UI that will consume this information but I'm comfortable calling this good now

kitaka (approved):
Looks good!

PR #345, opened 2019-11-15
[AC-7217] expose version number to all apps

philophilo:
### Changes introduced in: [AC-7217](https://masschallenge.atlassian.net/browse/AC-7217):
- creates a release number as an environment variable

### How to test
on this branch `AC-7217`
- delete the docker image for `impact-api`
- run `docker rmi impact-api_web`
- run `make run-detached-server`
- run `docker exec -it impact-api_web_1 /bin/bash` to access the impact-api container
- inside the container's terminal run `printenv | grep ACCELERATE_VERSION`
- this should print the latest release tag


pbx (changes requested):
Your build is failing due to a syntax problem.

In the interest of readability it might be worth trying the alternative form I suggested to Kevin -- `git describe --contains HEAD`

.travis.yml
```diff
@@ -40,13 +40,16 @@ before_script:
 - git clone https://$MC_DEV_ADMIN_GH_TOKEN@github.com/masschallenge/front-end -b $FRONT_END_REVISION ../front-end
 - cp .prod.env ../front-end/.env
 - cp .prod.env .env
+- ACCELERATE_VERSION=$(git describe --tags --abbrev=0)
+- export ACCELERATE_VERSION
+- echo $ACCELERATE_VERSION
```
pbx: I'd rather not add more output to the (already very noisy) Travis log, but if we include this let's do something like:
 `echo "ACCELERATE_VERSION: $ACCELERATE_VERSION"`

.travis.yml
```diff
@@ -64,6 +67,8 @@ after_success:
 - export ECR_HOST=`aws ecr get-authorization-token --region us-east-1 --output text --query 'authorizationData[].proxyEndpoint'`
 - export DOCKER_USER=AWS
 - export DOCKER_PASSWORD=$ECR_TOKEN
+- ACCELERATE_VERSION=$(git describe --tags --abbrev=0)
+- echo $ACCELERATE_VERSION
```
pbx: As above -- if we're adding a line to the logged output, let's make it clear.

Makefile
```diff
@@ -457,6 +457,7 @@ endif
 	export ECR_HOST=`aws ecr get-authorization-token --region us-east-1 --output text --query 'authorizationData[].proxyEndpoint'`; \
 	export DOCKER_USER=AWS; \
 	export DOCKER_PASSWORD=$$ECR_TOKEN; \
+	export ACCELERATE_VERSION=$('`git status | grep HEAD | cut -d ' ' -f4`')
```
pbx: Let's use the `git describe` recipe here too.

pbx (changes requested):
Just a couple things to tidy up.

.travis.yml
philophilo: sure, I was testing the `git` command if it had got the latest tag. It will not be existing in the final. The remaining bit is figuring the testing instructions. Since we cannot `ssh` into the containers, hoping to use cloudwatch since it also needs to be tested for production environments.

Makefile
philophilo: indeed, thanks! I will remove this line, it has no role here.

geofrocker:
QA

pbx:
I got an error when following the instructions above: 
```
$ docker exec -it impact-api_web_1 /bin/bash
Error: No such container: impact-api_web_1
```
Fixed it by running `make run-all-servers` first.

pbx (dismissed):
Following the instructions, I see `ACCELERATE_VERSION=development` rather than the latest release tag.

philophilo:
@pbx aah that is the default version for local development, updating the testing instructions to see it workout locally.
edit
testing instructions updated

pbx:
@philophilo - Thanks for the updated instructions. Unfortunately when I followed them, the `ACCELERATE_VERSION` env var existed, but was empty.

geofrocker (commented):
This is what I get when I follow the test instructions
```
root@7a38a0259958:/wwwroot# printenv | grep ACCELERATE_VERSION
ACCELERATE_VERSION=

**Comment by geofrocker** [comment]
Showing ACCELERATE_VERSION=v1.2.27 yet we are on v1.2.31.

**Comment by philophilo** [comment]
@pbx Locally I had an environment variable that misled my test. I have updated the `Makefile` to extend the version for local development. In this case, the command `git describe --tags --abbrev=0` is run from the `Makefile`

**Comment by pbx** [comment]
@philophilo - I followed the instructions (thanks for the update), and got:
```
ACCELERATE_VERSION=vTR-AC-7075
```
(The tagged commit: `ba0aa7d9d459bfb801e130027dd4b86cda260fb7 (tag: vTR-AC-7217-113, tag: vTR-AC-7217-112, tag: vTR-AC-7217-111, tag: vTR-AC-7075-v7, tag: vTR-AC-7075-v6, tag: vTR-AC-7075-v5, tag: vTR-AC-7075-v4, tag: vTR-AC-7075-v3, tag: vTR-AC-7075-v2, tag: vTR-AC-7075-deploy-2, tag: vTR-AC-7075-c, tag: vTR-AC-7075)`)

If these problems are side effects of the difficulty of testing this locally, perhaps we need a different approach? In the end what matters is, **the running accelerate code in the cloud can determine what release version it is.** Perhaps we could contrive a test whereby we do a deploy to a test machine and confirm it works there?

(A side benefit of that approach is that it's a lot less disruptive to reviewers' WIP in impact-api. This review work has been expensive for me as I'm trying to make progress on AC-7029 as well.)

**Comment by philophilo** [comment]
@pbx That is the expected tag. My assumption was that if a deployment is made on `production`, `staging` or a `test machine`, the tag should be of the release that was deployed. `git describe --tags --abbrev=0` gets the closest tag to the commit and removes the suffix.

**Comment by pbx** [comment]
@philophilo - but I don't think `vTR-AC-7075` _is_  the closest commit to `HEAD`. According to this check (which I believe is sound), the most recent release is much closer.
```
$ git rev-list HEAD...vTR-AC-7075 | wc -l
     154
$ git rev-list HEAD...v1.2.31 | wc -l
      40
```
Why do you say it's the expected tag?

If this turns out to be just a quirk with the local environment, I could let it pass -- **given a confirmation that be behavior is correct on AWS** -- but it looks like a bug to me.

shanbady (approved):
worked as described on my end. @geofrocker feel free to merge when satisfied. i spoke to paul and he  said we can dismiss his review here.

geofrocker (approved):
QA looks good

PR #353, opened 2019-12-05
[AC-7291] P0 - Running make run-server on impact rebuilds images

philophilo:
### Changes introduced in [AC-7291](https://masschallenge.atlassian.net/browse/AC-7291):
- Removes 'docker-compose build' when running containers in the `Makefile`

### How to test
on this branch
- Notice that when running `make run-server` or `make run-all-servers` impact is not rebuilt

Test that the latest version is still accessible in the container

- run `make build`
- run `make run-server`
- run `docker exec -it impact-api_web_1 /bin/bash` to access the impact-api container
- inside the container's terminal run `printenv | grep ACCELERATE_VERSION`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -182,7 +182,6 @@ setup: .env
 
 ACCELERATE_VERSION:=$(shell git describe --tags --abbrev=0)
 build: shutdown-vms delete-vms setup
-	@echo $(ACCELERATE_VERSION)
 	@docker build -f Dockerfile.fpdiff -t masschallenge/fpdiff .
 	@docker build  -f Dockerfile.semantic-release -t semantic-release .
 	@docker-compose build --build-arg ACCELERATE_VERSION=$(ACCELERATE_VERSION)
--- a/Makefile
+++ b/Makefile
@@ -180,10 +180,12 @@ setup: .env
 	@mkdir -p ./mysql/data
 	@mkdir -p ./redis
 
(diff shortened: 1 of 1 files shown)


jonkiparsky:
Will do QA

pbx:
Checking out the code.

Makefile
```diff
@@ -180,10 +180,12 @@ setup: .env
 	@mkdir -p ./mysql/data
 	@mkdir -p ./redis
 
+ACCELERATE_VERSION:=$(shell git describe --tags --abbrev=0)
 build: shutdown-vms delete-vms setup
+	@echo $(ACCELERATE_VERSION)
```
pbx: I'd like to see this removed. It feels like a debugging leftover from AC-7217.

pbx (approved):
This works as advertised -- I just have one small cleanup request (not a blocker, hence my approval).

Makefile
jonkiparsky: Either remove, or add some text such as "Using accelerate version: " so it's obvious what we're seeing. (It's not totally insane to confirm the version that you're building!)

jonkiparsky (approved):
Approving, leaving open in case @philophilo wants to take either of the suggested paths re: ACCELERATE_VERSION

@philophilo please feel free to merge this if you want to ignore those comments, or push the change if you want to respond. Thanks!

PR #357, opened 2019-12-18
[AC-7269] Use numpy array in place of deprecated matrix

jonkiparsky:
https://masschallenge.atlassian.net/browse/AC-7269

Verify that correct matrix multiplication syntax is used, replacing deprecated `matrix` class

Note: I used the `matmul` function rather than the `@` symbol, since the former seemed more communicative.

### DIFF
--- a/web/impact/impact/v1/classes/judge_data_cache.py
+++ b/web/impact/impact/v1/classes/judge_data_cache.py
@@ -1,5 +1,5 @@
 from collections import OrderedDict
-from numpy import matrix
+from numpy import array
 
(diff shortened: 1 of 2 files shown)


pbx (approved):
As far as I can tell from the code, [the docs](https://docs.scipy.org/doc/numpy/reference/generated/numpy.matrix.html), the passing tests, and the lack of matrix-related deprecation warnings, this is good.

shakirandagire (approved):
The changes made look good to me.

PR #363, opened 2020-01-16
[AC-6440] Remove current_program from graphql output

jonkiparsky:
https://masschallenge.atlassian.net/browse/AC-6440

Does what it says on the tin: removes current_program from the graphql output

Note that this should merge and deploy after or simultaneously with the related PR on front-end


freshprincekla (approved):
Code and QA look good.

iankigen (approved):

PR #369, opened 2020-02-10
[AC-7430] Display relevant program match I have with other users for all the user types

iankigen:
### Changes introduced in: [AC-7430](https://masschallenge.atlassian.net/browse/AC-7430):
- Provides API functionality to ensure that all the users (not only experts) can GET their `confirmed_user_program_families` and `latest_active_program_location` to be used on the front-end to get relevant program matches
### How to test
- Check out to  `AC-7430`  on `impact-api`
- Visit http://localhost:8000/api/v1/user/(logged-in-user-id)/
You should be able to see the current authenticated user dict of `confirmed_user_program_families` and string of `latest_active_program_location`
example of data: 
```
    "confirmed_user_program_families": {
        "Israel": "2019-02-06T22:46:00Z",
        "Boston": "2019-05-15T19:03:22Z",
        "Mexico": "2019-06-28T22:31:18Z"
    },
    "latest_active_program_location": "Mexico"
```
Date as value in confirmed_user_program_families dict will be used to obtain the most recent program role for the user that matches with a program on the card incase of multiple matches on the front-end

[Link to front end PR](https://github.com/masschallenge/front-end/pull/208)

### DIFF
--- a/web/impact/impact/permissions/v1_api_permissions.py
+++ b/web/impact/impact/permissions/v1_api_permissions.py
@@ -8,12 +8,9 @@ SAFE_METHODS = ['GET', 'HEAD', 'OPTIONS']
 
 
 def can_view_user_details_page(request):
-    # importing here to avoid circular import
-    from impact.v1.views.user_detail_view import UserDetailView  # noqa: E402
-    if request.resolver_match.url_name == UserDetailView.view_name:
-        user_id = request.parser_context['kwargs'].get('pk', '')
-        if user_id and request.method in SAFE_METHODS:
-            return str(user_id) == str(request.user.id)
+    user_id = request.parser_context['kwargs'].get('pk', '')
+    if user_id and request.method in SAFE_METHODS:
+        return str(user_id) == str(request.user.id)
 
 
 class V1APIPermissions(BasePermission):
@@ -21,5 +18,10 @@ class V1APIPermissions(BasePermission):
 
     def has_permission(self, request, view):
         return (is_employee(request.user) or
-                can_view_user_details_page(request) or
                 request.user.groups.filter(name=settings.V1_API_GROUP).exists())
+
+
+class UserDetailViewPermission(V1APIPermissions):
+    def has_permission(self, request, view):
+        return (super().has_permission(request, view) or
+                can_view_user_details_page(request))
--- a/web/impact/impact/permissions/v1_api_permissions.py
+++ b/web/impact/impact/permissions/v1_api_permissions.py
@@ -1,12 +1,23 @@
 from accelerator_abstract.models.base_user_utils import is_employee
 from impact.permissions import (
     settings,
-    BasePermission)
+    BasePermission,
+)
+
+SAFE_METHODS = ['GET', 'HEAD', 'OPTIONS']
+
+
+def is_object_owner(request, user_id):
+    if user_id and request.method in SAFE_METHODS:
+        return str(user_id) == str(request.user.id)
 
 
 class V1APIPermissions(BasePermission):
     authenticated_users_only = True
 
     def has_permission(self, request, view):
+        user_id = request.parser_context['kwargs'].get('pk', '')
+
         return request.user.groups.filter(
-            name=settings.V1_API_GROUP).exists() or is_employee(request.user)
+            name=settings.V1_API_GROUP
+        ).exists() or is_employee(request.user) or is_object_owner(request, user_id)
--- a/web/impact/impact/permissions/v1_api_permissions.py
+++ b/web/impact/impact/permissions/v1_api_permissions.py
@@ -7,17 +7,19 @@ from impact.permissions import (
 SAFE_METHODS = ['GET', 'HEAD', 'OPTIONS']
 
 
-def is_object_owner(request, user_id):
-    if user_id and request.method in SAFE_METHODS:
-        return str(user_id) == str(request.user.id)
+def can_view_user_details_page(request):
+    # importing here to avoid circular import
+    from impact.v1.views.user_detail_view import UserDetailView  # noqa: E402
+    if request.resolver_match.url_name == UserDetailView.view_name:
+        user_id = request.parser_context['kwargs'].get('pk', '')
+        if user_id and request.method in SAFE_METHODS:
+            return str(user_id) == str(request.user.id)
 
 
 class V1APIPermissions(BasePermission):
     authenticated_users_only = True
 
     def has_permission(self, request, view):
-        user_id = request.parser_context['kwargs'].get('pk', '')
-
-        return request.user.groups.filter(
-            name=settings.V1_API_GROUP
-        ).exists() or is_employee(request.user) or is_object_owner(request, user_id)
+        return (is_employee(request.user) or
+                can_view_user_details_page(request) or
+                request.user.groups.filter(name=settings.V1_API_GROUP).exists())
--- a/web/impact/impact/tests/contexts/user_context.py
+++ b/web/impact/impact/tests/contexts/user_context.py
@@ -11,8 +11,6 @@ from impact.tests.factories import (
     IndustryFactory,
     MemberProfileFactory,
     UserFactory,
-    ProgramRoleFactory,
-    ProgramRoleGrantFactory,
 )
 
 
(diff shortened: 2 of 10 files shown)


jonkiparsky:
@iankigen your PR request notes refer to a ticket which is not this one. Please update to describe the work done in AC-7430. Also, please update that ticket to say what you're actually doing, as there's no indication of what I'm supposed to be reviewing here.

iankigen:
@jonkiparsky done. For context, the previous functionality done using graphql only allowed experts to see the relevant program matches. This PR removes that functionality and  uses the API to deliver the same for all user types

geofrocker (approved):
Focused on the QA and it looks good

web/impact/impact/permissions/v1_api_permissions.py
```diff
@@ -1,12 +1,23 @@
...
+        return str(user_id) == str(request.user.id)
 
 
 class V1APIPermissions(BasePermission):
     authenticated_users_only = True
 
     def has_permission(self, request, view):
+        user_id = request.parser_context['kwargs'].get('pk', '')
+
         return request.user.groups.filter(
```
jonkiparsky: let's structure this a little more readably. 
Rearranging the conditions like so

```
        return (is_employee(request.user) or 
                    is_object_owner(request, user_id) or 
                    request.user.groups.filter(name=settings.V1_API_GROUP).exists())
```
avoids awkward line breaks.

web/impact/impact/permissions/v1_api_permissions.py
```diff
@@ -1,12 +1,23 @@
...
+def is_object_owner(request, user_id):
+    if user_id and request.method in SAFE_METHODS:
+        return str(user_id) == str(request.user.id)
 
 
 class V1APIPermissions(BasePermission):
     authenticated_users_only = True
 
     def has_permission(self, request, view):
+        user_id = request.parser_context['kwargs'].get('pk', '')
```
jonkiparsky: why is this called "user_id"? It looks like it's the object of any object we're querying on. Is this only ever called with a user as the object?

web/impact/impact/tests/contexts/user_context.py
```diff
@@ -44,4 +48,9 @@ def __init__(self,
         elif user_type == "MEMBER":
             self.profile = MemberProfileFactory(user=self.user)
             user.memberprofile = self.profile
+        if program:
```
jonkiparsky: Looking at the way this is used, I'd rather not add this to the context. See note on the test below.

web/impact/impact/tests/test_graphql.py
```diff
@@ -398,27 +396,3 @@ def test_non_staff_user_can_access_finalist_graphql_view(self):
                     }
                 }
             )
-
-    def test_query_expert_by_email(self):
```
jonkiparsky, line 402: Why is this test being removed?

web/impact/impact/tests/test_user_detail_view.py
```diff
@@ -628,6 +630,36 @@ def test_patch_user_with_no_profile_not_allowed_on_missing_fields(self):
             assert MISSING_PROFILE_ERROR.format(user.id) in response.data
             assert MISSING_SUBJECT_ERROR in response.data
 
+    def test_get_user_latest_active_program_location(self):
+        program = ProgramFactory(
+            program_status=ACTIVE_PROGRAM_STATUS,
+            program_family=ProgramFamilyFactory(),
```
jonkiparsky: ProgramFamily will be provided by the ProgramFactory, only needs to be specified if you have a particular Family in mind

web/impact/impact/tests/test_user_detail_view.py
```diff
@@ -628,6 +630,36 @@ def test_patch_user_with_no_profile_not_allowed_on_missing_fields(self):
...
+        )
+        context = UserContext(user_type=EXPERT_USER_TYPE, program=program)
```
jonkiparsky: Rather than adding the program role grant in the context, let's do it here:
```
role_grant = ProgramRoleGrantFactory(program_role__program_status=ACTIVE_PROGRAM_STATUS)
```
Then we can get the user_program from the role_grant:
```
...
user_program=role_grant.program_role.program
...
``` 
(line 643)

web/impact/impact/tests/test_user_detail_view.py
```diff
@@ -628,6 +630,36 @@ def test_patch_user_with_no_profile_not_allowed_on_missing_fields(self):
...
+        user = context.user
+        with self.login(email=self.basic_user().email):
+            url = reverse(UserDetailView.view_name, args=[user.id])
+            response = self.client.get(url)
+            user_program = user.programrolegrant_set.order_by(
+                '-created_at'
+            ).first().program_role.program
+            program_location = response.data['latest_active_program_location']
+            assert user_program.program_family.name == program_location
```
jonkiparsky, line 643: This smells funny. Shouldn't we be testing program family location against the received location instead?

web/impact/impact/tests/test_user_detail_view.py
```diff
@@ -628,6 +630,36 @@ def test_patch_user_with_no_profile_not_allowed_on_missing_fields(self):
...
+
+    def test_get_user_confirmed_program_families(self):
+        program = ProgramFactory(
+            program_status=ACTIVE_PROGRAM_STATUS,
+            program_family=ProgramFamilyFactory(),
+        )
+        context = UserContext(user_type=EXPERT_USER_TYPE, program=program)
```
jonkiparsky: see suggested change above

jonkiparsky (dismissed):
Some code cleanup suggestions

web/impact/impact/permissions/v1_api_permissions.py
```diff
@@ -1,12 +1,23 @@
...
     settings,
-    BasePermission)
+    BasePermission,
+)
+
+SAFE_METHODS = ['GET', 'HEAD', 'OPTIONS']
+
+
+def is_object_owner(request, user_id):
+    if user_id and request.method in SAFE_METHODS:
```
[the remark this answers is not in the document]
iankigen: @jonkiparsky I added that to allow a logged-in user to be able to access their user detail page which was previously restricted to employees and users in v1 api group

web/impact/impact/permissions/v1_api_permissions.py
iankigen: On it

web/impact/impact/permissions/v1_api_permissions.py
iankigen: I should probably make sure that the user is on the user detail page before getting that.

web/impact/impact/tests/contexts/user_context.py
iankigen: Didn't have context on that. Thanks for the suggestion

web/impact/impact/tests/test_graphql.py
iankigen: I removed this because we no longer have the functionality to get query by email on expert profile(graphql)

web/impact/impact/tests/test_user_detail_view.py
iankigen: `program_location = response.data['latest_active_program_location']` is the received location which i'm testing against

web/impact/impact/permissions/v1_api_permissions.py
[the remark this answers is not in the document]
jonkiparsky: Okay. Maybe the name should change to reflect the limited use case? `is_object_owner` looks like it's checking whether the current user is the owner of some arbitrary object. 

I'm not sure what the right name is at this point, to be honest.

iankigen:
@jonkiparsky, PR ready for review.

web/impact/impact/permissions/v1_api_permissions.py
```diff
@@ -1,12 +1,25 @@
...
+    BasePermission,
+)
+
+SAFE_METHODS = ['GET', 'HEAD', 'OPTIONS']
+
+
+def can_view_user_details_page(request):
+    # importing here to avoid circular import
+    from impact.v1.views.user_detail_view import UserDetailView  # noqa: E402
+    if request.resolver_match.url_name == UserDetailView.view_name:
```
jonkiparsky: Because there is no else clause on this statement, this function will return a None (which is falsey) for all views that are not UserDetailView. This is probably not what you want. :)

Since this permission is specific to exactly one view, I would suggest that it should be implemented locally on that view. 
If you prefer, it would be reasonable to define a permission function here and import it into UserDetailView. Let me know if you want more concrete implementation ideas

iankigen: @jonkiparsky that is intentional, as I want the function to be default falsey in all the other views so that the permission check is determined by `is_employee(request.user)` and `request.user.groups.filter(name=settings.V1_API_GROUP).exists()` as `can_view_user_details_page(request)` will be False in those cases

iankigen: However, I like the idea of implementing it locally on the view

jonkiparsky: yes, that's a very confusing pattern to me. let's bring it into the view

freshprincekla (approved):
Code changes look good to me.

PR #372, opened 2020-02-15
[AC-7450] P1: v0 and V1 API permissions failure

philophilo:
### Changes introduced in: [AC-7450](https://masschallenge.atlassian.net/browse/AC-7450):
- Converts permission value for `v0` and `v1` endpoints to string

### How to test
- Authenticate follow these instructions https://github.com/masschallenge/impact-api/blob/development/api.md#register-your-use-of-the-api
- Required data https://github.com/masschallenge/impact-api/blob/development/api.md#register-your-use-of-the-api

on development:
- send a `POST` request to http://localhost:8000/api/v0/startup_list/ notice that it returns the 403 response
```
{
    "detail": "You do not have permission to perform this action."
}
```
on this branch:
- send a `POST` request to http://localhost:8000/api/v0/startup_list/ and notice that a list of startups is returned

### DIFF
--- a/web/impact/impact/settings.py
+++ b/web/impact/impact/settings.py
@@ -197,8 +197,8 @@ class Base(Configuration):
     V0_SITE_NAME = bytes(os.environ.get(
         'IMPACT_API_V0_SITE_NAME', 'masschallenge.org'), 'utf-8')
 
-    V0_API_GROUP = bytes(os.environ.get(
-        'IMPACT_API_V0_API_GROUP', 'v0_clients'), 'utf-8')
+    V0_API_GROUP = os.environ.get(
+        'IMPACT_API_V0_API_GROUP', 'v0_clients')
 
     # This and the above should get generalized.  See AC-4574.
     V1_API_GROUP = bytes(os.environ.get(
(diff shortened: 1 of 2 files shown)


shakirandagire:
Doing QA for the PR

web/impact/impact/settings.py
```diff
@@ -197,8 +197,8 @@ class Base(Configuration):
     V0_SITE_NAME = bytes(os.environ.get(
         'IMPACT_API_V0_SITE_NAME', 'masschallenge.org'), 'utf-8')
 
-    V0_API_GROUP = bytes(os.environ.get(
-        'IMPACT_API_V0_API_GROUP', 'v0_clients'), 'utf-8')
+    V0_API_GROUP = os.environ.get(
```
geofrocker, line 200: The changes look good. Did we get to know what is causing this(`bytes`) to fail now and why is it still working for V0_SITE_NAME and V1_API_GROUP?

philophilo: I am not certain why in the first place it was changed to bytes but under v1 if the user is an employee, then the permission is granted which is not given for v0. Sorry, pushing a commit to update v1 as well.

shakirandagire (approved):
QA looks good to me.
What I did:
- I followed the instructions and registered the use of the API
- I copied the clientid and logged into the impact using (username, password, clientid) on the endpoint http://localhost:8000/oauth/token/ 
- I visited http://localhost:8000/api/v0/startup_list/ and posted the startups. and didnot get "You do not have permission to perform this action." error

geofrocker (approved):
Code looks good

PR #382, opened 2020-02-28
[AC-7504] Error on /v1/user when PRG has null created_at

kaymation:
[JIRA](https://masschallenge.atlassian.net/browse/AC-7504)

older users have null `created_at`s in their PRGs. 
Since we use this value for comparison, it creates an error for some users

To test
- on development
- Masquerade as a user with PRGs with null `created_at`'s e.g `shankar@masschallenge.org` if you have a recent prod dump
- visit localhost:8000/api/v1/user/{
- should be greated with a 500 error
- check out this branch
- error should be resolved

### DIFF
--- a/web/impact/impact/tests/test_user_detail_view.py
+++ b/web/impact/impact/tests/test_user_detail_view.py
@@ -651,8 +651,12 @@ class TestUserDetailView(APITestCase):
         with self.login(email=self.basic_user().email):
             url = reverse(UserDetailView.view_name, args=[user.id])
             response = self.client.get(url)
-            confirmed_program_families = response.data['confirmed_user_program_families']
-            assert user_program.program_family.name in confirmed_program_families.keys()
+            confirmed_program_families = response.data[
+                'confirmed_user_program_families'
+            ]
+
+            keys = confirmed_program_families.keys()
+            assert user_program.program_family.name in keys
 
     def test_get_user_with_no_created_at_in_prgs_handled(self):
         role_grant = ProgramRoleGrantFactory(
@@ -675,12 +679,15 @@ class TestUserDetailView(APITestCase):
 
         role_grant.save()
         role_grant2.save()
-        
+
         with self.login(email=self.basic_user().email):
             url = reverse(UserDetailView.view_name, args=[user.id])
             response = self.client.get(url)
-            confirmed_program_families = response.data['confirmed_user_program_families']
-            assert user_program.program_family.name in confirmed_program_families.keys()
+            confirmed_program_families = response.data[
+                'confirmed_user_program_families'
+            ]
+            keys = confirmed_program_families.keys()
+            assert user_program.program_family.name in keys
 
 
 def _valid_note(messages):
--- a/web/impact/impact/tests/test_user_detail_view.py
+++ b/web/impact/impact/tests/test_user_detail_view.py
@@ -662,7 +662,6 @@ class TestUserDetailView(APITestCase):
         role_grant = ProgramRoleGrantFactory(
             program_role__program__program_status=ACTIVE_PROGRAM_STATUS,
             program_role__program__program_family__name="Quigville",
-            created_at=None
         )
         user = role_grant.person
         user_program = role_grant.program_role.program
@@ -671,7 +670,6 @@ class TestUserDetailView(APITestCase):
             person=user,
             program_role__program__program_status=ACTIVE_PROGRAM_STATUS,
             program_role__program__program_family__name="Quigville",
-            created_at=None
         )
 
         role_grant.created_at = None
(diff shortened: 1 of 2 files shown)


web/impact/impact/v1/helpers/profile_helper.py
```diff
@@ -369,11 +369,13 @@ def latest_distinct_program_families_dict(program_families):
     program_families_dict = {}
     for program_family in program_families:
         location, created_at = program_family[0], program_family[1]
+        user_created_at = program_family[2]
         if location in program_families_dict.keys():
+            created_at = (created_at or user_created_at)
```
shanbady: could set the created_at variable up a level and eliminate the need for comparing ```(created_at or user_created_at)``` again in the else statement?

shanbady (changes requested):

web/impact/impact/v1/helpers/profile_helper.py
```diff
@@ -506,10 +508,15 @@ def is_non_member(self):
     def confirmed_user_program_families(self):
         prg = _confirmed_non_future_program_role_grant(self.subject)
         program_ids = _latest_program_id_foreach_program_family()
-        program_families = list(prg.filter(
-            program_role__program__pk__in=program_ids
-        ).values_list(
-            'program_role__program__program_family__name', 'created_at'))
+        program_families = list(
```
pbx, line 516: The reason we are including `person__date_joined` is non-obvious to the average observer. I think a short comment at the top of this method would be a kindness. E.g. 
`# We include person__date_joined as a fallback for legacy PRGs with no date`

Also, this is bikeshedding but `program_family_data` would be a less misleading name for the working variable, since it now contains non-program-family data.  (Likewise `confirmed_user_program_data` would be an improved name for this method.)

pbx (changes requested):
Thanks for fixing this! A couple very minor change requests and some bikeshedding.

shanbady (approved):

web/impact/impact/v1/helpers/profile_helper.py
```diff
@@ -514,10 +513,12 @@ def confirmed_user_program_families(self):
             ).values_list(
                 'program_role__program__program_family__name',
                 'created_at',
+                # We include person__date_joined as a
```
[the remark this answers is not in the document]
jonkiparsky: Actually is there any reason we're not using the `date_joined` attribute of the user? That seems to make more sense than looking it up from program family created dates.

web/impact/impact/tests/test_user_detail_view.py
```diff
@@ -662,7 +662,6 @@ def test_get_user_with_no_created_at_in_prgs_handled(self):
         role_grant = ProgramRoleGrantFactory(
             program_role__program__program_status=ACTIVE_PROGRAM_STATUS,
             program_role__program__program_family__name="Quigville",
```
jonkiparsky: no need to specify a program_family__name here, please remove

web/impact/impact/tests/test_user_detail_view.py
```diff
@@ -671,7 +670,6 @@ def test_get_user_with_no_created_at_in_prgs_handled(self):
             person=user,
             program_role__program__program_status=ACTIVE_PROGRAM_STATUS,
             program_role__program__program_family__name="Quigville",
```
jonkiparsky: no need to specify a program_family__name here, please remove

jonkiparsky (changes requested):

web/impact/impact/v1/helpers/profile_helper.py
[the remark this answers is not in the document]
pbx: @jonkiparsky  - I'm answering for Kevin since I've taken the ticket over in his absence -- the `created_at` here is from the `ProgramRoleGrant` objects, not the `ProgramFamily`.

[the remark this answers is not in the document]
pbx: Made the change.

web/impact/impact/tests/test_user_detail_view.py
pbx: Done

web/impact/impact/tests/test_user_detail_view.py
pbx: Done

web/impact/impact/v1/helpers/profile_helper.py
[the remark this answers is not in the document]
pbx: Made the switch to `values`, good readability improvement.

pbx (approved):

jonkiparsky (approved):
looks good to me

PR #388, opened 2020-03-19
[AC-7542] Ability to filter /api/v1/user/ by date modified

kaymation:
[JIRA](https://masschallenge.atlassian.net/browse/AC-7542)

Turns out we already have this functionality, so this PR just adds a test to verify the overall functionality requested by the ticket indeed operated correctly 

To Test:
- navigate to `/api/v1/user`
- add `?updated_at.after=2016-04-23T18:25:43.511Z`(or any proper ISO 8601 date)
- * should only report users modified after this date
- replace the parameter with `?updated_at.before=2016-04-23T18:25:43.511Z`(or any proper ISO 8601 date)
- * should only report users modified before this date
- now, try adding `?updated_at.after=2016-04-23T18:25:43.511Z&?updated_at.before=2019-04-23T18:25:43.511Z`(or any two proper iso dates where the latter is more recent than former)
- * should only report users modified between these two dates

### DIFF
--- a/web/impact/impact/tests/test_user_list_view.py
+++ b/web/impact/impact/tests/test_user_list_view.py
@@ -498,7 +498,7 @@ class TestUserListView(APITestCase):
         updated_exactly = _user_for_date(week_ago)
         updated_after = _user_for_date(week_ago + one_day)
         with self.login(email=self.basic_user().email):
-            url = "{base_url}?updated_at.after={datestr}".format(
+            url = "{base_url}?date_modified__gt={datestr}".format(
                 base_url=self.url,
                 datestr=week_ago.strftime("%Y-%m-%dT%H:%M:%S.%fZ"))
             response = self.client.get(url)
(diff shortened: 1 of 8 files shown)


web/impact/impact/tests/test_user_list_view.py
```diff
@@ -507,6 +507,21 @@ def test_updated_at_after_datetime_filter(self):
             assert _contains_user(updated_exactly, response.data)
             assert _contains_user(updated_after, response.data)
 
+    def test_updated_at_filter_when_same_day(self):
```
jonkiparsky, line 510: drive-by comment: there is a lot of duplication in these tests. While we're here, it would be great if we could clean that up and make it pretty.

geofrocker:
will do QA

geofrocker (commented):
I am not seeing the parameter names that have been updated. Only seeing a test. I am I missing something? @kaymation 
>so this PR just updates the parameter names with those specified in the ticket

geofrocker (approved):
Code and QA look good

philophilo (approved):
Tests look good ✅ 
Date formats make the query ✅

PR #394, opened 2020-04-08
[AC-7570] Add minimal mail handler to impact-api

jonkiparsky:
https://masschallenge.atlassian.net/browse/AC-7570

Review notes:
This is prep work for office hours api endpoints that will need to send emails. It's a thin wrapper on the EmailMultiAlternatives provided by django, and does not add much to that. 

Difficult to QA since I have not been able to get it deployed successfully to test4 yet, but working on that.

### DIFF
--- a/web/impact/impact/minimal_email_handler.py
+++ b/web/impact/impact/minimal_email_handler.py
@@ -1,0 +1,33 @@
+from django.conf import settings
+from django.core.mail import EmailMultiAlternatives
+
+class MinimalEmailHandler:
+    '''
+    A simplistic outgoing email handler, to consolidate outgoing
+    mail sending needs of impact-api calls
+    attachment, if any, should be a triple of (filename, content, filetype)
+    attach_alternative, if any , should be 
+    '''
+    def __init__(self,
+                 to,
+                 subject,
+                 body,
+                 from_email=None,
+                 bcc=None,
+                 attachment=None,
+                 attach_alternative=None):
+        self.email = EmailMultiAlternatives(
+            subject,
+            body,
+            to=to,
+            bcc=bcc or [settings.BCC_EMAIL],
+            from_email = from_email or [settings.NO_REPLY_EMAIL])
+        if attachment:
+            self.email.attach(*attachment)
+        if attach_alternative:
+            self.email.attach_alternative(*attach_alternative)
+    
+    def send(self):
+        self.email.send()
+        
+            
--- a/web/impact/impact/minimal_email_handler.py
+++ b/web/impact/impact/minimal_email_handler.py
@@ -1,12 +1,14 @@
 from django.conf import settings
 from django.core.mail import EmailMultiAlternatives
 
+
 class MinimalEmailHandler:
     '''
     A simplistic outgoing email handler, to consolidate outgoing
     mail sending needs of impact-api calls
     attachment, if any, should be a triple of (filename, content, filetype)
-    attach_alternative, if any, should be a tuple of (attachment_text, MIME_type)
+    attach_alternative, if any, should be a tuple of
+      (attachment_text, MIME_type)
       (generally used for html version of the email)
     '''
     def __init__(self,
@@ -22,13 +24,11 @@ class MinimalEmailHandler:
             body,
             to=to,
             bcc=bcc or [settings.BCC_EMAIL],
-            from_email = from_email or [settings.NO_REPLY_EMAIL])
+            from_email=from_email or [settings.NO_REPLY_EMAIL])
         if attachment:
             self.email.attach(*attachment)
         if attach_alternative:
             self.email.attach_alternative(*attach_alternative)
-    
+
     def send(self):
         self.email.send()
-        
-            
--- a/web/impact/impact/minimal_email_handler.py
+++ b/web/impact/impact/minimal_email_handler.py
@@ -6,7 +6,8 @@ class MinimalEmailHandler:
     A simplistic outgoing email handler, to consolidate outgoing
     mail sending needs of impact-api calls
     attachment, if any, should be a triple of (filename, content, filetype)
-    attach_alternative, if any , should be 
+    attach_alternative, if any, should be a tuple of (attachment_text, MIME_type)
+      (generally used for html version of the email)
     '''
     def __init__(self,
                  to,
--- a/web/impact/impact/settings.py
+++ b/web/impact/impact/settings.py
@@ -361,7 +361,9 @@ class Base(Configuration):
             },
         }
     }
-
+    NO_REPLY_EMAIL = "noreply@masschallenge.org"
+    BCC_EMAIL = "systemlogbcc@masschallenge.org"
+    EMAIL_BACKEND = "mc.email_backends.AccelerateEmailBackend"
 
 class Dev(Base):
     DEBUG = True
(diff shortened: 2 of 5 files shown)


pbx:
Taking code review duties.

web/impact/impact/minimal_email_handler.py
```diff
@@ -0,0 +1,33 @@
+from django.conf import settings
+from django.core.mail import EmailMultiAlternatives
+
+class MinimalEmailHandler:
+    '''
+    A simplistic outgoing email handler, to consolidate outgoing
+    mail sending needs of impact-api calls
+    attachment, if any, should be a triple of (filename, content, filetype)
+    attach_alternative, if any , should be
```
pbx: Looks like this line didn't get finished!

web/impact/impact/minimal_email_handler.py
```diff
@@ -0,0 +1,33 @@
+from django.conf import settings
+from django.core.mail import EmailMultiAlternatives
+
+class MinimalEmailHandler:
+    '''
+    A simplistic outgoing email handler, to consolidate outgoing
```
pbx, line 7: Thanks a ton for adding docstrings. Warms my heart.

web/impact/impact/settings.py
```diff
@@ -361,7 +361,9 @@ class Base(Configuration):
             },
         }
     }
-
+    NO_REPLY_EMAIL = "noreply@masschallenge.org"
+    BCC_EMAIL = "systemlogbcc@masschallenge.org"
+    EMAIL_BACKEND = "mc.email_backends.AccelerateEmailBackend"
```
pbx: The Python path in this string will not be reachable from impact. Did you mean to reference the new `ImpactEmailBackend`?

web/impact/impact/tests/test_minimal_mail_handler.py
```diff
@@ -0,0 +1,43 @@
+from django.test import TestCase
+from django.core import mail
+from impact.minimal_email_handler import MinimalEmailHandler
+
+
+DEFAULT_PARAM_KEYS = ("to",
+                      "subject",
+                      "body")
+
+DEFAULT_PARAM_VALS = (["a@b.com"],
```
pbx: I prefer using `@example.com` addresses rather than potentially deliverable ones.

pbx (changes requested):
Thanks for picking this up. I have a few small change requests. Also: do you think there should be a test for failure-to-send?

web/impact/impact/settings.py
pbx: (A second question comes to mind -- the fact that this bad setting didn't foil the test suggests maybe the test suite has a hole in it? I.e. if somebody were to break email sending by messing up `EMAIL_BACKEND` in the future, can we have a test that would alert us?)

jonkiparsky: Thanks for catching that. 
Not sure what sort of test that would be. Any suggestions?

pbx: I don't know offhand - if I'm correct in what I think we saw here, having an unusable `EMAIL_BACKEND` value did not cause the `EmailMultiAlternatives.send()` method to throw an error; this may be due to the monkeypatching that happens with email under the test runner, I don't know.

Alternative angle: let's make sure we have a test that we know exercises `ImpactEmailBackend`.

jonkiparsky: Will try to come up with a test that exercises the backend... it's a little tricky though...

iankigen:
Fixing failing tests on impact.
Missed it while working on AC-7365

kaymation:
I would prefer to see the email handle offer a more straightforward implemention, since not having html/text distinguished correctly has burned us in the past

```python
class MinimalEmailHandler:
  def __init__(self,
                 to,
                 subject,
                 body=None,
                 body_html=None,
                 from_email=None,
                 bcc=None,
                 attachment=None,):
        if body is None and body_html is None:
            return SomeException
        self.email = EmailMultiAlternatives(
            subject,
            body,
            to=to,
            bcc=bcc or [settings.BCC_EMAIL],
            from_email = from_email or [settings.NO_REPLY_EMAIL])
        if attachment:
            self.email.attach(*attachment)
        if body_html:
            self.email.attach_alternatives(body_html, "text/html")
```
We are almost always going to want an html option for emails. This way, we have a straightforward implementation of sending text and HTML, instead of depending on future engineers knowing the inner workings the EmailMultiAlternative class, which would make the handler a bit redundant IMO

kaymation (approved):

pbx (changes requested):
Test suite looks good to me. Can we round it out with one that checks `mocked_backend.called`?

Oh, and the backend setting in the tests is causing `ModuleNotFoundError: No module named 'mc'`

web/impact/impact/settings.py
```diff
@@ -361,7 +361,10 @@ class Base(Configuration):
             },
         }
     }
-
+    NO_REPLY_EMAIL = "noreply@masschallenge.org"
+    BCC_EMAIL = "systemlogbcc@masschallenge.org"
+    EMAIL_BACKEND = "impact.email_backends.ImpactEmailBackend"
```
pbx: This looks wrong. There's no `email_backends` module or directory.

web/impact/impact/tests/test_impact_email_backend.py
```diff
@@ -0,0 +1,63 @@
+from mock import patch
+from django.core import mail
+from django.test import TestCase
+from django.urls import reverse
+
+from impact.minimal_email_handler import MinimalEmailHandler
+
+
+IMPACT_BACKEND_PATH = 'impact.impact_email_backend.ImpactEmailBackend'
```
pbx, line 7: Why are we using this value instead of getting it from `settings.EMAIL_BACKEND`?

pbx (changes requested):
Some questions about the `*_BACKEND` values.

web/impact/impact/tests/test_impact_email_backend.py
jonkiparsky: Good thought

jonkiparsky: @pbx I attempted to make this change, but it caused the tests to fail. How critical is this for you? I will spend more time on it if you think it's important that I do so.

pbx: Sorry it's been so time-consuming, but yeah, I do think it's important.  I like the idea of a minimal email handler that will serve as the basis for future expansion, but I want to make sure it's right and that it's defended against accidental breakage later. Thanks.

pbx (changes requested):
Per our discussion I've groomed AC-7670 (the follow-on to this ticket, to handle testing).  The only remaining thing to do here that I can see: the `IMPACT_BACKEND_PATH` value is now identical to the one in `settings.py`, so we might as well replace it with a reference to `settings.EMAIL_BACKEND`. Right?

jonkiparsky:
@pbx We'd like to think that would be possible. However, that path is identical to the value in production settings, and what we see when this test runs are the test settings, which do not include that value.

pbx:
Thanks @jonkiparsky - then let's add a comment to `IMPACT_BACKEND_PATH` explaining why it's set the way it is, because it would be easy for anyone to see the duplication and (erroneously, it seems) replace that local value with an import of `settings.EMAIL_BACKEND`. Mentioning AC-7670 (where we presume this issue will be resolved) in the comment would probably be helpful too.

pbx (approved):

PR #396, opened 2020-04-15
[AC 7642] P1: Users unable to access some people directory profiles

iankigen:
## [AC 7642](https://masschallenge.atlassian.net/browse/AC-7642)

**Changes introduced**
- Implement access control for people directory profiles

**Testing**
- Check out this branch 
- Login or masquerade as a user with the current user role of [Staff, Finalist, Alumni in residence, Mentor, Partner, Alum or No user role]
- Visit the people directory pages: `/people/<user_id>` of the users they can view and confirm that access control is as defined in the table below:
Screenshot 2020-04-15 at 17 19 27

### DIFF
--- a/web/impact/impact/graphql/query.py
+++ b/web/impact/impact/graphql/query.py
@@ -1,42 +1,21 @@
 import graphene
+from graphql import GraphQLError
 
-from impact.graphql.types import (
-    EntrepreneurProfileType,
-    ExpertProfileType,
-)
 from accelerator.models import (
     EntrepreneurProfile,
-    ExpertProfile,
-    UserRole,
+    ExpertProfile
 )
-from accelerator_abstract.models.base_user_utils import is_employee
-from graphql import GraphQLError
+from impact.graphql.types import (
+    EntrepreneurProfileType,
+    ExpertProfileType
+)
+from impact.permissions.graphql_permissions import (
+    can_view_entrepreneur_profile
+)
 
 ENTREPRENEUR_NOT_FOUND_MESSAGE = 'Entrepreneur matching the id does not exist.'
 EXPERT_NOT_FOUND_MESSAGE = 'Expert matching the id does not exist.'
 NOT_ALLOWED_ACCESS_MESSAGE = 'Sorry, You are not allowed to access this page.'
-
-
-def visible_roles(current_user):
-    basic_user_roles = [
-        UserRole.FINALIST,
-        UserRole.AIR,
-        UserRole.MENTOR,
-        UserRole.PARTNER,
-        UserRole.ALUM
-    ]
-    basic_visible_roles = [UserRole.FINALIST, UserRole.STAFF, UserRole.ALUM]
-
-    current_user_roles = list(
-        current_user.programrolegrant_set.values_list(
-            'program_role__user_role__name', flat=True).distinct()
-    )
-    if not current_user_roles:
-        return [UserRole.STAFF]
-    if set(basic_user_roles).intersection(current_user_roles):
-        return basic_visible_roles + [UserRole.MENTOR]
-    if UserRole.JUDGE in current_user_roles:
-        return basic_visible_roles
 
 
 def has_permission(profile_user, roles):
@@ -66,9 +45,8 @@ class Query(graphene.ObjectType):
         if not entrepreneur:
             return GraphQLError(ENTREPRENEUR_NOT_FOUND_MESSAGE)
 
-        if not is_employee(info.context.user):
-            roles = visible_roles(info.context.user)
-            if not has_permission(entrepreneur.user, roles):
-                return GraphQLError(NOT_ALLOWED_ACCESS_MESSAGE)
+        if not can_view_entrepreneur_profile(info.context.user,
+                                             entrepreneur.user):
+            return GraphQLError(NOT_ALLOWED_ACCESS_MESSAGE)
 
         return entrepreneur
--- a/web/impact/impact/graphql/query.py
+++ b/web/impact/impact/graphql/query.py
@@ -7,13 +7,41 @@ from impact.graphql.types import (
 from accelerator.models import (
     EntrepreneurProfile,
     ExpertProfile,
+    UserRole,
 )
-from accelerator_abstract.models.base_user_role import is_finalist_user
 from accelerator_abstract.models.base_user_utils import is_employee
 from graphql import GraphQLError
+
 ENTREPRENEUR_NOT_FOUND_MESSAGE = 'Entrepreneur matching the id does not exist.'
 EXPERT_NOT_FOUND_MESSAGE = 'Expert matching the id does not exist.'
-NON_FINALIST_PROFILE_MESSAGE = 'Sorry, You are not allowed to access this page.'
+NOT_ALLOWED_ACCESS_MESSAGE = 'Sorry, You are not allowed to access this page.'
+
+
+def visible_roles(current_user):
+    basic_user_roles = [
+        UserRole.FINALIST,
+        UserRole.AIR,
+        UserRole.MENTOR,
+        UserRole.PARTNER,
+        UserRole.ALUM
+    ]
+    basic_visible_roles = [UserRole.FINALIST, UserRole.STAFF, UserRole.ALUM]
+
+    current_user_roles = list(
+        current_user.programrolegrant_set.values_list(
+            'program_role__user_role__name', flat=True).distinct()
+    )
+    if not current_user_roles:
+        return [UserRole.STAFF]
+    if set(basic_user_roles).intersection(current_user_roles):
+        return basic_visible_roles + [UserRole.MENTOR]
+    if UserRole.JUDGE in current_user_roles:
+        return basic_visible_roles
+
+
+def has_permission(profile_user, roles):
+    return profile_user.programrolegrant_set.filter(
+        program_role__user_role__name__in=roles).exists()
 
 
 class Query(graphene.ObjectType):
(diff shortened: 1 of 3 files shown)


web/impact/impact/graphql/query.py
```diff
@@ -7,13 +7,41 @@
...
+    basic_user_roles = [
+        UserRole.FINALIST,
+        UserRole.AIR,
+        UserRole.MENTOR,
+        UserRole.PARTNER,
+        UserRole.ALUM
+    ]
+    basic_visible_roles = [UserRole.FINALIST, UserRole.STAFF, UserRole.ALUM]
+
+    current_user_roles = list(
```
jonkiparsky: Why are we making this into a list here?

If we leave it as a queryset, then we do not need to convert it back into a set on line 36

web/impact/impact/graphql/query.py
```diff
@@ -7,13 +7,41 @@
...
+        current_user.programrolegrant_set.values_list(
```
jonkiparsky: This query does not get current user roles, it gets all user roles that this user holds.

Presumably you could fix this by filtering `user.programrolegrant_set` on program_role.program.program_status

Probably we'd want to write a test about this, also, since it seems like an easy mistake to make. 
test should ensure that a user with a certain user role for an inactive program does not get to see things that only active grant holders should see.

web/impact/impact/tests/test_graphql.py
```diff
@@ -582,10 +548,98 @@ def test_query_for_user_without_confirmed_mentor_program_families(self):
...
+                            data={'query': query}) as (response, _):
+            error_messages = [x['message']
+                              for x in response.json()['errors']]
+        self.assertIn(error_message, error_messages)
+
+    def _expert_user(self, role=None):
+        user = ExpertFactory()
+        if role:
+            user_role = get_user_role_by_name(role)
+            program_role = ProgramRoleFactory.create(user_role=user_role)
```
jonkiparsky: we do not need to use - and prefer not to use - the `.create` syntax, since `ProgramRoleFactory(user_role=user_role)` means and does exactly the same thing.

jonkiparsky (changes requested):
Some changes requested

freshprincekla (approved):
QA checks out. 
- Was able to login as the various users and verify that they have access to the correct profiles.

web/impact/impact/graphql/query.py
iankigen: I'm making `current_user_roles` a list so that I can perform a set intersection operation on `basic_user_roles`.
I'm converting `basic_user_roles` to set, not `current_user_roles`

iankigen:
Working on feedback

iankigen:
@jonkiparsky PR ready for a second review

web/impact/impact/tests/test_graphql.py
```diff
@@ -582,10 +551,144 @@ def test_query_for_user_without_confirmed_mentor_program_families(self):
...
+
+    def test_allowed_user_can_view_mentor_entrepreneur_profile(self):
+        allowed_user_roles = [
+            UserRole.FINALIST,
+            UserRole.AIR,
+            UserRole.MENTOR,
+            UserRole.PARTNER,
+            UserRole.ALUM
+        ]
+        for role in allowed_user_roles:
```
jonkiparsky: Not a blocker for this ticket, but if we have multiple assertions in a single test (including assertions in a loop) this indicates that this is testing multiple things and therefore should be multiple tests. 

In future, please write individual tests for individual assertions. (and if you see tests like this, please fix them!)

web/impact/impact/tests/test_graphql.py
```diff
@@ -582,10 +551,144 @@ def test_query_for_user_without_confirmed_mentor_program_families(self):
...
+
+    def test_user_cannot_view_profile_with_non_current_allowed_roles(self):
+        allowed_user = self._expert_user(UserRole.FINALIST)
+        with self.login(email=allowed_user.email):
+            user = EntrepreneurFactory()
+            UserRoleContext(
+                UserRole.MENTOR,
+                user=user,
+                program=ProgramFactory(program_status=ENDED_PROGRAM_STATUS))
+            query = """
```
jonkiparsky, line 689: It looks like this repeated definition indicates that we have a common pattern which should have been extracted. Hopefully this will be caught and cleaned up the next time we touch this test case, not going to block on this today.

web/impact/impact/permissions/graphql_permissions.py
```diff
@@ -0,0 +1,43 @@
+from accelerator.models import (
+    UserRole,
+)
+from accelerator_abstract.models.base_user_utils import is_employee
+from accelerator.models import ACTIVE_PROGRAM_STATUS
+
+
+def visible_roles(current_user):
```
jonkiparsky: This function is longer than our standards call for. 
In future, consider taking the following steps to get functions under 10 lines

- declare constants outside of functions (this is a benefit, since it allows them to be referred to in tests)
- extract complex queries to helper functions (allowing them to be re-used, and also granting them a helpful name)

web/impact/impact/permissions/graphql_permissions.py
```diff
@@ -0,0 +1,43 @@
...
+
+    current_logged_in_user_roles = list(
+        current_user.programrolegrant_set.filter(
+            program_role__program__program_status=ACTIVE_PROGRAM_STATUS
+        ).values_list(
+            'program_role__user_role__name', flat=True).distinct()
+    )
+    if not current_logged_in_user_roles:
+        return [UserRole.STAFF]
+    if set(basic_user_roles).intersection(current_logged_in_user_roles):
```
jonkiparsky: In future, consider using a construction like

```
if any([role in basic_user_roles for role in current_logged_in_user_roles]):
   ....
```
instead of set intersections, to help readability. This will allow the reader to focus more directly on the logic.

jonkiparsky (commented):
@iankigen @hfreund QA note: on cleandb, I signed in as user 76104-user@example.com, who has a Judge program role grant in an active program. This user was unable to access http://localhost:1234/people/

Therefore I was unable to verify that this user was able to see the correct users.

iankigen:
@jonkiparsky this ticket is only focused on the people profile page, I believe since [http://localhost:1234/people/](http://localhost:1234/people/) is not the only entry point for the people profile pages.

jonkiparsky:
@iankigen If you can provide better QA directions, that would be helpful. I would have assumed that access to the people directory meant access to  /people/ - is this not so?

jonkiparsky:
@iankigen working as user 15-user@example.com (no user roles at all) I was able to access http://localhost:1234/directory/people/20863, which is a non-staff user. This seems inconsistent with the QA chart.

jonkiparsky (changes requested):
QA fail: working as user 15-user@example.com (no user roles at all) I was able to access http://localhost:1234/directory/people/20863, which is a non-staff user. This is inconsistent with the QA chart, which says that users with no roles should only be able to access staff cards.

iankigen:
Resolving conflicts

hfreund:
@iankigen can you deploy this to a test machine so I can run through some QA as well?

jonkiparsky (commented):
I have been unable to do a satisfactory QA pass. This work has been deployed to a test machine for Product to review, will approve when Hugh gives the okay

hfreund:
@iankigen while doing QA on test2 I found the following problems:
Masquerading as https://test2.masschallenge.org/admin/simpleuser/user/49509/masquerade/
1. I removed the PRG for confirmed mentor so the user had only confirmed judge roles
2. Was unable to view https://test2.masschallenge.org/people/18223/ who is a 2017 Finalist for Boston (as a superuser I am able to view this user)
3. Added a confirmed mentor role grant to user 49509 and attempted again to view https://test2.masschallenge.org/people/18223/; was again unsuccessful
4. I was successfully able to view https://test2.masschallenge.org/people/78199/ who is an **alum.**
It appears users with the current program Global Alumni will appear in the People Directory filter for Global Alumni even if they do not have the Alum role grant. This seems like a separate issue but may have contributed to some of the challenges in QAing this ticket (the profiles listed on the original ticket all appear to have this data issue).

Masquerading as https://test2.masschallenge.org/admin/simpleuser/user/1277/masquerade/ which is a member profile type and has only partner admin role grants I was unable to view  https://test2.masschallenge.org/people/18223/ which is the expected outcome, although based on the failures detailed above I'm not confident if this block is due to an appropriate measure in the code or by some accident that may also be causing the access issue detailed in points 103 above (possibly this is because 18223 is a user in a program that is ended and therefore no one can see them, except super users). If that is the case the issue is a data one: user 18223 was not tagged as a global alum.

While still masquerading as 1277, I can navigate directly to https://test2.masschallenge.org/people/ (unclear if this should be possible since a member profile with a partner-admin user role does not have a defined visibility setting) however when I click on any profile I get an error.

@jonkiparsky I don't have strong feelings against merging this code since it does not appear to create undesired outcomes, so if you want to merge it please do so. However, this additional QA has revealed more problems and unsupported use-cases that should be addressed in follow on work:
1. Investigate why so many users with the current program **Global Alumni** do not have an **Alum** role grant
2. Investigate why the directory filter for Program uses **Current program** instead of the role grants a user has. This leads to unintended and inconsistent results and access
3. Investigate why a profile which cannot be accessed still appears in search results
4. Simplify access to directories, directory search results, and profile pages. Consider how different roles, user types (staff, finalist, expert, member), and program status does or should impact access

iankigen:
@hfreund https://test2.masschallenge.org/admin/simpleuser/user/49509/change/ is not a staff user

jonkiparsky (approved):

hfreund:
> @hfreund https://test2.masschallenge.org/admin/simpleuser/user/49509/change/ is not a staff user

@iankigen Correct. They should be able to see user 18223 but cannot (I am able to see them as a staff user). My suspicion is that 18223 should have been tagged as an Alum but was not for some reason (possibly due to oversight, possibly due to a broken process). This is one of the tickets that I will create to investigate.

PR #397, opened 2020-04-15
[AC-7648] Remove obsoleted tests

jonkiparsky:
AC-7365 made some of the tests on impact-api obsolete, since the conditions they were testing were no longer possible due to the new db constraints. Removed those tests, which were now failing. 

There was also a null check in our code which has been removed since null is no longer allowed for that field

To test: 
verify that the changes are consistent with the changes made on AC-7365


iankigen (approved):
code looks good
- consistent with the changes made on AC-7365
- all test passing

geofrocker (approved):
QA looks good
- The tests are passing

PR #398, opened 2020-04-15
AC-7643 :: P1: Entrant label is displayed to non-staff users

freshprincekla:
**Changes introduced in [AC-7643 ](https://masschallenge.atlassian.net/browse/AC-7643)**
- Added flag for entrant status

**How to test:** 
- Login in as a non staff [user](http://localhost:8181/admin/simpleuser/user/20697/masquerade)
- Navigate to this user's [profile](http://localhost:1234/directory/people/22973)
- See that the user has no Entrant label

- Login in as a staff user (Demoadmin)
- Navigate to this user's [profile](http://localhost:1234/directory/people/22973)
- See that the user now has Entrant label

### DIFF
--- a/web/impact/impact/graphql/types/entrepreneur_profile_type.py
+++ b/web/impact/impact/graphql/types/entrepreneur_profile_type.py
@@ -8,7 +8,7 @@ from accelerator.models import (
     StartupRole,
     StartupTeamMember
 )
-
+from accelerator_abstract.models.base_user_utils import is_employee
 from impact.graphql.types.entrepreneur_startup_type import (
     EntrepreneurStartupType,
 )
@@ -50,7 +50,7 @@ class EntrepreneurProfileType(DjangoObjectType):
     def resolve_program_roles(self, info, **kwargs):
         user_roles_of_interest = [UserRole.FINALIST, UserRole.ALUM]
         startup_roles_of_interest = StartupRole.WINNER_STARTUP_ROLES
-        if info.context.user.is_staff:
+        if is_employee(info.context.user):
             startup_roles_of_interest += [StartupRole.ENTRANT]
         return get_user_program_and_startup_roles(
             self.user, user_roles_of_interest, startup_roles_of_interest)
--- a/web/impact/impact/graphql/types/entrepreneur_profile_type.py
+++ b/web/impact/impact/graphql/types/entrepreneur_profile_type.py
@@ -49,7 +49,8 @@ class EntrepreneurProfileType(DjangoObjectType):
 
     def resolve_program_roles(self, info, **kwargs):
         user_roles_of_interest = [UserRole.FINALIST, UserRole.ALUM]
-        startup_roles_of_interest = [StartupRole.ENTRANT]
-        startup_roles_of_interest += StartupRole.WINNER_STARTUP_ROLES
+        startup_roles_of_interest = StartupRole.WINNER_STARTUP_ROLES
+        if info.context.user.is_staff:
+            startup_roles_of_interest += [StartupRole.ENTRANT]
         return get_user_program_and_startup_roles(
             self.user, user_roles_of_interest, startup_roles_of_interest)
--- a/web/impact/impact/graphql/types/entrepreneur_profile_type.py
+++ b/web/impact/impact/graphql/types/entrepreneur_profile_type.py
@@ -1,26 +1,21 @@
 import graphene
-from graphene.types.generic import GenericScalar
-from graphene_django import DjangoObjectType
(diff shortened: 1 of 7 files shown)


web/impact/impact/graphql/types/expert_profile_type.py
```diff
@@ -132,8 +132,9 @@ def resolve_program_roles(self, info, **kwargs):
         without changing GraphQL queries on the front end.
         """
         user_roles_of_interest = [UserRole.FINALIST, UserRole.ALUM]
-        startup_roles_of_interest = [StartupRole.ENTRANT]
-        startup_roles_of_interest += StartupRole.WINNER_STARTUP_ROLES
+        startup_roles_of_interest = StartupRole.WINNER_STARTUP_ROLES
+        if info.context.user.is_staff:
```
iankigen: why not also check for `CLEARANCE_LEVEL_STAFF`

web/impact/impact/graphql/types/entrepreneur_profile_type.py
```diff
@@ -49,7 +49,8 @@ def resolve_title(self, info, **kwargs):
 
     def resolve_program_roles(self, info, **kwargs):
         user_roles_of_interest = [UserRole.FINALIST, UserRole.ALUM]
-        startup_roles_of_interest = [StartupRole.ENTRANT]
-        startup_roles_of_interest += StartupRole.WINNER_STARTUP_ROLES
+        startup_roles_of_interest = StartupRole.WINNER_STARTUP_ROLES
+        if info.context.user.is_staff:
```
jonkiparsky: We should not be checking the built-in "is_staff", since that only checks whether a user has access to the admin. Instead, use the `is_employee` function which can be imported from `accelerator_abstract.models.base_user_utils`

web/impact/impact/graphql/types/expert_profile_type.py
jonkiparsky: `is_employee` will do this check.

web/impact/impact/graphql/types/entrepreneur_profile_type.py
```diff
@@ -49,7 +49,8 @@ def resolve_title(self, info, **kwargs):
 
     def resolve_program_roles(self, info, **kwargs):
```
jonkiparsky: I notice that these functions are duplicates of each other. Can we possible refactor out the duplication?

web/impact/impact/tests/test_graphql.py
```diff
@@ -510,35 +510,40 @@ def test_query_program_roles_for_expert_with_same_program(self):
         self._assert_response_equals_json(query, expected_json)
 
     def test_query_prg_roles_for_selected_roles(self):
```
jonkiparsky: This test is very obscure - the name does not tell me anything about what it's meant to be testing, which _seems_ to be that a staff user gets a certain bundle of data back from this function `get_user_program_and_startup_roles`, and also gets the same bundle of data back from graphql, which by a coincidence, happens to call that function. 
I'm not sure that's a very interesting test, tbh, but clearly we need a test for the functionality change: non-staff users should not get back the entrant labels. 
We should also at least improve the naming of the test. 

If we can do this without adding (even more) duplication in this set of tests, that would be lovely.

jonkiparsky (changes requested):

web/impact/impact/graphql/types/expert_profile_type.py
```diff
@@ -132,8 +133,9 @@ def resolve_program_roles(self, info, **kwargs):
         without changing GraphQL queries on the front end.
         """
         user_roles_of_interest = [UserRole.FINALIST, UserRole.ALUM]
```
jonkiparsky: This function is an exact duplicate of the identically-named function on entrepreneur_program_type.py

web/impact/impact/tests/test_graphql.py
```diff
@@ -582,8 +562,13 @@ def test_query_for_user_without_confirmed_mentor_program_families(self):
             self.assertEqual(expert_profile["confirmedMentorProgramFamilies"],
                              [])
 
-    def _assert_response_equals_json(self, query, expected_json):
-        with self.login(email=self.basic_user().email):
+    def _assert_response_equals_json(self, query, expected_json,
```
jonkiparsky: if we have to break lines for any arguments, let's put each on its own line for readability

iankigen (approved):
QA looks good.
Confirmed:
- Entrant labels are not visible to any non-staff user

jonkiparsky (approved):

PR #400, opened 2020-04-22
[AC-7614] An endpoint to return all program role grants for a user

kaymation:
[JIRA](https://masschallenge.atlassian.net/browse/AC-7614)

# Description
To power some of our new pages, we need an endpoint that returns information on all PRGs for a given user. 
This delivers that. 

To Test - 
- running this endpoint locally, pick a user that has a nonzero # of program role grants, preferably a few
- visit the new http://localhost:8000/api/v1/program_role_grant/{user_id}/
- you should see results as per the fields requested in the ticket

NB: User Role is not a required field for a Role Grant, so you might see many 
NBB: if you enter an invalid user_id, empty results will be show. i figured that this was preferred to adding an extra DB hit to validate a users existence on all requests, since this endpoint will be used frequently.


geofrocker (approved):
QA looks good
- I have tried a couple of users and I am satisfied with the response from the API as it conforms with ticket's tickets done.

kaymation:
this works was deemed redundant as graphql can deliver this

PR #402, opened 2020-04-22
[AC-7596]: Create API call for expert confirmation

iankigen:
## [AC-7596](https://masschallenge.atlassian.net/browse/AC-7596)

**Changes introduced**
- add API for expert confirmation

**Testing**
On this branch
- As an expert user, visit [program confirmation page](http://localhost:8000/api/v1/expert/program-confirmation/).

Confirm that:
- A mentor can defer their participation in a program
- A mentor can confirm their participation in a program
- An email with a summary of programs is sent after program confirmation
    run `impact-api` on debug mode to see email on the terminal instance: `make run-server-1`
- Posting invalid data returns an error
- Non-expert users cannot POST 
   

input data format:
```
{
    "confirmed": ["list", "of", "program", "ids"],
    "deferred": ["list", "of", "program", "ids"],
}
```


shakirandagire:
Doing QA for this task

shakirandagire (approved):
QA looks good to me:
I was able to visit the expert confirmation API link
- I created deferred mentor program participation as well as  confirmed mentor program participation
- I got an error message stipulating the deferred or confirmed inputs should be lists.
- I received an email with a summary of programs after program confirmation

web/impact/impact/v1/views/user_program_confirmation_detail_view.py
```diff
@@ -0,0 +1,101 @@
...
+        programrolegrants = [
+            ProgramRoleGrant(
+                person=self.user,
+                program_role=ProgramRole.objects.filter(
+                    program=program,
+                    user_role=user_role_obj
+                ).first())
+            for program in programs
+        ]
+        ProgramRoleGrant.objects.bulk_create(
```
jonkiparsky: I appreciate that you thought to use bulk_create to spare the DB, but you're still hitting the DB in the loop. We could change that by doing something like:
```
program_roles = dict(ProgramRole.objects.filter(user_role=user_role).order_by('-id').values_list('program_id', 'id')
```
before the loop starts and then specify 
`program_role_id=program_roles[program_id]` 
as we create each ProgramRoleGrant in the loop. 

This would also allow you to loop over `program_ids`, rather than getting the programs from the db. 

(the `order_by` is a bit of a hack - in case staff create later ProgramRoles with the UserRoles we're interested in , this should ensure that the dict has the first created ProgramRole with this UserRole, which will be the one we want)

Also, as long as I'm nerding out on db access, I notice that you use a throwaway `get` to retrieve the user_role_obj by name rather than using `user_role__name=user_role` in the ProgramRoleGrant creation. Is this to avoid the join?

web/impact/impact/v1/views/user_program_confirmation_detail_view.py
```diff
@@ -0,0 +1,101 @@
...
+        if confirmed_programs:
+            self.update_user_confirmation(
+                UserRole.DEFERRED_MENTOR, UserRole.MENTOR, confirmed_programs)
+            self.send_email()
+        return Response({'success': True})
+
+    def send_email(self):
+        context = {
+            'mentor_first_name': self.user.first_name,
+            'confirmed_programs': Program.objects.filter(
```
jonkiparsky: Interesting. Do we want to list 

- all of the programs that the user has confirmed for in this particular action
- all of the upcoming and active programs that the user has confirmed for, or
- all of the programs that the user has confirmed for, ever?

This does the latter, which might not be what we want. We should check with Hugh about which we want.

web/impact/impact/v1/views/user_program_confirmation_detail_view.py
```diff
@@ -0,0 +1,101 @@
...
+                pk__in=self.user.programrolegrant_set.filter(
+                    program_role__user_role__name=UserRole.MENTOR
+                ).values_list('program_role__program__pk', flat=True)
```
jonkiparsky: This query is a little hard to make sense of. The first thought would be to pull out the embedded query, which would help, but maybe we can do better. 

One thing that comes to mind is something like
```
program_data= self.user.programrolegrant_set.filter(
    program_role__user_role__name=UserRole.MENTOR).values(
        start_date=F('program_role__program__start_date'), 
        end_date=F('program_role__program__end_date'), .... etc etc)
```

Since template syntax uses the same dot lookup for dict values as for object lookup, you should be able to pass this off to the template without changing anything in the template itself.

jonkiparsky (commented):
A few suggestions.

web/impact/impact/tests/test_user_confirm_participation_view.py
```diff
@@ -0,0 +1,155 @@
...
+        user = self.basic_user()
+        with self.settings(
+                EMAIL_BACKEND=IMPACT_BACKEND_PATH):
+            with self.login(email=user.email):
+                url = reverse(UserProgramConfirmationDetailView.view_name,
+                              args=[user.pk])
+                self.client.post(url, {
+                    'confirmed': [self.program.pk],
+                })
+            self.assertTrue(mocked_backend.called)
```
jonkiparsky: I would prefer to write tests using the `mail.outbox`, see test_minimal_email_handler.py for examples. 
One of the benefits of getting in this habit is that we don't need to mock, another benefit is that we can write tests that actually verify the content of the email, not just that an email was sent. 

That is, if we want to test that an email goes to two distinct users, we can look for both email addresses in `mail.outbox`, if we want to verify that a given program is mentioned in an email, we can look for that in the content of the outgoing email, etc.

web/impact/impact/v1/views/user_program_confirmation_detail_view.py
```diff
@@ -0,0 +1,101 @@
...
+        deferred_programs = validate_input(request, 'deferred')
+        confirmed_programs = validate_input(request, 'confirmed')
+
+        if deferred_programs:
+            self.update_user_confirmation(
+                UserRole.MENTOR, UserRole.DEFERRED_MENTOR, deferred_programs)
+        if confirmed_programs:
+            self.update_user_confirmation(
+                UserRole.DEFERRED_MENTOR, UserRole.MENTOR, confirmed_programs)
+            self.send_email()
```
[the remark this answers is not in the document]
iankigen: Confirmed by Hugh:
- we are only sending emails for confirmed participation.

web/impact/impact/v1/views/user_program_confirmation_detail_view.py
iankigen: didn't think of this. great suggestion

web/impact/impact/v1/views/user_program_confirmation_detail_view.py
iankigen: Confirmed by Hugh and implemented:
- Listing only the confirmed program participation selections the user recently made. 
- Include a link to view all current participation as well.

iankigen:
@jonkiparsky I have implemented PR feedback  + clarified with Hugh on the confirmation email

web/impact/impact/v1/views/user_program_confirmation_detail_view.py
```diff
@@ -0,0 +1,107 @@
...
+        self.user.programrolegrant_set.filter(
+            program_role__program__in=program_ids,
+            program_role__user_role__name=user_role
+        ).delete()
+
+    def update_user_confirmation(self, delete_roles, add_roles, program_ids):
+        self.delete_program_role_grants(program_ids, delete_roles)
+        self.add_program_role_grants(program_ids, add_roles)
+
+    def post(self, request, pk):
```
jonkiparsky: POST requests don't usually take args, do they? Shouldn't the user id be passed in as part of POST.data?

iankigen: POST requests can take args from the URL  and its always a good idea to pass data as part of post.data.
I used it since the required user `pk` existed as an arg and getting the same `pk`  from POST.data seemed redundant. Do you suggest I switch to getting it from POST.data?
the `pk` is there to facilitate GET on `/api/v1/user/<pk>/program_confirmation/`.

web/impact/impact/tests/test_user_confirm_participation_view.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def _get_user_programs_for_role(user, role):
+    return list(
+        user.programrolegrant_set.filter(
+            program_role__user_role__name=role,
+        ).values_list('program_role__program', flat=True).distinct())
+
+
+def _get_programrolegrant_for_role(role, user, program=None):
```
jonkiparsky: since this creates rather than getting, I would like to call it `_create_program_role_grant_for_role` (use underscores to separate words!)

web/impact/impact/v1/views/user_program_confirmation_detail_view.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def extract_values(request, key):
+    try:
+        return [int(i) for i in request.data.get(key, [])]
+    except ValueError:
+        raise ParseError(INVALID_INPUT_ERROR.format(key))
+
+
+class UserProgramConfirmationDetailView(BaseDetailView):
```
jonkiparsky: Since this does not return details about an object in the database, I don't think the "Detail" part of this name is correct. What this does is it allows users to confirm and see mentor participation in a program, so maybe it should be called `MentorParticipationView`?

web/impact/impact/v1/views/user_program_confirmation_detail_view.py
jonkiparsky: Clarified in slack convo.

web/impact/impact/v1/views/user_program_participation_view.py
```diff
@@ -0,0 +1,115 @@
...
+            ignore_conflicts=True
+        )
+
+    def delete_program_role_grants(self, program_ids=None, user_role=None):
+        self.user.programrolegrant_set.filter(
+            program_role__program__in=program_ids,
+            program_role__user_role__name=user_role
+        ).delete()
+
+    def update_user_confirmation(self, delete_roles, add_roles, program_ids):
```
geofrocker: NON-BLOCKING: `handle_user_confirmation` looks more intuitive here. `update_user_confirmation` sounds more like we are updating a previous confirmation

geofrocker (commented):
Thanks, Ian. I am sorry for coming to this late. I only realised when I tried to use this branch for my current blocked ticket. One blocking issue and two non-blocking issues.

web/impact/impact/v1/views/user_program_participation_view.py
```diff
@@ -0,0 +1,115 @@
...
+        self.delete_program_role_grants(program_ids, delete_roles)
+        self.add_program_role_grants(program_ids, add_roles)
```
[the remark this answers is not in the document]
hfreund: @iankigen the ticket requires a check for an existing deferred or confirmed program role grant before adding another PRG:

> **A mentor confirms their participation in a program:**
> - Check for a Deferred Mentor program role grant for the program in question; if one exists, remove it
> - Check for a Confirmed Mentor program role grant for the program in question; if one does not already exist, add it

> **A mentor defers their participation in a program:**
> - Check for a Confirmed Mentor program role grant for the program in question; if one exists, remove it
> - Check for a Deferred Mentor program role grant for the program in question; if one does not already exist, add it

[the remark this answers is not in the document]
geofrocker: @hfreund That's is right and this PR is doing exactly that. This is just concern on the workflow. Perhaps this comment should have been added to the ticket. Do we expect a user to be both a desired mentor and a confirmed mentor in the same program?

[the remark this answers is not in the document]
hfreund: Yes, that is the current behavior and we do not expect that to change. This way staff can use the workflow of desired >> confirmed to account for the number of experts who commit to a program after doing outreach.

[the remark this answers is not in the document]
geofrocker: Thanks for that clarification

iankigen:
From sync with Geoffrey:
- GET method is not required: I have removed the GET method
- Using `user/<user_id>/program-confirmation` endpoint will require an additional API call on the front end to get user id: I have switched to using the session to get user-id

- I have also added IsExpertUser permission.

web/impact/impact/v1/views/user_program_participation_view.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def extract_values(request, key):
+    try:
+        return [int(i) for i in request.data.get(key, [])]
+    except ValueError:
+        raise ParseError(INVALID_INPUT_ERROR.format(key))
+
+
+class ExpertParticipationView(ImpactView):
```
jonkiparsky: I notice that the view name and the class name are out of sync. I personally think `MentorParticipationView` is more apt than `ExpertParticipationView`, but I would be willing to accept the latter - in any case, the view_name should conform to the class name, and the file name should also agree with both.

hfreund: FWIW this view will be used for the foreseeable future for mentor confirmation only. There is a desire, but no current plan, to have a more similar (possibly the same) workflow for all types of expert confirmation (judge and mentor), so you may consider that when deciding on a naming convention.

iankigen: switched to `MentorParticipationView`

PR #404, opened 2020-04-27
AC-7697 - installing waitress and increasing timeouts for nginx

shanbady:
This PR swaps out gunicorn for waitress as the wsgi server used to serve impact. This is to resolve a bad issue that happens when salesforce does its daily pull of user updates. with gunicorn, a long running request takes causes the entire server to go unresponsive and return 503s.

to test:
- checkout this branch and re-run "make build" in impact.
- re-start servers via "make run-all-servers" and confirm that the impact api is accessible and works as expected. There are no out-facing functional changes


pbx:
Because [AC-7511 hasn't merged yet](https://github.com/masschallenge/impact-api/pull/403) and I'm one of the affected people,  a `make build` in impact-api doesn't work.  I merged the head of that branch (`a17daaf`) into AC-7697 on my machine.

pbx (changes requested):
I like this, and waitress looks like a great find.

Once I merged AC-7511 into this branch locally (which I needed to do for the build to succeed) I was able to `make run-server`; requests to several endpoints functioned normally, and I confirmed the `Server: waitress` header on the response.

Just a few gunicorn references that need to be cleaned up:
```
$ ack guni
web/impact/requirements/base.txt
15:gunicorn
23:gunicorn

web/impact/impact/urls.py
136:# use staticfiles with gunicorn (not recommneded!)

README.md
132:At times, you may want to see the gunicorn + supervisor
```

I do wonder if we should have some follow-up work to look at the requests/queries that were so intensive as to cause timeouts, but that question doesn't need to hold up shipping this work.

pbx:
(Apologies, I accidentally pushed my local merge earlier. I've reset it to the previous head, `d4859bd`.)

shanbady:
@pbx removed references to gunicorn

pbx (approved):

jonkiparsky (approved):
QA: 
- checked out the branch, ran `make build`, then `make run-all-servers`
-- accelerate failed to load
- nuked local environment (`make stop-all-servers; make delete-vms remove_images=`) and ran `make build-all; make run-all-servers`
-- impact and accelerate servers both loaded

PR #405, opened 2020-04-29
[AC-7561] Create "cancel_office_hour_session" API endpoint

iankigen:
## [AC-7561](https://masschallenge.atlassian.net/browse/AC-7561)

**Changes Introduced**
- add cancel office hour session API endpoint

**Testing**
- checkout to this branch and `make run-server-1`
- As a staff user or mentor visit [cancel office hour endpoint](http://localhost:8000/api/v1/cancel_office_hour_session/). 
Allowed methods = [POST]
Allowed users = [Staff users, Office hour session owner(mentor)]
- POST data format:
```
 {
    'id': <office hour session id>,
    'message': <optional message to attendees>
}
```
Confirm that:
- Only allowed users can post data
- On successful offices hour session cancellation, you get a response with detail as follows:
  - cancelled by staff: 'Canceled office hours on behalf of {mentor user name} at {start time} - {end time} on {date}'
  - cancelled by mentor: 'Canceled office hours at {start time} - {end time} on {date}'
- On successful cancellation, the creator of the session should be notified in the following ways that either they or MassChallenge staff canceled the session
- A mentor cannot cancel reserved office hour session, instead, they get with details as follows:
  - 'Office hour session could not be canceled. You do not have permission to cancel that session'
- A notification email sent to the office hours holder’s email address when the session is canceled on their behalf by a staff person
  - including the message to attendees if provided
- A notification email sent to the office hours holder and finalist when the session is canceled on their behalf by a staff person
  - including the message to attendees if provided


jonkiparsky (changes requested):
First pass at review

web/impact/impact/v1/views/cancel_office_hour_session_view.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def send_email(self, context, to_addr):
+        html_email = loader.render_to_string(
+            'emails/cancel_office_hour_session_email.html',
+            context
+        )
+        MinimalEmailHandler(
+            to=to_addr,
+            subject=SUBJECT.format(**context),
+            body=None,
```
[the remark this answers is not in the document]
iankigen: I have attached html alternative instead

geofrocker:
Doing QA

web/impact/impact/tests/test_cancel_office_hour_session_view.py
```diff
@@ -0,0 +1,185 @@
...
+
+class TestCancelOfficeHourSession(APITestCase):
+    url = reverse(CancelOfficeHourSessionView.view_name)
+
+    def test_mentor_can_cancel_their_own_unreserved_office_hour(self):
+        mentor = self._expert_user(UserRole.MENTOR)
+        office_hour = MentorProgramOfficeHourFactory(
+            mentor=mentor, finalist=None
+        )
+        with self.login(email=mentor.email):
```
jonkiparsky: the code from 33 to 36 is repeated many times. Should be encapsulated in a function like `self.post_cancellation(user, office_hour)`, which would return the response for the tests that need it.

web/impact/impact/v1/views/cancel_office_hour_session_view.py
```diff
@@ -0,0 +1,126 @@
...
+from impact.minimal_email_handler import MinimalEmailHandler
+from impact.permissions.v1_api_permissions import OfficeHourPermission
+from impact.v1.views.impact_view import ImpactView
+
+User = get_user_model()
+
+DEFAULT_TIMEZONE = 'UTC'
+SUBJECT = '[Office Hours] Canceled: {date}, {start_time}'
+
+STAFF_NOTIFICATION = 'Canceled office hours on behalf of ' \
```
jonkiparsky: Not super important, but house style typically uses parens and not trailing backslashes to concatenate strings across line breaks.  
```
STAFF_NOTIFICATION = ('Canceled office hours on behalf of ' 
                     '{mentor_name} at {start_time} - {end_time} on {date}')
```

For consistency, let's stick with that.

web/impact/impact/v1/views/cancel_office_hour_session_view.py
```diff
@@ -0,0 +1,126 @@
...
+        'location': office_hour.location.name if office_hour.location else '',
+        'dashboard_url': get_office_hours_list_url(family_slug, program_slug),
+        'mentor_name': office_hour.mentor.get_profile().full_name(),
+        'phone': office_hour.program.program_family.phone_number,
+        'message': message,
+    }
+
+
+def get_ui_notification(context, staff=False):
+    if staff:
```
jonkiparsky, line 57: Not a blocker, but dicts are preferable over ifs for data lookup, since they are extensible.
For example, if we decided to notify finalists of cancelled sessions, this would need to be changed, but

```
def get_ui_notification(context, staff=False):
    notification = UI_NOTIFICATIONS[participant_role]
    return notification.format(**context)
```

would not.

web/impact/impact/v1/views/cancel_office_hour_session_view.py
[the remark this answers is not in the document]
jonkiparsky: Hm. I don't love this. Plain text is the standard for email, html is an alternative. 
But let it go, it won't break much for many users.

web/impact/impact/v1/views/cancel_office_hour_session_view.py
```diff
@@ -0,0 +1,126 @@
...
+            subject=SUBJECT.format(**context),
+            body=None,
+            from_email=settings.NO_REPLY_EMAIL,
+            attach_alternative=[html_email, 'text/html'],
+        ).send()
+
+    def post(self, request):
+        message = request.data.get('message', None)
+        id = request.data.get('id', None)
+        office_hour = get_object_or_404(MentorProgramOfficeHour, id=id)
```
jonkiparsky: I think this will throw an error if a user commits a double-tap when they post their request. (since there will be two requests, and the second one will find the office hour has been deleted and throw a 404)

This is probably not ideal UX. 

I would say, if no office hour with that ID exists, we should either consider that a success (they wanted it deleted, it's deleted...) or report back something informative like "You wanted to delete that office hour, but unfortunately, it doesn't exist, so we can't... um... delete it. Right. Nevermind, we're good, bye!"

web/impact/impact/v1/views/cancel_office_hour_session_view.py
```diff
@@ -0,0 +1,126 @@
...
+            office_hour.delete()
+            return ui_notification
+        else:
+            self.handle_notification(
+                office_hour, shared_context, 'MassChallenge has')
+            ui_notification = get_ui_notification(shared_context, staff=True)
+            office_hour.delete()
+            return ui_notification
+
+    def send_email(self, context, to_addr):
```
jonkiparsky: would be nice if this were adjacent to the other notification code...

web/impact/impact/v1/views/cancel_office_hour_session_view.py
```diff
@@ -0,0 +1,126 @@
...
+    def get_addressee_info(self, office_hour, mentor_name):
+        addressee_dict = {
+            'mentor': {'hours_with': '',
+                       'to_addrs': [office_hour.mentor.email]}
+        }
+        if office_hour.finalist:
+            addressee_dict['mentor']['hours_with'] = ' with {}'.format(
+                office_hour.finalist.get_profile().full_name())
+            addressee_dict['finalist'] = {
+                'hours_with': f' with {mentor_name}',
```
jonkiparsky: you use `.format` in most places let's be consistent and use it here as well.

web/impact/impact/v1/views/cancel_office_hour_session_view.py
```diff
@@ -0,0 +1,126 @@
...
+
+class CancelOfficeHourSessionView(ImpactView):
+    permission_classes = (OfficeHourPermission,)
+    view_name = 'cancel_office_hour_session_view'
+
+    def cancel_office_hour_session(self, office_hour, user, message):
+        shared_context = get_office_hour_shared_context(office_hour, message)
+        if user == office_hour.mentor:
+            if office_hour.finalist:
+                raise PermissionDenied(PERMISSION_DENIED)
```
jonkiparsky, line 70: This message doesn't tell the user why they can't cancel (because the hour is reserved) or what they can do about it. Is that okay?

jonkiparsky (changes requested):

geofrocker (approved):
QA Looks good. Thanks, Ian
- An endpoint exists on the API which takes an office hour id and an option message-to-attendees ✅
- Staff user can cancel office hours on behalf of a mentor ✅
- When a reserved office hour session is cancelled, two emails are sent out the hour holder and the finalist with appropriate messages ✅
- When an open office hour session is cancelled by a staff user, an email is sent out the hour holder with an appropriate message ✅
- When an office hour holder cancels their own session, they receive an email with an appropriate message ✅
- Finalists users cannot cancel office hour session ✅

web/impact/impact/v1/views/cancel_office_hour_session_view.py
iankigen: PERMISSION_DENIED text is ` "Office hour session could not be canceled. You do not have permission to cancel that session"` as per the ticket description

jonkiparsky: would be nice to check with product when we have questions about the ticket specification, but okay.

jonkiparsky (approved):

web/impact/impact/v1/views/cancel_office_hour_session_view.py
hfreund: I agree we should let the user delete an office hour with a nonexistent ID if it is theirs to delete. There is no need to display a UI message in this case. However, if we are concerned there may be instances of non-existing office hour ID's, would it make sense to have some logging for our records at the very least?

jonkiparsky: @hfreund the issue is, if a user submits a request with an id that does not correspond to any existing office hour, then the most likely scenario is that they've tried to delete a session that has already been deleted. In this case, we'd rather not error out on them, because probably they'll have just double-tapped the submit button. 

Anyway, this ship has sailed. If you have questions about the existing behavior, ping me and we can get you answers.

hfreund: Nope. I think we're in agreement here.

PR #406, opened 2020-04-29
[AC-7562] Create API endpoint to cancel office hours reservations

jonkiparsky:
https://masschallenge.atlassian.net/browse/AC-7562

adds cancel office hour reservation API endpoint

Testing notes: 
- endpoint at http://localhost:8000/api/v1/cancel_office_hour_reservation/
- POST should include 
"office_hour_id" = ID of the MentorProgramOfficeHour whose reservation should be cancelled
 "requesting_user": ID of the user in whose name this action is to be taken
"message": optional message to include in notifications


iankigen:
Doing QA/Code review

web/impact/impact/tests/test_cancel_office_hour_reservation.py
```diff
@@ -0,0 +1,116 @@
...
+        '''A non-staff user attempting to cancel someone else's reservation
+        should not receive a notification'''
+        office_hour = MentorProgramOfficeHourFactory()
+        self._submit_cancellation(office_hour, user=self.basic_user())
+        self.assert_not_notified(office_hour.mentor)
+
+    def test_user_attempts_to_cancel_non_existent_reservation(self):
+        office_hour = MentorProgramOfficeHourFactory(finalist=None)
+        self._submit_cancellation(office_hour, user=self.basic_user())
+        # assert what?
```
iankigen, line 78: is this intentional?

web/impact/impact/v1/views/cancel_office_hour_reservation_view.py
```diff
@@ -0,0 +1,91 @@
...
+SUBJECT_LINE = "MassChallenge | Cancelled Office Hours with {} {}"
+
+
+class CancelOfficeHourReservationView(ImpactView):
+    view_name = "cancel_office_hours_reservation"
+
+    def post(self, request):
+        office_hour_id = request.POST.get("office_hour_id")
+        message = request.POST.get("message", "")
+        requesting_user_id = request.POST.get("user_id")
```
geofrocker: We don't expect `requesting_user_id` to be different from `request.user.id`. This would be redundant.

web/impact/impact/v1/views/cancel_office_hour_reservation_view.py
```diff
@@ -0,0 +1,91 @@
...
+        requesting_user = User.objects.get(id=requesting_user_id)
```
geofrocker: Let's handle a few edge cases here and return appropriate messages
- Return a meaningful message in case the user(this can be ignored if we use `request.user.id`) or office_hour does not exist.

geofrocker (changes requested):
A few issues for cleanup.

web/impact/impact/v1/views/cancel_office_hour_reservation_view.py
iankigen: we should be using `request.data` since `request.POST`  only handles form data.

iankigen (changes requested):

web/impact/impact/v1/views/cancel_office_hour_reservation_view.py
```diff
@@ -0,0 +1,91 @@
...
+        office_hour = MentorProgramOfficeHour.objects.get(pk=office_hour_id)
```
iankigen: `get_object_or_404` is better to avoid 500 when objects don't exist

web/impact/impact/v1/views/cancel_office_hour_reservation_view.py
iankigen: This is necessary since the staff can also cancel reservations

iankigen:
**From QA**
- An endpoint exists on the API which takes an office hour id and a user id and an optional notification message and cancels the user's reservation for that office hour session if it exists :white_check_mark:
- The user is allowed to cancel that reservation (ie, is the user in question or is MC staff) :white_check_mark:
  - Both staff and finalist can cancel a reservation
- On successful cancellation, the mentor and mentee of the session should be notified that either they or MassChallenge staff cancelled the reservation :x:
  - Email is sent as expected, however `canceled by` is always referencing the **finalist** even when MassChallenge staff cancels the reservation

- Notification sent to users will be a stock template that will include a user message if submitted.:white_check_mark:

- Notifications should be sent using an external mail-sending utility rather than sending email in-line as we currently do :white_check_mark:

- If the cancellation fails, return a reason ("You do not have permission to cancel that reservation" or "That reservation does not exist") :x:
  - No reason is provided. returns `{'success': false}`

web/impact/impact/v1/views/cancel_office_hour_reservation_view.py
geofrocker: Either way. When a staff user makes the api call, the still the `_can_cancel` check.

web/impact/impact/v1/views/cancel_office_hour_reservation_view.py
```diff
@@ -0,0 +1,91 @@
...
+                                                   message,
+                                                   office_hour.mentor,
+                                                   mentor_template_name))
+            _send_email(prepare_email_notification(office_hour,
+                                                   requesting_user,
+                                                   office_hour.mentor,
+                                                   message,
+                                                   office_hour.finalist,
+                                                   finalist_template_name))
+            _cancel_reservation(office_hour)
```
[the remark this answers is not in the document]
jonkiparsky: This implies that cancellation of a reservation can fail, which would be a possibility we'd have to handle. Can you map out what it is that we should be guarding against here?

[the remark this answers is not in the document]
geofrocker: This is non-blocking. We would be guarding against false alerts. My idea here is just to be sure that we always send notifications after successful completion of processes.

web/impact/templates/emails/cancel_office_hour_reservation_email_to_finalist.html
```diff
@@ -0,0 +1,17 @@
+Hi {{recipient.first_name }} {{recipient.last_name}},
+
+Your upcoming office hour session at {{office_hour.start_date_time |date:"h:m"}} on
```
iankigen: `office_hour`, `mentor_name` and `startup_name` are no longer in the context

iankigen (changes requested):
A few changes request otherwise everything LGTM

web/impact/impact/v1/views/cancel_office_hour_reservation_view.py
```diff
@@ -0,0 +1,94 @@
...
+    def _extract_posted_data(self, data, fields):
+        for field in fields:
+            self._extract_field(data, field)
+
+    def post(self, request):
+        self._extract_posted_data(request.data, self.posted_fields)
+        self.user = request.user
+        self.office_hour = MentorProgramOfficeHour.objects.get(
+            pk=self.office_hour_id)
+        success = False
```
iankigen: If the cancellation fails, return a reason ("You do not have permission to cancel that reservation" or "That reservation does not exist")

iankigen (changes requested):

web/impact/templates/emails/cancel_office_hour_reservation_email_to_mentor.html
```diff
@@ -0,0 +1,18 @@
+Hi {{mentor.first_name}},
+
+Your upcoming office hour session at {{office_hour_date_time|date:"h:m"}} on
+{{office_hour_date_time| date}} with {{startup_name}} was
+cancelled by {{cancelling_party.first_name}} {{cancelling_party.last_name}}. This
```
iankigen: **Non-blocker:** Would it be better to mention `MassChallenge staff canceled the reservation` instead of staff username. In the case, that a staff user cancels the reservation?

iankigen (approved):
**From QA**
- An endpoint exists on the API which takes an office hour id and a user id and an optional notification message and cancels the user's reservation for that office hour session if it exists :white_check_mark:
- The user is allowed to cancel that reservation (ie, is the user in question or is MC staff) :white_check_mark:
  - Both staff and finalist can cancel a reservation
  - Non- (staff/finalist) cannot cancel a reservation
- On successful cancellation, the mentor and mentee of the session should be notified that either they or MassChallenge staff canceled the reservation:
    - On successful cancellation, the mentor and mentee of the session should be notified that either they or MassChallenge staff canceled the reservation ✅ 

- Notification sent to users will be a stock template that will include a user message if submitted.:white_check_mark:

- Notifications should be sent using an external mail-sending utility rather than sending email in-line as we currently do :white_check_mark:

- If the cancellation fails, return a reason ("You do not have permission to cancel that reservation" or "That reservation does not exist") ✅ 
  - A reason is provided when cancellation fails or succeeds.

geofrocker (approved):
Code looks good. Thanks, Jon

PR #408, opened 2020-04-30
AC-7719 Resolve merge Conflicts on impact-api

geofrocker:


pbx (approved):

pbx:
The conflict looks to be a side effect of some recently merged work. (Side note: the reason the version number there is out of date is, **our release tooling does not merge the updated version number back to development**, so it almost never gets touched there.)

PR #412, opened 2020-05-12
[AC-7752]: Create Utility for obtaining the user id to be used for new React apps

PrinceDavis:
https://masschallenge.atlassian.net/browse/AC-7752

Create Utility for obtaining the user id to be used for new React apps

Testing notes: 
- make a graphQL request for either an expert or an entreprenuer without providing an id and see that server does not break
- Fire up directory and see that graphQL server still response to queries with ID


shakirandagire (approved):
The code changes look good to me

web/impact/impact/graphql/query.py
```diff
@@ -34,7 +35,7 @@ def resolve_expert_profile(self, info, **kwargs):
         return expert
 
     def resolve_entrepreneur_profile(self, info, **kwargs):
-        user_id = kwargs.get('id')
+        user_id = kwargs.get('id') or info.context.user.id
```
jonkiparsky: use the default option for kwargs.get:

`user_id = kwargs.get('id', info.context.user.id)`

web/impact/impact/graphql/query.py
```diff
@@ -25,7 +25,8 @@ class Query(graphene.ObjectType):
         EntrepreneurProfileType, id=graphene.Int())
 
     def resolve_expert_profile(self, info, **kwargs):
-        user_id = kwargs.get('id')
+        user_id = kwargs.get('id') or info.context.user.id
```
jonkiparsky: use the default option for kwargs.get:

`user_id = kwargs.get('id', info.context.user.id)`

web/impact/impact/tests/test_graphql.py
```diff
@@ -658,6 +658,60 @@ def test_current_judge_cannot_view_mentor_profile(self):
...
+        user_role = get_user_role_by_name(UserRole.FINALIST)
+        program_role = ProgramRoleFactory.create(
+            user_role=user_role,
+            program__program_status=ACTIVE_PROGRAM_STATUS
+        )
+        ProgramRoleGrantFactory(person=user,
+                                program_role=program_role)
+        user.set_password('password')
+        user.save()
+        query = """
```
jonkiparsky: would be nice to DRY this out, see comment above.

web/impact/impact/tests/test_graphql.py
```diff
@@ -658,6 +658,60 @@ def test_current_judge_cannot_view_mentor_profile(self):
         self._assert_expert_cannot_view_profile(UserRole.JUDGE,
                                                 UserRole.MENTOR)
 
+    def test_loggedin_expert_data_is_returned_on_missing_id(self):
+        user = expert_user(UserRole.MENTOR)
+        query = """
```
jonkiparsky: Can we DRY out this repeated code?
The query and expected_json defined here are pretty stereotyped. `query` could probably be a constant and `expected_json` could be a function taking the user and possibly the key string, in this case "expertProfile" and returning a dict of this shape.

web/impact/impact/tests/test_graphql.py
```diff
@@ -658,6 +658,60 @@ def test_current_judge_cannot_view_mentor_profile(self):
...
+                'expertProfile': {
+                    'user': {
+                        'lastName': user.last_name
+                    }
+                }
+            }
+        }
+        self._assert_response_equals_json(query, expected_json, email=user.email)
+
+    def test_loggedin_entrepnur_data_is_returned_on_missing_id(self):
```
jonkiparsky: s/b test_logged_in_entrepreneur_data_is_returned_on_missing_id

jonkiparsky (changes requested):
Change looks good. Would be nice to clean up the tests a bit.

web/impact/impact/tests/test_graphql.py
PrinceDavis: Say more here @jonkiparsky what is s/b?

jonkiparsky: sorry, "s/b" stands for "should be"

jonkiparsky (approved):

PR #414, opened 2020-05-15
[AC-7557] Cancellation workflow for office hours

iankigen:
## [AC-7557](https://masschallenge.atlassian.net/browse/AC-7557)

**Changes Introduced:**
- standardize office hour cancellation response

**Testing**
- checkout to `AC-7557` on `accelerate`, `semantic-ui-theme` and `impact-api`
- run servers ( accelerate and impact-api)
- On semantic-ui-theme:
    - run `gulp build`
    - run `yarn link`
- On accelerate
    - run `yarn link "semantic-ui-theme"`
    - run `yarn start`

- Visit [http://localhost:8181/newofficehour](http://localhost:8181/newofficehour)
For test purposes you can set `office_hour_id` and `mentor_id (office hour holder id)` using URL query. i.e `http://localhost:8181/newofficehour?office_hour_id=1&mentor_id=2`
- Verify that the cancellation workflow meets the definition of done according to [AC-7557](https://masschallenge.atlassian.net/browse/AC-7557)
- Allowed users: 
    - Staff users - Can cancel session/reservation on behalf of mentor/finalist
    - Office hour session owner(mentor) - can cancel their unreserved office hour sessions
    - Finalist- can cancel their reserved office hour sessions

**Sibling PRs**
- [semantic UI theme PR](https://github.com/masschallenge/semantic-ui-theme/pull/44)
- [Accelerate PR](https://github.com/masschallenge/accelerate/pull/2614)

### DIFF
--- a/web/impact/impact/tests/api_test_case.py
+++ b/web/impact/impact/tests/api_test_case.py
@@ -99,4 +99,4 @@ class APITestCase(TestCase):
                 self.assertEqual(options[key], params)
 
     def assert_ui_notification(self, response, notification):
-        self.assertEqual(response.data['detail'], notification)
+        self.assertEqual(response.data, notification)
--- a/web/impact/impact/tests/api_test_case.py
+++ b/web/impact/impact/tests/api_test_case.py
@@ -98,5 +98,10 @@ class APITestCase(TestCase):
                 self.assertIn(key, options)
                 self.assertEqual(options[key], params)
 
-    def assert_ui_notification(self, response, notification):
-        self.assertEqual(response.data, notification)
+    def assert_ui_notification(self, response, success, notification):
+        expected_ui_notification = {
+            'success': success,
+            'header': self.success_header if success else self.fail_header,
+            'detail': notification
+        }
+        self.assertEqual(response.data, expected_ui_notification)
--- a/web/impact/impact/tests/api_test_case.py
+++ b/web/impact/impact/tests/api_test_case.py
@@ -99,9 +99,10 @@ class APITestCase(TestCase):
                 self.assertEqual(options[key], params)
 
     def assert_ui_notification(self, response, success, notification):
-        expected_ui_notification = {
-            'success': success,
-            'header': self.success_header if success else self.fail_header,
-            'detail': notification
-        }
-        self.assertEqual(response.data, expected_ui_notification)
+        data = response.data
+        header = self.success_header if success else self.fail_header
+        self.assertTrue(all([
+            data['success'] == success,
+            data['header'] == header,
+            data['detail'] == notification
+        ]), msg='Notification data was not as expected')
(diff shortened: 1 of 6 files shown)


jonkiparsky (changes requested):
These changes look reasonable, however there are some tests required

iankigen:
I have updated ui notification tests to factor in header, success, and detail in response data.

web/impact/impact/tests/test_cancel_office_hour_reservation.py
```diff
@@ -36,7 +42,9 @@ def test_finalist_cancels_and_gets_ui_notification(self):
         response = self._submit_cancellation(office_hour,
                                              user=office_hour.finalist)
         notification = formatted_success_notification(office_hour)
-        self.assert_ui_notification(response, notification)
+        expected_ui_notification = get_ui_notification_dict(
```
jonkiparsky: Can we maybe try to keep this clean? 

The consistent pattern here is, we have a success or failure, which conditions a success flag, a "header" and a "notification" - which are not actually header or notification, they are two fragments of a notification, which is being glommed together by the UI, but we'll use those words until we find the right ones. (which might not happen here, but should happen)

The "header" part and the "success" part are not independent, so we should only pass `success` and `notification`  to `self.assert_ui_notification` and let it decide what the "header" will be. Then we don't need a gratuitous `get_ui_notification_dict` function (which actually has nothing to do with notifications, and just composes a dict from the elements passed to it)

Also note that putting this into the APITestCase class implies that this pattern is going to be how we do UI notifications in the future, since we're building this concept into our tooling. I think this is a good thing and we should probably double down on it - for instance, it should probably be documented and maybe we should look for ways to incorporate this into our tooling on the views side as well as on the testing side.

web/impact/impact/tests/api_test_case.py
```diff
@@ -98,5 +98,10 @@ def assert_options_include(self, method, expected_options, object_id=None):
...
 
-    def assert_ui_notification(self, response, notification):
-        self.assertEqual(response.data, notification)
+    def assert_ui_notification(self, response, success, notification):
+        expected_ui_notification = {
+            'success': success,
+            'header': self.success_header if success else self.fail_header,
+            'detail': notification
+        }
+        self.assertEqual(response.data, expected_ui_notification)
```
jonkiparsky: Is there any possibility that response.data might ever include information that is not part of the notification bundle? For example, a list of created objects, or something of that sort?

If that is a possibility, then we should check the individual keys, something like:
```
    def assert_ui_notification(self, response, success, notification):
        data = response.data
        self.assertTrue(all(
            data['success'] == success,
            data['header'] == self.success_header if success else self.fail_header,
            data['detail'] == notification
        ), msg="Notification data was not as expected")

```

It would be possible to make this more informative, by including the precise failure in the `msg` but we can save that for another day.

iankigen: from the current office hours tickets, the response for create/update doesn't return any additional data however I think asserting for the individual keys is the best idea in the long run. I'm on it

jonkiparsky (approved):
This looks great, thanks for bearing with me on those changes!

shanbady (approved):

PR #423, opened 2020-05-27
[AC-7790v2] P0: Mentor confirmation emails not sending

freshprincekla:
Changes introduced in [AC-7790](https://masschallenge.atlassian.net/browse/AC-7790)
 -  change the from_email from an array to a string as per django.core.mail docs, so as to get rid of the `smtplib.SMTPDataError: (554, b'Transaction failed: Illegal address') ` error


shanbady (approved):
looks good to me now.

jonkiparsky (approved):

PR #437, opened 2020-06-10
AC-7833, AC-7833 and AC-7834: Fix impact Api tests and get them to run as part of accelerate test suite

shanbady:
See associated PR for instructions:
https://github.com/masschallenge/accelerate/pull/2635

### Travis build is not expected to pass.


pbx (dismissed):
Odd Travis failure: `django.db.utils.OperationalError: table "auth_user" already exists`

shanbady:
> Odd Travis failure: `django.db.utils.OperationalError: table "auth_user" already exists`
@pbx 


This appears to be also happening on the AC-7469. I suspected that the changes to settings.py have caused this error so i reverted the settings.py file which seemed to resolve the initial issue but uncovered something else - the tests when running standalone on this branch and AC-7469 will fail due to `from mc.models import (
ModuleNotFoundError: No module named 'mc'`. we can either decide to resolve all these by using the model swapper or we can move forward with the assumption that we only care about the test suite passing in the merged accelerate repo context.

pbx:
@shanbady - Interesting. I saw one recent build (outside this epic) that failed with `LookupError: No installed app with label 'mc'` -- maybe only a superficial similarity, don't know yet.

Re getting the test suite passing standalone, my apologies, in that moment I forgot that this was an impact-api PR. Failure is definitely expected.

Yeah, I think passing standalone impact-api builds will be difficult to impossible, given the types of changes we're making. Let's just make sure that each impact-api PR is paired with an accelerate PR where the API tests do run (and pass).

web/impact/impact/graphql/types/base_user_profile_type.py
```diff
@@ -31,7 +31,7 @@ def resolve_program_roles(self, info, **kwargs):
         without changing GraphQL queries on the front end.
         """
         user_roles_of_interest = [UserRole.FINALIST, UserRole.ALUM]
-        startup_roles_of_interest = StartupRole.WINNER_STARTUP_ROLES
+        startup_roles_of_interest = StartupRole.WINNER_STARTUP_ROLES.copy()
```
pbx, line 34: Great find, this.

web/impact/impact/settings.py
```diff
@@ -14,7 +14,7 @@
```
pbx, line 14: I'm assuming that changes in this file are to be ignored. Please correct me if something (eek) is still importing from here.

web/impact/impact/tests/test_create_edit_office_hours_view.py
```diff
@@ -2,13 +2,14 @@
...
 from django.urls import reverse
-
+import swapper
 from accelerator.models import MentorProgramOfficeHour, UserRole
 from accelerator.tests.contexts import UserRoleContext
 from accelerator.tests.factories import (
     MentorProgramOfficeHourFactory,
     ProgramFactory
 )
+from accelerator.apps import AcceleratorConfig
```
pbx, line 12: I commented on this earlier -- since we're only pulling this in to reference `AcceleratorConfig.name`, and that is always (unless I'm missing something) set to `"accelerator"`, can we use the literal instead and drop the imports?

web/impact/impact/tests/test_graphql.py
```diff
@@ -154,7 +153,7 @@ def test_mentor_program_role_grants_only_returns_mentor_roles(self):
             program_role__user_role=_mentor, person=user)
         program = mentor_role_grant.program_role.program
         program_overview_link = program.program_overview_link
-        judge_role_grant = ProgramRoleGrantFactory.create(
+        ProgramRoleGrantFactory.create(
```
pbx, line 156: Good catch, I think that assignment was bit of cruft I neglected to clean up in AC-7771.

pbx (changes requested):
This is a great piece of work. Just a few questions/requests.

geofrocker (approved):
QA looks good ✅
- Ran tests locally and they all passed

web/impact/impact/tests/test_user_history_view.py
```diff
@@ -23,6 +23,8 @@
     days_from_now,
     find_events,
 )
+import factory
```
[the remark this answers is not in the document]
shanbady: its used by the decorator https://github.com/masschallenge/impact-api/pull/437/files/09e3f2a8c1c3f70e311bcc4d0de139812e833b13#diff-637ab44ef3afd24a6aa1fb2860501349R67

web/impact/impact/urls.py
```diff
@@ -42,37 +42,35 @@
 simpleuser_router = routers.DefaultRouter()
 simpleuser_router.register('User', GeneralViewSet, base_name='User')
 
-for model in apps.get_models('accelerator'):
-    if (model._meta.app_label == 'accelerator' and not
-            model._meta.auto_created and
-            model.__name__ not in MODELS_TO_EXCLUDE_FROM_URL_BINDING):
-        schema_router.register(
-            model, url=model_name_to_snake(model.__name__))
+if len(schema_router.registry) == 0:
```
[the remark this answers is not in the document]
shanbady: this is needed because it seems that in the context of running impact in accelerate, the urls file gets loaded by the interpreter twice and so a double-registration of these views happen.

web/impact/impact/tests/test_create_edit_office_hours_view.py
pbx: (We've discussed this and agreed that factoring this out deserves its own ticket -- there are over 100 `AcceleratorConfig` references.)

web/impact/impact/tests/test_impact_email_backend.py
```diff
@@ -4,13 +4,13 @@
 from ..minimal_email_handler import MinimalEmailHandler
 
 
-IMPACT_BACKEND_PATH = 'impact.impact_email_backend.ImpactEmailBackend'
+IMPACT_BACKEND_PATH = 'api.web.impact.impact.impact_email_backend.ImpactEmailBackend'
```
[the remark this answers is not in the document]
shanbady: I'm not entirely sure myself why the prefix can be dropped from settings but not from this test. my best guess is that within the context of the test run, the settings mock does not yet recognize impact as an app at the time we patch this setting.

web/impact/impact/tests/test_user_history_view.py
```diff
@@ -62,6 +64,7 @@ def test_user_joined_startup(self):
                 id=startup.organization.id)
             self.assertEqual(startup_str, events[0]["description"])
 
+    @factory.django.mute_signals(signals.pre_save, signals.post_save)
```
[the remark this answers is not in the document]
shanbady: added comment.

web/impact/impact/tests/test_create_edit_office_hours_view.py
pbx: https://masschallenge.atlassian.net/browse/AC-7856

web/impact/impact/tests/test_user_history_view.py
```diff
@@ -64,6 +64,11 @@ def test_user_joined_startup(self):
                 id=startup.organization.id)
             self.assertEqual(startup_str, events[0]["description"])
 
+    '''
```
pbx: Thanks for adding some explanation. Can you please make this an actual comment, rather than a string-floating-in-space?

web/impact/impact/tests/test_user_history_view.py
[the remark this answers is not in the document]
pbx: Ah, sorry, don't know how I missed that!

web/impact/impact/tests/test_user_history_view.py
shanbady: done.

web/impact/impact/settings.py
shanbady: correct - we can safely ignore changes to this file.

pbx (approved):

PR #449, opened 2020-06-29
[AC-7888] Prevent experts from creating conflicting office hours

iankigen:
## [AC-7888](https://masschallenge.atlassian.net/browse/AC-7888)

**Changes introduced**
- prevent experts from creating conflicting office hours

**Testing**
- As an office hour holder in an active program, go to http://localhost:8000/api/v1/office_hour/
- Creating and editing conflicting office hours should not be possible

Confirm that this PR meets the definition of done according to  [AC-7888](https://masschallenge.atlassian.net/browse/AC-7888)

### DIFF
--- a/web/impact/impact/tests/test_create_edit_office_hours_view.py
+++ b/web/impact/impact/tests/test_create_edit_office_hours_view.py
@@ -275,6 +275,18 @@ class TestCreateEditOfficeHourView(APITestCase):
         with self._assert_office_hour_created(created=True):
             self._create_office_hour_session(mentor, data)
 
+    def test_cant_create_session_when_conflict_exists_in_the_block(self):
+        mentor = self._expert_user(UserRole.MENTOR)
+        start_time = _now()
+        self._create_office_hour_obj(
+            mentor, start_date_time=start_time, block_duration=120)
+        data = self._get_post_request_data(
+            mentor,
+            get_data={'start_date_time': start_time + timedelta(minutes=60),
+                      'end_date_time': start_time + timedelta(minutes=150)})
+        with self._assert_office_hour_created(count=0):
+            self._create_office_hour_session(mentor, data)
+
     def _expert_with_inactive_program(self, role):
         program = ProgramFactory(program_status='ended')
         return self._expert_user(role, program)
@@ -315,10 +327,11 @@ class TestCreateEditOfficeHourView(APITestCase):
                                 minutes_from_now=0,
                                 finalist=None,
                                 start_date_time=None,
+                                block_duration=30,
                                 ):
         start_time = start_date_time or _now() + timedelta(
             minutes=minutes_from_now)
-        end_time = start_time + timedelta(minutes=30)
+        end_time = start_time + timedelta(minutes=block_duration)
         return MentorProgramOfficeHourFactory(
             mentor=mentor,
             start_date_time=start_time,
--- a/web/impact/impact/tests/test_create_edit_office_hours_view.py
+++ b/web/impact/impact/tests/test_create_edit_office_hours_view.py
@@ -286,6 +286,32 @@ class TestCreateEditOfficeHourView(APITestCase):
         with self._assert_office_hour_created(count=0):
             self._create_office_hour_session(mentor, data)
 
+    def test_cant_create_session_with_enclosing_conflict(self):
+        mentor = self._expert_user(UserRole.MENTOR)
+        start_time = _now()
+        self._create_office_hour_obj(
+            mentor, start_date_time=start_time, block_duration=120)
+        data = self._get_post_request_data(
+            mentor,
+            get_data={'start_date_time': start_time + timedelta(minutes=60),
+                      'end_date_time': start_time + timedelta(minutes=90)})
+        with self._assert_office_hour_created(count=0):
+            self._create_office_hour_session(mentor, data)
+
+    def test_mentor_cant_update_session_with_enclosing_conflicts(self):
+        mentor = self._expert_user(UserRole.MENTOR)
+        start_time = _now()
+        self._create_office_hour_obj(
+            mentor,
+            start_date_time=start_time + timedelta(minutes=-30),
+            block_duration=120)
+        office_hour = self._create_office_hour_obj(mentor)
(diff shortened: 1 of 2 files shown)


web/impact/impact/tests/test_create_edit_office_hours_view.py
```diff
@@ -216,6 +217,64 @@ def test_admin_cant_create_office_hour_for_non_mentor_user(self):
         with self._assert_office_hour_created(created=False):
             self._create_office_hour_session(self.staff_user(), data)
 
+    def test_mentor_cant_create_conflicting_sessions(self):
```
jonkiparsky, line 220: This test fails if we start the new sessions at `start_time + timedelta(hours=2)`

web/impact/impact/v1/serializers/office_hours_serializer.py
```diff
@@ -30,13 +31,27 @@ class Meta:
             'topics', 'description', 'location',
         ]
 
+    def handle_conflicting_sessions(self, attrs):
```
jonkiparsky: the def of done says
> If a create or edit request would create any office hour session in conflict with any existing session for the current user, return an error



This only checks one session at a time, so if the first session(s) requested are not in conflict, they are created, which is incorrect.

jonkiparsky (changes requested):
This does not conform to the def of done, see comment.

web/impact/impact/v1/serializers/office_hours_serializer.py
iankigen: That is not the case, `handle_conflicting_sessions` will be called on every session and if any of the session has an error, none will be created [office_hour_view.py#L122-L125](https://github.com/masschallenge/impact-api/blob/6160eded8d31962dd6e52dc7f23ba36753efd1ef/web/impact/impact/v1/views/office_hour_view.py#L122-L125)

web/impact/impact/tests/test_create_edit_office_hours_view.py
iankigen: That is the expected behavior. The test is expecting no office hours to be created since there will be conflicts when `start_date_time` = `start_time`. However, when `start_date_time` = `start_time + timedelta(hours=2)` the office hour is created because there are no conflicts which fails the test

jonkiparsky: Okay. In that case, let's write a test that explicitly tests that. 

For example, add a `block_duration=30` parameter to `_create_office_hour` and set `end_time=start_time+timedelta(block_duration)`

Then you can `self._create_office_hour(start_date_time=now, duration=120)` and request a block starting at `now + timedelta(60)`

web/impact/impact/v1/serializers/office_hours_serializer.py
jonkiparsky: 👍

web/impact/impact/tests/test_create_edit_office_hours_view.py
```diff
@@ -269,6 +344,11 @@ def _assert_update_office_hour(self, office_hour):
             pk=office_hour.id)
         self.assertEqual(updated_office_hour.topics, self.updated_topics)
 
+    def _assert_office_hour_not_updated(self, office_hour):
+        updated_office_hour = MentorProgramOfficeHour.objects.get(
+            pk=office_hour.id)
+        self.assertEqual(updated_office_hour.topics, office_hour.topics)
```
jonkiparsky: What if the test updates start or end time rather than topics?

iankigen: I have updated `_assert_office_hour_not_updated` to check for all the office hours fields

web/impact/impact/tests/test_create_edit_office_hours_view.py
```diff
@@ -276,6 +376,18 @@ def _assert_update_office_hour(self, office_hour):
             pk=office_hour.id)
         self.assertEqual(updated_office_hour.topics, self.updated_topics)
 
+    def _assert_office_hour_not_updated(self, office_hour):
+        updated_office_hour = MentorProgramOfficeHour.objects.get(
```
jonkiparsky: Hm. Just realized that there's a field called `updated_at` :)

No need to spend time on it, but the following would probably work just as well:

```
def _assert_office_hour_not_updated(self, office_hour):
    previously_updated_at = office_hour.updated_at
    office_hour.refresh_from_db()
    self.assertEqual(previously_updated_at, office_hour.updated_at)
```

web/impact/impact/v1/serializers/office_hours_serializer.py
```diff
@@ -32,6 +34,29 @@ class Meta:
...
+        office_hour = self.instance
+        start_time = attrs.get('start_date_time', None)
+        end_time = attrs.get('end_date_time', None)
+        skip_check = (office_hour and
+                      (office_hour.start_date_time == start_time and
+                       office_hour.end_date_time == end_time))
+        if (start_time or end_time) and not skip_check:
```
jonkiparsky, line 59: Extracting the body of this conditional to a separate function may resolve the code climate warning

web/impact/impact/v1/serializers/office_hours_serializer.py
```diff
@@ -32,6 +34,29 @@ class Meta:
...
+                            Q(end_date_time__lt=end_time))
+            enclosing_conflict = (Q(start_date_time__lte=start_time) &
+                                  Q(end_date_time__gte=end_time))
+            conflict = mentor.mentor_officehours.filter(
+                start_conflict | end_conflict | enclosing_conflict
+            ).exists()
+
+            if conflict:
+                raise ValidationError({
+                    'start_date_time': CONFLICTING_SESSIONS})
```
jonkiparsky, line 58: This probably doesn't need to be on its own line

jonkiparsky (approved):

PrinceDavis (approved):
Almost Done.
- error message `There are conflicts with existing office hours` when conflict arises while creating new office hour ✅

PR #486, opened 2020-08-05
AC-7927 Remove files left from API-service not needed by API-app

pbx:
[AC-7927](https://masschallenge.atlassian.net/browse/AC-7927)

**Testing**: Confirm the following: 
* all the API app tests pass
* the API app is functional without this stuff
* that there isn't anything deleted here which should be retained (and which is not already covered by another ticket in the epic).

Bonus: identify things that can be dropped that I missed.

(Notice that unlike all the other AC-7469-based PRs, this one does _not_ fail Travis, because it no longer triggers a Travis build at all!)


philophilo:
@pbx Is this not the right time for the Makefile to also go, it doesn't seem applicable anymore.

pbx:
@philophilo correct! Funny that I missed that. Done now.

philophilo (approved):
✅ all tests pass
✅ the app is still functional
✅ identifiable files have been removed

PR #488, opened 2020-08-05
AC-7773: Prune API code that was only for standalone (service) mode, not used in submodule (app) mode

pbx:
## [AC-7773](https://masschallenge.atlassian.net/browse/AC-7773)

**Testing**: 
* Run the tests locally
* Look over the diff and agree that these things are not needed for the API app (things we still need for finishing the epic, e.g. `settings` contents, are still in `development` when we want to refer to them)

The goal here was not to be exhaustive -- which is to say, while I'm open to suggestion of other bits of code that could be removed, most of the rest of that sweeping-up work will happen in the remaining tickets in the epic. 

This ticket was mostly for things that 1) are unneeded by the API app _and_ 2) are less-likely candidates for opportunistic cleanup, because their files might not be touched by the remaining tickets.


philophilo (approved):
✅  all tests passed
✅  app runs as expected
✅  code still looks good

PR #493, opened 2020-08-13
[AC-8010]: Remove legacy program picker code

shakirandagire:
**Changes introduced in [AC-8010](https://masschallenge.atlassian.net/browse/AC-8010)**
Removed the program_family attribute from office hours

Test instructions found in the [sibling PR](https://github.com/masschallenge/front-end/pull/251) in the frontend


jonkiparsky (approved):
Approving. It is alarming to note that we are able to remove this code without any tests needing to be changed, but I can't think of any tests that need to be added here.

iankigen (approved):
Looks good

PR #499, opened 2020-09-03
[AC-8037] Update directory office hours calendar picker

PrinceDavis:
## [AC-8037](https://masschallenge.atlassian.net/browse/AC-8037)

**Changes introduced**
- allow consumers of `/api/v1/office_hours_calendar/` to pass `calendar_span` argument that specify the length of data to return
- change the request argument `date_spec` to `focal_date`

**Testing**
- See sibling PRs

### [Sibling 2](https://github.com/masschallenge/accelerate/pull/2742)

### DIFF
--- a/web/impact/impact/tests/test_office_hours_calendar_view.py
+++ b/web/impact/impact/tests/test_office_hours_calendar_view.py
@@ -46,18 +46,18 @@ from .utils import nonexistent_object_id
 class TestOfficeHoursCalendarView(APITestCase):
     view = OfficeHoursCalendarView
 
-    def test_no_focal_dateified_sees_current_week(self):
+    def test_no_focal_date_specified_sees_current_week(self):
         office_hour = self.create_office_hour()
         response = self.get_response(user=office_hour.mentor)
         self.assert_hour_in_response(response, office_hour)
 
-    def test_no_focal_dateified_does_not_see_last_week(self):
+    def test_no_focal_date_specified_does_not_see_last_week(self):
         office_hour = self.create_office_hour(
             start_date_time=days_from_now(-9))
         response = self.get_response(user=office_hour.mentor)
         self.assert_hour_not_in_response(response, office_hour)
 
-    def test_focal_dateified_sees_sessions_in_range(self):
+    def test_focal_date_specified_sees_sessions_in_range(self):
         two_weeks_ago = days_from_now(-14)
         focal_date = two_weeks_ago.strftime(ISO_8601_DATE_FORMAT)
         office_hour = self.create_office_hour(
@@ -67,30 +67,16 @@ class TestOfficeHoursCalendarView(APITestCase):
         self.assert_hour_in_response(response, office_hour)
 
     def test_calendar_data_for_month_span_last_day_of_the_month(self):
-        today = datetime.now()
-        days_in_month = calendar.monthrange(today.year, today.month)[1]
-        days_to_month_end = days_in_month - int(today.strftime("%d"))
-        month_end = days_from_now(days_to_month_end)
-        office_hour = self.create_office_hour(start_date_time=month_end)
-        focal_date = today.strftime(ISO_8601_DATE_FORMAT)
+        last_month = days_from_now(-40)
+        last_month_end = days_from_now(-31)
+        office_hour = self.create_office_hour(start_date_time=last_month_end)
+        focal_date = last_month.strftime(ISO_8601_DATE_FORMAT)
         response = self.get_response(user=office_hour.mentor,
                                      focal_date=focal_date,
                                      calendar_span="month")
         self.assert_hour_in_response(response, office_hour)
 
-    def test_calendar_data_for_month_span_mid_month(self):
-        today = datetime.now()
-        days_in_month = calendar.monthrange(today.year, today.month)[1]
-        days_to_mid_month = (days_in_month - int(today.strftime("%d"))) / 2
-        mid_month = days_from_now(days_to_mid_month)
-        office_hour = self.create_office_hour(start_date_time=mid_month)
-        focal_date = today.strftime(ISO_8601_DATE_FORMAT)
-        response = self.get_response(user=office_hour.mentor,
-                                     focal_date=focal_date,
-                                     calendar_span="month")
-        self.assert_hour_in_response(response, office_hour)
-
-    def test_focal_dateified_does_not_see_sessions_not_in_range(self):
+    def test_focal_date_specified_does_not_see_sessions_not_in_range(self):
         two_weeks_ago = days_from_now(-14)
         focal_date = two_weeks_ago.strftime(ISO_8601_DATE_FORMAT)
         office_hour = self.create_office_hour()
--- a/web/impact/impact/tests/test_office_hours_calendar_view.py
+++ b/web/impact/impact/tests/test_office_hours_calendar_view.py
@@ -26,6 +26,7 @@ from accelerator.tests.factories import (
     ProgramFamilyLocationFactory,
     ProgramRoleGrantFactory,
     StartupTeamMemberFactory,
+    UserFactory
 )
 from accelerator.tests.contexts import UserRoleContext
 from accelerator.tests.contexts.context_utils import get_user_role_by_name
@@ -88,6 +89,19 @@ class TestOfficeHoursCalendarView(APITestCase):
                                      focal_date=focal_date,
                                      calendar_span="month")
         self.assert_hour_in_response(response, office_hour)
+
+    def test_return_upcoming_calendar_data(self):
+        two_weeks_ago = days_from_now(-14)
+        focal_date = two_weeks_ago.strftime(ISO_8601_DATE_FORMAT)
+        office_hour = self.create_office_hour(
+            start_date_time=two_weeks_ago,
+            finalist=UserFactory())
+        office_hour1 = self.create_office_hour(start_date_time=two_weeks_ago)
+        response = self.get_response(user=office_hour1.mentor,
+                                     focal_date=focal_date,
+                                     upcoming=True)
+        self.assert_hour_not_in_response(response, office_hour)
+        self.assert_hour_in_response(response, office_hour1)
 
     def test_focal_dateified_does_not_see_sessions_not_in_range(self):
         two_weeks_ago = days_from_now(-14)
@@ -454,7 +468,8 @@ class TestOfficeHoursCalendarView(APITestCase):
                      user=None,
                      target_user_id=None,
                      focal_date=None,
-                     calendar_span=None):
+                     calendar_span=None,
+                     upcoming=None):
         user = user or self.staff_user()
         user.set_password("password")
         user.save()
@@ -464,6 +479,8 @@ class TestOfficeHoursCalendarView(APITestCase):
             data['focal_date'] = focal_date
         if calendar_span is not None:
             data['calendar_span'] = calendar_span
+        if upcoming is not None:
+            data['upcoming'] = upcoming
         if target_user_id is not None:
             data['user_id'] = target_user_id
         with self.login(email=user.email):
--- a/web/impact/impact/tests/test_office_hours_calendar_view.py
+++ b/web/impact/impact/tests/test_office_hours_calendar_view.py
@@ -65,7 +65,7 @@ class TestOfficeHoursCalendarView(APITestCase):
                                      focal_date=focal_date)
         self.assert_hour_in_response(response, office_hour)
 
-    def test_calendar_data_for_monthly_span_last_day_of_the_month(self):
+    def test_calendar_data_for_month_span_last_day_of_the_month(self):
         today = datetime.now()
         days_in_month = calendar.monthrange(today.year, today.month)[1]
         days_to_month_end = days_in_month - int(today.strftime("%d"))
@@ -74,10 +74,10 @@ class TestOfficeHoursCalendarView(APITestCase):
         focal_date = today.strftime(ISO_8601_DATE_FORMAT)
         response = self.get_response(user=office_hour.mentor,
                                      focal_date=focal_date,
-                                     calendar_span="monthly")
+                                     calendar_span="month")
         self.assert_hour_in_response(response, office_hour)
 
-    def test_calendar_data_for_monthly_span_mid_month(self):
+    def test_calendar_data_for_month_span_mid_month(self):
         today = datetime.now()
         days_in_month = calendar.monthrange(today.year, today.month)[1]
         days_to_mid_month = (days_in_month - int(today.strftime("%d"))) / 2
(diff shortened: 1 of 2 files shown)


web/impact/impact/tests/test_office_hours_calendar_view.py
```diff
@@ -44,45 +45,69 @@
 class TestOfficeHoursCalendarView(APITestCase):
     view = OfficeHoursCalendarView
 
-    def test_no_date_specified_sees_current_week(self):
+    def test_no_focal_dateified_sees_current_week(self):
```
jonkiparsky: global-replace issues...

(also see below)

web/impact/impact/v1/views/office_hours_calendar_view.py
```diff
@@ -323,23 +330,28 @@ def _make_f_expression(field, location_path):
...
         initial_date = datetime.now()
     initial_date = utc.localize(initial_date)
     # This calculation depends on the fact that monday == 0 in python
-    start_date = initial_date - timedelta(initial_date.weekday())
-    end_date = start_date + ONE_WEEK + ONE_DAY
-    adjusted_start_date = start_date - ONE_DAY
+    if calendar_span == "monthly":
+        start_date = initial_date - timedelta(int(initial_date.strftime("%d")))
+        end_date = start_date + ONE_MONTH + ONE_DAY
+        adjusted_start_date = start_date + ONE_DAY
```
jonkiparsky: shouldn't this be minus ONE_DAY?

web/impact/impact/v1/views/office_hours_calendar_view.py
```diff
@@ -323,23 +330,28 @@ def _make_f_expression(field, location_path):
...
+    else:
+        start_date = initial_date - timedelta(initial_date.weekday())
```
jonkiparsky: the comment associated with this line should stay with this line (line 346)

web/impact/impact/v1/views/office_hours_calendar_view.py
```diff
@@ -323,23 +330,28 @@ def _make_f_expression(field, location_path):
...
+        initial_date = datetime.strptime(focal_date, ISO_8601_DATE_FORMAT)
     else:
         initial_date = datetime.now()
     initial_date = utc.localize(initial_date)
     # This calculation depends on the fact that monday == 0 in python
-    start_date = initial_date - timedelta(initial_date.weekday())
-    end_date = start_date + ONE_WEEK + ONE_DAY
-    adjusted_start_date = start_date - ONE_DAY
+    if calendar_span == "monthly":
+        start_date = initial_date - timedelta(int(initial_date.strftime("%d")))
```
jonkiparsky: This computation differs from the one for "week" only in two numbers, the stepback number that you use to compute the start of the range and the span. Everything else about the computation, including the padding by a day, is the same. 

You could do:
```
if span = "month":
    step_back = timedelta(int(initial_date.strftime("%d")))
    span = ONE_MONTH
else:
    step_back = timedelta(initial_date.weekday())
    span = ONE_WEEK
return compute_date_range(initial_date, step_back, range)
```

web/impact/impact/v1/views/office_hours_calendar_view.py
```diff
@@ -33,9 +35,13 @@
 from accelerator_abstract.models.base_user_utils import is_employee
 User = get_user_model()
 
+today = datetime.today()
+DAYS_IN_MONTH = calendar.monthrange(today.year, today.month)[1]
```
jonkiparsky: This assumes that we're only interested in the current month. Users who consult this page in February to find out about office hours in March might find themselves disappointed! :)

I would suggest that we just say there's 31 days in a month, and let the front end discard anything that it doesn't want to show. (just as we are already padding out the week)

web/impact/impact/v1/views/office_hours_calendar_view.py
```diff
@@ -113,13 +119,14 @@ def _check_request_user_type(self, request):
         return True
 
     def _get_date_range(self, request):
-        date_spec = request.query_params.get("date_spec", None)
+        focal_date = request.query_params.get("focal_date", None)
+        calendar_span = request.query_params.get("calendar_span", "weekly")
```
jonkiparsky: I would prefer to use "week" and "month" rather than "weekly" and "monthly" 
Reasoning: "weekly" and "monthly" indicate the period on which an event repeats: this is a weekly event, that one is monthly. Here, we want to know what span of time the user is interested in, either a week or a month.

jonkiparsky (changes requested):
A few first-pass changes.

web/impact/impact/v1/views/office_hours_calendar_view.py
PrinceDavis: Nope

jonkiparsky:
```
>>> initial_date = datetime.now()
>>> start_date = initial_date - timedelta(int(initial_date.strftime("%d")))
>>> start_date
datetime.datetime(2020, 8, 31, 8, 59, 11, 83622)
>>> start_date + timedelta(1)
datetime.datetime(2020, 9, 1, 8, 59, 11, 83622)
```

Probably could get away with leaving off the adjustment step here, but adding a day cuts into the first of the month.

web/impact/impact/v1/views/office_hours_calendar_view.py
PrinceDavis: The similarity is just one line `(the end_date line)`  `adjusted_start_date` calculation is different for both

web/impact/impact/v1/views/office_hours_calendar_view.py
PrinceDavis: I am not sure I follow

jonkiparsky: 9/1/2020 at 8:59 is nine hours into the month. When we use this to filter, we will miss any sessions scheduled for the morning of 9/1 (in this case)
```
        return MentorProgramOfficeHour.objects.filter(
            active_mentors,
            start_date_time__range=[self.start_date, self.end_date])
```

If we just take the originally-calculated initial_date, we get a date that has effectively got an adjustment between 1 and 3600 seconds before the start of month, which is good, but leaves us maybe vulnerable to timezone damage. ie, there may be scenarios where a user at certain times-of-day in certain timezones might end up not getting all of the sessions scheduled for that month, in their timezone. 

The easiest solution is to simply add 24 hours on either end as a buffer, which is what we do in the week case. This ensures that the front-end gets everything it needs (and if it gets some extra, it can drop those on the floor, we don't care)

web/impact/impact/v1/views/office_hours_calendar_view.py
```diff
@@ -34,10 +36,13 @@
 from accelerator_abstract.models.base_user_utils import is_employee
 User = get_user_model()
 
+
 ISO_8601_DATE_FORMAT = "%Y-%m-%d"
 ONE_DAY = timedelta(1)
 ONE_WEEK = timedelta(8)
 
+
```
geofrocker, line 43: Only return upcoming office hours for the directory?

geofrocker (changes requested):

web/impact/impact/v1/views/office_hours_calendar_view.py
PrinceDavis: Thanks for clarifying

web/impact/impact/v1/views/office_hours_calendar_view.py
```diff
@@ -66,18 +71,18 @@ class OfficeHoursCalendarView(ImpactView):
...
         (self._get_target_user(request) and
-         self._check_request_user_type(request) and
+         self._check_request_user_type() and
          self._get_date_range(request) and
          self._set_user_query() and
-         self._get_office_hours_data())
+         self._get_office_hours_data(request))
         return Response(self.response_elements)
 
     def _get_target_user(self, request):
```
geofrocker, line 87: We no longer need the employee check. This logics to me means that if the user making the request is not a staff user and the user_id that has been passed is not there own, return a permission error. We need finalists to be able to fetch a specific mentor's office hours too. cc @jonkiparsky

geofrocker (changes requested):

web/impact/impact/v1/views/office_hours_calendar_view.py
PrinceDavis: @jonkiparsky what is your take on the above?

jonkiparsky: Yes, that makes sense

web/impact/impact/v1/views/office_hours_calendar_view.py
```diff
@@ -327,22 +337,44 @@ def _make_f_expression(field, location_path):
...
+    # throws ValueError if focal_date is not in ISO-8601 format
 
-    if date_spec:
-        initial_date = datetime.strptime(date_spec, ISO_8601_DATE_FORMAT)
+    if focal_date:
+        initial_date = datetime.strptime(focal_date, ISO_8601_DATE_FORMAT)
     else:
         initial_date = datetime.now()
+
+    if upcoming and initial_date < datetime.now():
```
jonkiparsky: The `upcoming` parameter, provided by the client, seems to exist only to override the date provided by the frontend client. This seems confusing and unnecessary. 
My understanding is that we providing the office hour sessions for a given calendar month. If so, we should do that. 
If not, maybe we need to clarify the objective.

PrinceDavis: We want to provide office hour for a calendar month, however sometimes we filter out out those that are past the current date, as as user shouldn't be able to reserve past office hours. cc @jonkiparsky

geofrocker: @jonkiparsky Unlike the new office hours on `accelerate`, the directory is only interested in upcoming office hours since it's only meant to be used by finalists for reservation.

jonkiparsky: Okay, so it seems like the simplest thing to do is to let the front end deal with this - either display past sessions as unreservable, or just drop them on the floor.

Doing this in the back end adds needless complexity both to the function itself and to the interface that it defines.

PrinceDavis: The front end code for this part of the interface is already involved, I think its best the back-end handle this filtering; adding another layer of complexity whereby the interface will loop through the list of return office hours will further complicate it.
Also if the backend handle this we get uniform performance across devices. and we don't have to transport as much data through the wire.
@jonkiparsky  @geofrocker your thought please

jonkiparsky: - The amount of data is a red herring, IMO. I think it's unlikely that a user will create enough office hours in a month for this to make a difference, and in any case, if the amount of data in a full month's sessions is problematic then we're going to have that problem at the start of month, so we'll have to deal with it regardless. 
- I would assume that the front end is already looping over the data in order to display it. This would simply ask it to check if the date whether the session is in the past in that already-existing loop.
- If we want to do this in the back end, we'll have to go back to the drawing board and redesign the endpoint's interface and also refactor the implementation. Having a parameter that nullifies other parameters will certainly lead to problems.

PrinceDavis: You make a good points here: 

- the frontend does not loop over the sessions, at least our code doesn't however the calendar package we use does but outside of our codebase.

- Given that the other option is to re-write the endpoint we can accept the trade off and do the implementation on the front end.

PrinceDavis:
@geofrocker @jonkiparsky this is ready for another round of review

web/impact/impact/tests/test_office_hours_calendar_view.py
```diff
@@ -44,45 +46,69 @@
...
```
jonkiparsky: @PrinceDavis Please fix the test names that got munged in global-replace

web/impact/impact/tests/test_office_hours_calendar_view.py
```diff
@@ -44,45 +46,69 @@
...
             start_date_time=two_weeks_ago)
         response = self.get_response(user=office_hour.mentor,
-                                     date_spec=date_spec)
+                                     focal_date=focal_date)
         self.assert_hour_in_response(response, office_hour)
 
-    def test_date_specified_does_not_see_sessions_not_in_range(self):
+    def test_calendar_data_for_month_span_last_day_of_the_month(self):
+        today = datetime.now()
+        days_in_month = calendar.monthrange(today.year, today.month)[1]
```
jonkiparsky: Tests should be consistent, ie, they should test the same thing every time. Please make this test a consistent month. By convention, we tend to use a month in the past. 
You can use freezegun to force the time to be some particular datetime (see examples in the codebase)

web/impact/impact/v1/views/office_hours_calendar_view.py
```diff
@@ -327,22 +326,36 @@ def _make_f_expression(field, location_path):
...
+def _date_range(calendar_span, focal_date=None):
     # returns (start_date, end_date)
+    # When calendar_span == week and upcoming == False
     # start_date is the latest monday that is less than or equal to today,
     # end_date is start_date + seven days
+    # When calendar_span == month and upcoming == False
+    # start_date is the first day of the new month
+    # end_date is start_date + length of the current month
     # both values are then padded by 24 hours to allow for TZ differences
+    # when upcoming == True
```
jonkiparsky: please update the docstring

web/impact/impact/v1/views/office_hours_calendar_view.py
```diff
@@ -327,22 +326,36 @@ def _make_f_expression(field, location_path):
...
     else:
         initial_date = datetime.now()
+
     initial_date = utc.localize(initial_date)
-    # This calculation depends on the fact that monday == 0 in python
-    start_date = initial_date - timedelta(initial_date.weekday())
-    end_date = start_date + ONE_WEEK + ONE_DAY
+    if calendar_span == "month":
+        start_date = initial_date - timedelta(int(initial_date.strftime("%d")))
+        days_in_month = calendar.monthrange(start_date.year, start_date.month)[1]
```
jonkiparsky, line 351: We don't seem to be using this

jonkiparsky: This calculation seems unnecessary. Why not just call a month 31 days?

web/impact/impact/v1/views/office_hours_calendar_view.py
jonkiparsky: Okay, we should probably avoid this pattern in the future but I'm going to let it go for now.

jonkiparsky (changes requested):
A few remaining bits of cleanup

jonkiparsky (approved):

geofrocker (approved):
QA looks good. Will merge when travis passes &  frontend is approved.

PR #502, opened 2020-09-15
AC-8187 - Fix failing Office Hours test on AC-7469

pbx:
## [AC-8187](https://masschallenge.atlassian.net/browse/AC-8187)

**Review suggestions:**
* Confirm that the test named in the ticket (`test_staff_user_sees_staff_hours`) now passes
* Check that the logic of the changed test still upholds the intended purpose of the test


philophilo (approved):
✅  code looks good
✅  test also works locally

PR #505, opened 2020-09-17
AC-8036 Update directory office hours reservation and confirmation method

geofrocker:
### Changes Introduced in [AC-8036](https://masschallenge.atlassian.net/browse/AC-8036):
- Remove program association to availableofficehours logic
- Remove old office hours view logic
- Update time_card info
### Testing:
- Follow testing instructions on the [sibling PR](https://github.com/masschallenge/front-end/pull/258)


web/impact/impact/v1/views/reserve_office_hour_view.py
```diff
@@ -213,8 +213,11 @@ def _succeed(self):
         self.timecard_info = {
             "finalist_first_name": self.target_user.first_name,
             "finalist_last_name": self.target_user.last_name,
+            "finalist_email": self.target_user.email,
             "topics": self.message,
-            "startup": startup_name}
+            "startup": startup_name,
+            "calendar_data": self.calendar_data,
```
jonkiparsky: This is unsafe. 
Since it `self.calendar_data` is set in `prepare_email_notification`, this will break if the order in which these functions run is ever changed. If we want to make this change, let's do it a little more cautiously. 


I would suggest that `self.get_calendar_data` should check to see if `hasattr(self, "_calendar_data)`. If so, return it. If not, compute the calendar data and set `self._calendar_data=calendar_data`. 

Having done that, you can safely call `self.get_calendar_data()` on lines 219 and 191 (and you don't need the assignment on line 186)

web/impact/impact/tests/test_reserve_office_hour_view.py
```diff
@@ -259,14 +259,18 @@ def test_outlook_link_in_reserve_office_hour_email_to_mentor(self):
     def assert_response_contains_session_details(self, response, office_hour):
         office_hour.refresh_from_db()
         timecard = response.data['timecard_info']
+        calendar_data = timecard['calendar_data']
         if office_hour.startup:
             startup_name = office_hour.startup.organization.name
         else:
             startup_name = ""
+        calendar_data = timecard['calendar_data']
```
jonkiparsky: this line is repeated

web/impact/impact/tests/test_reserve_office_hour_view.py
```diff
@@ -259,14 +259,18 @@ def test_outlook_link_in_reserve_office_hour_email_to_mentor(self):
...
         oh_details = {'finalist_first_name': office_hour.finalist.first_name,
                       'finalist_last_name': office_hour.finalist.last_name,
                       'topics': office_hour.topics,
-                      'startup': startup_name}
+                      'startup': startup_name,
+                      "finalist_email": office_hour.finalist.email,
+                      "calendar_data": calendar_data}
         self.assertDictEqual(timecard, oh_details)
```
jonkiparsky: This is a weird pattern - we construct `oh_details` using data from the `timecard` and then we compare the result to `timecard`. 

I understand the motivation - you want to use `assertDictEqual` and you need the calendar_data to match -  but inserting the data from the actual result into the expected result seems like it crosses a line.

jonkiparsky (changes requested):

web/impact/impact/tests/test_reserve_office_hour_view.py
geofrocker: `calendar_data` returned from the Add2cal package has variables data like UUID's that will always return different values. Hence we can't get the same result from different calls. My intension here was to just test that the response contains the `calendar_data` key.

web/impact/impact/v1/views/reserve_office_hour_view.py
```diff
@@ -256,7 +260,7 @@ def get_calendar_data(self, counterpart_name):
         location_info = location_info.format(location=location,
                                              separator=separator,
                                              meeting_info=meeting_info)
-        return Add2Cal(
+        self.calendar_data = Add2Cal(
```
jonkiparsky, line 263: Typically you'd use a leading underscore to indicate that this field is in some sense "private" (ie, that one should not rely on it being there, and should call the get method instead) 

Not super critical, but it's useful to abide by these conventions

jonkiparsky (approved):

freshprincekla (approved):
Code changes look good. And QA passes:
- created office hours in a future date
- Masqueraded as a finalist in the mentors program
- successfully was able to schedule an office hour

PR #509, opened 2020-10-05
AC-8246 re-ordering url includes

shanbady:
This PR fixes a url resolution error in the monolith api.

To test:

1- visit the endpoint on the latest release with monolith code (https://staging.masschallenge.org/api/oauth/token/) - note that it fails with a 500
2 - I have this deployed to test3 currently. visit the test3 version (https://test3.masschallenge.org/api/oauth/token/) and note that it responds with a 405 (expected response for that route) which indicates the url is resolving properly. 
3 - similarly the graphql endpoint gives a 404 on staging (https://staging.masschallenge.org/api/graphql/) but will give an appropriate response on test3 (https://test3.masschallenge.org/api/graphql/)

### DIFF
--- a/web/impact/impact/urls.py
+++ b/web/impact/impact/urls.py
@@ -66,10 +66,25 @@ urlpatterns = [
     url(r'^$', IndexView.as_view()),
     url(r'^v0/', include(v0_urlpatterns)),
     url(r'^v1/', include(v1_urlpatterns)),
+    url(r'^oauth/', include('oauth2_provider.urls',
+        namespace='oauth2_provider')),
     url(r'^(?P<app>\w+)/(?P<model>[a-z_]+)/'
         r'(?P<related_model>[a-z_]+)/$',
         GeneralViewSet.as_view({'get': 'list', 'post': 'create'}),
         name='related-object-list'),
+    url(r'^graphql/$',
+        csrf_exempt(SafeGraphQLView.as_view(
+            graphiql=settings.DEBUG,
+            schema=schema,
+            middleware=[
+                IsAuthenticatedMiddleware])),
+        name="graphql"),
+    url(r'^graphql/auth/$',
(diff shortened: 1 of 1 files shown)


pbx:
@shanbady I'm thinking about how to get this deployed. Seems to me we'll also need an accelerate PR that updates the submodule to commit `4ceb9ba5f`, and we'll need to roll a new release. 

If you have ideas for a faster path to production (that don't leave us in a messy state) I'm interested.

pbx (changes requested):
@shanbady this looks good, but I want to double check before  merge because I'm not seeing the response codes you named in steps 1 and 2. 

#1 `accelerate.masschallenge.org/api/oauth/token/` gives me a 404 rather than a 503.
#2 `test3.masschallenge.org/api/oauth/token/` gives me a (Cloudflare) 405 rather than a 402
#3 (These URLS do respond as you describe)

pbx:
(In the interest of getting this merged -- since @philophilo and @PrinceDavis will both be in before me tomorrow, either one of them can approve if they see fit, and I'll dismiss my review)

PrinceDavis (changes requested):
The outcome of my review is same result as @pbx 

#1 `accelerate.masschallenge.org/api/oauth/token/` returns 404 rather than a 503.
#2 `test3.masschallenge.org/api/oauth/token/`returns 405 rather than a 402
#3  graphQL endpoint works as expected

shanbady:
@pbx @philophilo 

- I had a mistake in the test instructions - 405 is the expected response for /api/oauth/token  (it requires a post for it to be valid)
- The actual errors (the 500) actually come from staging not production (which presumably doesn't have the merged api code). I've updated the test instructions with the corrections.

pbx (approved):
The code change looks good to me and I'm satisfied to merge this.

PrinceDavis (approved):

